PR #2, opened 2014-11-23
Update to Rust 81eeec094 and fix warnings

yuriks:


nikomatsakis:
Wow, I somehow totally missed this PR at the time, but I think it is no longer relevant. Thanks anyhow!

PR #3, opened 2015-03-12
Updated code to work in alpha1 (2015-03-10) + added testcase

faern:
Made lib itself compile in newer Rust.
Added test calculating fib

Not sure if this is the best solution, but I made it compile and the tests seems fine.


nikomatsakis:
Wow, I somehow totally missed this PR at the time, but I think it is no longer relevant. Thanks anyhow!

PR #4, opened 2015-12-19
Fix broken example in Readme.md

diwic:
The example becomes silly for real if it never increments anything. :-)


nikomatsakis:
True!

PR #11, opened 2016-01-01
Differentiate between spin and lock latches

### LINKED ISSUE #9: Busy yield-loops hurt performance
I noticed that the kernel scheduler was showing up a lot in my perf report, and I tracked this down to having so many `sched_yield` syscalls in waiting threads.  Such busy work is wasted cpu time, not only for your own process but also if anything else is trying to run on the system.

I tried a crude patch below to use `Condvar` for the blocking latch wait, and it already shows a clear benefit.  The other yield loop is in `steal_until`, which is harder because that's looping for either latch completion or available work -- suggestions welcome.

```diff
diff --git a/src/latch.rs b/src/latch.rs
index a81831d1cc7a..fbeff95acbd4 100644
--- a/src/latch.rs
+++ b/src/latch.rs
@@ -1,9 +1,11 @@
+use std::sync::{Mutex, Condvar};
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::thread;
 
 /// A Latch starts as false and eventually becomes true. You can block
 /// until it becomes true.
 pub struct Latch {
+    m: Mutex<()>,
+    c: Condvar,
     b: AtomicBool
 }
 
@@ -11,19 +13,24 @@ impl Latch {
     #[inline]
     pub fn new() -> Latch {
         Latch {
+            m: Mutex::new(()),
+            c: Condvar::new(),
             b: AtomicBool::new(false)
         }
     }
 
     /// Set the latch to true, releasing all threads who are waiting.
     pub fn set(&self) {
+        let _guard = self.m.lock().unwrap();
         self.b.store(true, Ordering::SeqCst);
+        self.c.notify_all();
     }
 
     /// Spin until latch is set. Use with caution.
     pub fn wait(&self) {
+        let mut guard = self.m.lock().unwrap();
         while !self.probe() {
-            thread::yield_now();
+            guard = self.c.wait(guard).unwrap();
         }
     }
 
```

Before, running on Fedora 23 x86_64 with an i7-2600:

```
     Running `target/release/quicksort 1000`
Array length    : 1000K 32-bit integers
Repetitions     : 10
Parallel time   : 281422159
Sequential time : 848279007
Parallel speedup: 3.014258045685734
```

After, with Condvar wait:

```
     Running `target/release/quicksort 1000`
Array length    : 1000K 32-bit integers
Repetitions     : 10
Parallel time   : 225591277
Sequential time : 873530292
Parallel speedup: 3.872181157075502
```

nikomatsakis:
cc @cuviper -- this is based on your patch. What do you think? Though re-reading the issue comments I see you mentioned you might have been doing something like this, if so, sorry to step on your toes!

(Locally I don't see any real change in performance, but I've not done extensive testing, just run the quicksort demo.)

Fixes #9.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,4 +1,4 @@
-use latch::Latch;
+use latch::{Latch, LockLatch, SpinLatch};
 #[allow(unused_imports)]
 use log::Event::*;
 use job::{Code, CodeImpl, Job};
@@ -45,7 +45,7 @@ pub fn join<A,B,RA,RB>(oper_a: A,
         // done here so that the stack frame can keep it all live
         // long enough
         let mut code_b = CodeImpl::new(oper_b, &mut result_b);
-        let mut latch_b = Latch::new();
+        let mut latch_b = SpinLatch::new();
         let mut job_b = Job::new(&mut code_b, &mut latch_b);
         (*worker_thread).push(&mut job_b);
 
@@ -77,12 +77,12 @@ unsafe fn join_inject<A,B,RA,RB>(oper_a: A,
 {
     let mut result_a = None;
     let mut code_a = CodeImpl::new(oper_a, &mut result_a);
-    let mut latch_a = Latch::new();
+    let mut latch_a = LockLatch::new();
     let mut job_a = Job::new(&mut code_a, &mut latch_a);
 
     let mut result_b = None;
     let mut code_b = CodeImpl::new(oper_b, &mut result_b);
-    let mut latch_b = Latch::new();
+    let mut latch_b = LockLatch::new();
     let mut job_b = Job::new(&mut code_b, &mut latch_b);
 
     thread_pool::get_registry().inject(&[&mut job_a, &mut job_b]);
@@ -112,7 +112,7 @@ impl ThreadPool {
         unsafe {
             let mut result_a = None;
             let mut code_a = CodeImpl::new(op, &mut result_a);
-            let mut latch_a = Latch::new();
+            let mut latch_a = LockLatch::new();
             let mut job_a = Job::new(&mut code_a, &mut latch_a);
             self.registry.inject(&[&mut job_a]);
             latch_a.wait();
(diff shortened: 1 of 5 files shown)


nikomatsakis:
PS, I really should disable the non-nightly travis builds, I guess, or find a way to make them skip over the compilertest tests.

cuviper:
I was indeed working on it, but too slowly.  No worries, let's compare:
https://gist.github.com/cuviper/cf7a055c468d0447b570

* We both turned `Latch` into a trait, but I only kept `set` in common.  I found that callers really only needed either `probe` or `wait`, not both, so I thought it better to make sure that only the best latch type is used for each case.
* I left it as a trait object pointer in `Job`, instead of using generics as you did.  I suppose your way allows static dispatch, but then doesn't allow mixing job types.  But you didn't need to mix. *shrug*  AIUI this isn't critical-path code for performance anyway.
* The bool does need to be set under mutex, as I'll show in reply to that comment.
* I'm surprised you see no performance difference, as it's very clear here.  What size are you running for the demo?  I used `quicksort 1000` and higher.  It also makes a difference in a few of my Project Euler solutions I toyed with.

nikomatsakis:
@cuviper adjusted somewhat, primarily to address the race. I also removed `probe` from the trait -- I only threw it in on a whim, but having a more minimal interface seems better, and it also exposed the fact that I was using a trait object in one place I did not intend to.

nikomatsakis:
@cuviper I take your point about removing `wait` from the `Latch` trait, seems better that way! (done)

nikomatsakis:
> I suppose your way allows static dispatch, but then doesn't allow mixing job types. But you didn't need to mix. shrug AIUI this isn't critical-path code for performance anyway.

Yeah, it's probably no big deal, but it seems better to get static dispatch and to keep the `Job` data structure as lightweight as possible.

cuviper:
Looks good to me!

PR #12, opened 2016-01-03
add a `weight` combinator

### LINKED ISSUE #7: No way to influence par_iter's decision when to fall back to sequential
The threshold and/or the cost function is highly dependant on the workload.
For maximum performance on different workloads the user should be able to modify them.

nikomatsakis:
This adds a method `weight` that can be used to specify the relative weight of an operation. 1.0 means "very cheap". As expressed in the thread on #7, I would like to find a more elegant solution to this problem, but this seems like a simple thing to add for now.

Fixes #7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 Rayon is a data-parallelism library for Rust. It is extremely
 lightweight and makes it easy to convert a sequential computation into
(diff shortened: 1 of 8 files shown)


mneumann:
This works for me! I think we should go with this solution for now and improve it later on.

PR #14, opened 2016-01-05
Add an `enumerate` combinator

cuviper:
Returns the index into the original sequence along with each item.
Works like `Iterator::enumerate()`, but of course with threaded jobs it
won't be visited in any guaranteed order.  It has to assume that the
size of each split is known exactly though, so this wouldn't work after
a filtering function, for instance.

### DIFF
--- a/src/par_iter/enumerate.rs
+++ b/src/par_iter/enumerate.rs
@@ -1,0 +1,65 @@
+use super::{ParallelIterator, ParallelIteratorState, ParallelLen};
+
+pub struct Enumerate<M> {
+    base: M,
+}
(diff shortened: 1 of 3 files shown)


nikomatsakis:
Great, thanks! Sorry for the delay. I have been pondering adding a `filter` combinator, and was debating how this PR would fit in with that (as you highlight). I'm not quite sure, but I think we'll want enumerate where possible anyway. :) 

(I'm thinking that the ideal would probably be to add another trait, `ExactParallelIterator` or something like that, but I have to do some experiments.)

PR #15, opened 2016-01-19
Rename PhantomData wrapper and add comment for clarity

bjz:
When reading this code I had trouble understanding what `PhantomMap` was for. Hopefully this makes it clearer. I also made it private, and added some missing trailing commas.


nikomatsakis:
Thanks @bjz

PR #16, opened 2016-01-20
Add filter and filter_map combinators

bjz:
:warning: Do not merge - causes a segmentation fault :warning:

### DIFF
--- a/src/par_iter/filter.rs
+++ b/src/par_iter/filter.rs
@@ -1,0 +1,80 @@
+use super::{ParallelIterator, ParallelIteratorState, ParallelLen};
+use std::marker::PhantomData;
+
+pub struct Filter<M, FILTER_PRED> {
+    base: M,
+    pred: FILTER_PRED,
+}
+
+impl<M, FILTER_PRED> Filter<M, FILTER_PRED> {
+    pub fn new(base: M, pred: FILTER_PRED) -> Filter<M, FILTER_PRED> {
+        Filter { base: base, pred: pred }
+    }
+}
+
+impl<M, FILTER_PRED> ParallelIterator for Filter<M, FILTER_PRED>
+    where M: ParallelIterator,
+          FILTER_PRED: Fn(&M::Item) -> bool + Sync,
+{
+    type Item = M::Item;
+    type Shared = FilterShared<M, FILTER_PRED>;
+    type State = FilterState<M, FILTER_PRED>;
+
+    fn state(self) -> (Self::Shared, Self::State) {
+        let (base_shared, base_state) = self.base.state();
+        (FilterShared { base: base_shared, pred: self.pred },
+         FilterState { base: base_state, pred: PhantomFilterPred::new() })
+    }
+}
+
+pub struct FilterShared<M, FILTER_PRED>
+    where M: ParallelIterator,
+{
+    base: M::Shared,
+    pred: FILTER_PRED
+}
+
+pub struct FilterState<M, FILTER_PRED>
+    where M: ParallelIterator,
+{
+    base: M::State,
+    pred: PhantomFilterPred<FILTER_PRED>,
+}
+
+struct PhantomFilterPred<FILTER_PRED>(PhantomData<*const FILTER_PRED>);
+
+impl<FILTER_PRED> PhantomFilterPred<FILTER_PRED> {
+    fn new() -> PhantomFilterPred<FILTER_PRED> {
+        PhantomFilterPred(PhantomData)
+    }
+}
+
+unsafe impl<FILTER_PRED: Sync> Send for PhantomFilterPred<FILTER_PRED> { }
+
+impl<M, FILTER_PRED> ParallelIteratorState for FilterState<M, FILTER_PRED>
+    where M: ParallelIterator,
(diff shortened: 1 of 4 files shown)


bjz:
Despite using the safe API, this causes a segmentation fault when run. Here is the output from gdb:

```
(gdb) r par_iter::test::check_filter
Starting program: /Users/brendan/dev/rust/rayon/target/debug/rayon-e94ee2831ded576d par_iter::test::check_filter

running 1 test
[New Thread 0x1413 of process 95072]
[New Thread 0x1503 of process 95072]

Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x1413 of process 95072]
0x000000010001c954 in rayon::ops::&'a usize.Rem<usize>::rem (self=0x0, other=2) at ../src/libcore/ops.rs:128
128	../src/libcore/ops.rs: No such file or directory.
(gdb) bt
#0  0x000000010001c954 in rayon::ops::&'a usize.Rem<usize>::rem (self=0x0, other=2) at ../src/libcore/ops.rs:128
#1  0x000000010001cc26 in fnfn (x=0x0) at src/par_iter/test.rs:91
#2  0x000000010001ca34 in rayon::iter::Iterator::all<core::slice::Iter<&usize>,closure> (self=0x700000403bc8, f=...) at ../src/libcore/iter.rs:883
#3  0x0000000100018d1f in rayon::par_iter::test::check_filter () at src/par_iter/test.rs:91
#4  0x00000001000600fc in boxed::_$LT$impl$GT$::call_box::call_box::h2589288275602263322 ()
#5  0x0000000100062801 in sys_common::unwind::try::try_fn::try_fn::h15738982480174195438 ()
#6  0x00000001000858d9 in __rust_try ()
#7  0x000000010008384f in sys_common::unwind::try::inner_try::hcc3a84c80e36120fy9r ()
#8  0x0000000100062bfb in boxed::_$LT$impl$GT$::call_box::call_box::h11788775615639144818 ()
#9  0x00000001000872ae in sys::thread::_$LT$impl$GT$::new::thread_start::h5776bf5f72f331a7Rvw ()
#10 0x00007fff93167c13 in _pthread_body () from /usr/lib/system/libsystem_pthread.dylib
#11 0x00007fff93167b90 in _pthread_start () from /usr/lib/system/libsystem_pthread.dylib
#12 0x00007fff93165375 in thread_start () from /usr/lib/system/libsystem_pthread.dylib
#13 0x0000000000000000 in ?? ()
```

I'm guessing this is because I just copied `MapState`'s `ParallelIteratorState::len` impl, meaning that it doesn't match the filtered length. It's a little concerning that the API is leaking unsafety out of the module boundary, polluting safe code. Not sure of a way around it though.

bjz:
So I tried keeping a count of the items I skipped in `FilterState` in order to figure out the size relative to the `base.len`, but I am having trouble implementing `split_at`. This is because I need to figure out how to split the filtered count. Any ideas on ways to get around this?

nikomatsakis:
Probably I should declare the functions in the base trait unsafe. Indeed though, filtering is not going to be quite this simple I think. It requires deeper integration with the collect/reduce ops.

nikomatsakis:
I was vaguely thinking of a design like this:

- add a new trait called `ExactParallelIterator: ParallelIterator` that corresponds to unfiltered things
- move e.g. `enumerate` to this trait
- modify `collect` so that it tracks, for each subslice, what number of items were created, and then basically does a "compact" operation (this operation can itself be parallelized) if needed to remove empty space
     - we also have to at some point handle panics here more gracefully, though I guess leaking a bunch of stuff, while suboptimal, is legal :)
- `reduce` doesn't really have to change
- have the `filter` op bump of the `weight` in its `len` routine to encourage finer-grained splitting, since the amount of work will be less predictable

Still, `flat_map` is a bit more complicated. Initially, I expected to add `flatten` as a separate, terminating step, and then have people do `map` and subsequently `flatten`. Part of my thought process was that I wanted an invariant that any chain of method calls you write corresponds to parallel execution, and I don't think it's possible to collect from a `flat_map` without introducing a parallel barrier. But it's totally possible to `reduce`, and it would be possible to `collect` into e.g. a rope-like data structure. So perhaps we want another iterator trait to distinguish the filter and flat-map case. (Specialization might be nice here, but I suspect we can do without it.)

nikomatsakis:
The three cases would then be:

- `ArbitraryParallelIterator` -- produces an arbitrary number of items
- `BoundedParallelIterator` -- produces at most N items (filtered)
- `ExactParallelIterator` -- produces N items

We also have to toy with the traits we expose to users a bit. I want people to be able to add something like:

```
use rayon::par_iter::*;
```

and get just a few traits in scope, in particular one that let's them replace calls to `iter()` with `par_iter()`, `par_iter_mut()`, that sort of thing -- mirror how iterators are used. This is not how it works today. :(

nikomatsakis:
OK, I suspect I ought to transcribe some of that into an issue :)

bjz:
Great! Yeah, I got the feeling that the API as it stands would have a hard time supporting filtering. It's good to try to push it though, to see what breaks!

nikomatsakis:
Superceded by #24 -- though I didn't yet add `filter_map`

PR #23, opened 2016-01-23
Par iter and par iter mut

nikomatsakis:
This branch rearranges the `par_iter` module somewhat and closes various relatively simple issues. In terms of how `par_iter` is used, it adds various traits, and the new recommend pattern is as follows:

```rust
use rayon::par_iter::*; // get all the relevant traits
```

and then you can write `vec.par_iter()` and `vec.par_iter_mut()` and that sort of thing as normal. Note that calling `into_par_iter` directly is not recommended, as I plan to implement it for `Vec<T>` eventually, which would cause your code to start moving out of the vector rather than iterating over references.

Other changes:

- add `for_each`
- change how `ParallelIteratorState` works to be external-iterator-like
- add `zip`
- add `weight_max` convenience function

cc @polyfractal @eddyb @bjz @kaksmet

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ integers, please see the [notes on atomicity](#atomicity).
 **WARNING:** Rayon is still in **experimental** status. It is probably
 not very robust and not (yet) suitable for production applications.
 In particular, the handling of panics is known to be terrible.  See
-nikomatsakis/rayon#10 for more details.
+[nikomatsakis/rayon#10](https://github.com/nikomatsakis/rayon/issues/10)
+for more details.
 
 ### Parallel Iterators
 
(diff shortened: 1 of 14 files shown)


nikomatsakis:
cc @cuviper

polyfractal:
:+1: This worked great for me. Thanks!

nikomatsakis:
I'm thinking of a different (more general) way to struct `ParallelIterator` now, but I think this is still a pretty good step forward. Going to merge. :)

cuviper:
I'm glad to see `par_iter_mut` -- I was about ready to try implementing for `&mut [T]` myself. :)

All is working well here!

PR #24, opened 2016-01-23
Implement filter and add notion of sparse iterators

nikomatsakis:
@bjz here is an implementation of filter that I believe to be correct. I also refined the `ParallelIterator` hierarchy and fixed a few minor things.

I think a good next step would be to add `flat_map`, but also to generalize `collect` so that it works for collections of arbitrary types. In particular, one could add a rope-like version of vector that DOES work for filtered (or flat-mapped) iterators.

Note that this builds on https://github.com/nikomatsakis/rayon/pull/23.

### DIFF
--- a/src/par_iter/slice_mut.rs
+++ b/src/par_iter/slice_mut.rs
@@ -1,4 +1,4 @@
-use super::{ParallelIterator, IntoParallelIterator, IntoParallelRefMutIterator};
+use super::*;
 use super::len::ParallelLen;
 use super::state::ParallelIteratorState;
 use std::mem;
@@ -34,6 +34,10 @@ impl<'r, T: Send> ParallelIterator for SliceIterMut<'r, T> {
         ((), self)
     }
 }
+
+unsafe impl<'r, T: Send> BoundedParallelIterator for SliceIterMut<'r, T> { }
+
+unsafe impl<'r, T: Send> ExactParallelIterator for SliceIterMut<'r, T> { }
 
 unsafe impl<'r, T: Send> ParallelIteratorState for SliceIterMut<'r, T> {
     type Item = &'r mut T;
--- a/src/par_iter/slice_mut.rs
+++ b/src/par_iter/slice_mut.rs
@@ -3,30 +3,30 @@ use super::len::ParallelLen;
 use super::state::ParallelIteratorState;
 use std::mem;
 
-pub struct SliceIterMut<'r, T: 'r + Send> {
-    slice: &'r mut [T]
+pub struct SliceIterMut<'data, T: 'data + Send> {
+    slice: &'data mut [T]
 }
 
-impl<'r, T: Send> IntoParallelIterator for &'r mut [T] {
-    type Item = &'r mut T;
-    type Iter = SliceIterMut<'r, T>;
+impl<'data, T: Send> IntoParallelIterator for &'data mut [T] {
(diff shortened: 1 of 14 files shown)


nikomatsakis:
Note: flat map is maybe a bit trickier, in that I would expect to use parallelism recursively within the `next` method. Have to think about it just a bit.

nikomatsakis:
Ah, I think changing `ParallelIteratorState` to always use `next` was a mistake. Rather, that should be an add'l API offered only by `ExactParallelIterator` types. That way, flat-map can be implemented using the `for_each` callback, which I think should be relatively simple. Or at least that's my going theory -- it might not work out -- unfortunately I have to stop thinking about this now.

bjz:
Looks great! :)

src/par_iter/slice_mut.rs
```diff
@@ -0,0 +1,67 @@
+use super::*;
+use super::len::ParallelLen;
+use super::state::ParallelIteratorState;
+use std::mem;
+
+pub struct SliceIterMut<'r, T: 'r + Send> {
```
bjz: What method are you using to name these lifetimes? I thought `'a` was the standard convention?

src/par_iter/zip.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    fn next(&mut self, shared: &Self::Shared) -> Option<Self::Item> {
+        let a_next = self.a.next(&shared.a);
+        let b_next = self.b.next(&shared.b);
+        if let Some(a_item) = a_next {
+            if let Some(b_item) = b_next {
+                return Some((a_item, b_item));
+            }
+        }
+        None
```
bjz:
```rust

match (a_next, b_next) {
    (Some(a), Some(b)) => Some((a, b)),
    (_, _) => None,
}
```
Perhaps?

src/par_iter/slice_mut.rs
nikomatsakis: I forget why I chose `'r`. Usually I prefer to choose meaningful words, actually, and quite dislike the `'a` "convention" except for small, random lifetimes specific to one fn signature. I guess I was just lazy.

src/par_iter/zip.rs
nikomatsakis: Perhaps, I didn't want to create a tuple just to match it. It may be that LLVM optimizes that to the `if let` chain above, not sure.

nikomatsakis:
@bjz ok updated name to `'data` for more readability; and I added `filter_map`, just because.

nikomatsakis:
I'm going to merge this PR, but I want to experiment now with alternative "impl traits" for parallel iterator -- I think I have the kernel of an idea how to make this more flexible, e.g., to support flat map, and hopefully reduce some code duplication, but I'm not sure yet.

nikomatsakis:
Rebased over the range PR. As I wrote above, I think I still want to do further tuning here, but this setup is closer to the eventual goal anyway, and so I'll merge it -- also, it's useful to have `filter` and `filter_map` available, for sure (though it'd be nice to get `collect` working for them eventually; I imagine it will require, for vecs at least, a separate compression step).

(For the time being of course you can always replace `filter` with a `map` to option, then do the compression step yourself (`result.into_iter().flat_map(|x| x).collect()`), but I'd rather build it into the collect operation, which can do it in parallel and more efficiently.)

PR #25, opened 2016-01-24
impl ParallelIterator for Range<T> (for some T)

cuviper:
A par_iter/range.rs was already started, but not actually pulled in.

Range<T> is annoying to deal with generically, since its Iterator
depends on unstable Step, and its ExactSizeIterator is only for select
integral types (all but i64/u64).

Use a macro to implement just those Range<T> with ExactSizeIterator.

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -31,6 +31,7 @@ pub mod state;
 pub mod map;
 pub mod weight;
 pub mod zip;
+pub mod range;
 
 #[cfg(test)]
 mod test;
--- a/src/par_iter/range.rs
+++ b/src/par_iter/range.rs
@@ -1,57 +1,66 @@
-use super::{IntoParallelIterator, ParallelIterator, ParallelIteratorState, ParallelLen};
+use super::{IntoParallelIterator, ParallelIterator};
+use super::len::ParallelLen;
+use super::state::ParallelIteratorState;
 use std::ops::Range;
 
 pub struct RangeIter<T> {
-    slice: Range<T>
+    range: Range<T>
 }
 
-pub trait ParallelRange: Eq + Send {
-    fn len(start: Self, stop: Self) -> usize;
-    fn mid(start: Self, stop: Self) -> Self;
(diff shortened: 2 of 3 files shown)


cuviper:
(Irrelevant side note, the lowercase-l next to uppercase-I in `ParallelIterator` is pretty hard to read here on GitHub.  Thankfully my monospace font tries harder to distinguish those...)

nikomatsakis:
Oh, nice, I was wanting to do this.

lilianmoraru:
@nikomatsakis After this, could we please also get a version update on crates.io?
There are many new useful features.

cuviper:
I got paranoid looking at the split computation, and realized a way that could overflow with a large index.  This will never actually happen with the current callers that just use `len/2`, but it's technically possible and easy to fix.  Added a new commit to use `wrapping_add`, with an explicit test.

nikomatsakis:
Thanks @cuviper. Just FYI, I've been working on a revamping of the whole `ParallelIteratorState` business, in order to allow for `flat_map`, but I still don't *quite* have it right. (The trick is you have to allow for parallel iterators like `flat_map` to drive the process, since they might need to produce new splits.) Anyway, I'll merge this in the meantime, since it'll be easy to rebase over.

nikomatsakis:
@lilianmoraru yes, I will push a new version very soon.

nikomatsakis:
Actually, looking at this more closely, @cuviper is there a reason we didn't just implement `ParallelIterator` for `Range` directly?

nikomatsakis:
Experimenting with it myself, it seems like this causes type inference failures because of the lack of a single, generic impl. Annoying. We could reproduce the functionality of `Step`, I guess, just to dodge the stability problem.

cuviper:
If we didn't have a new type, wouldn't that also make most iterator calls ambiguous?
e.g. `(1..10).map(/*...*/)`, is that calling `Iterator::map` or `ParallelIterator::map`?

nikomatsakis:
@cuviper yes, I realized this later and meant to leave a comment to that effect :)

PR #27, opened 2016-01-26
Implement a new parallel iterator scheme that supports flat-map

nikomatsakis:
This commit history is kind of rambly but I don't have time to clean it up. The big bit of work here is support for flat-map, which required a lot of redesign. See the `par_iter/README.md` file for an overview. 

I think there is still tweaking to be done on these traits. For example, I'd like to reduce the unsafety, and maybe add some lifetimes to allow passing references for the shared state in the producer. But this will do for now.

cc @cuviper @bjz

### DIFF
--- a/src/par_iter/range.rs
+++ b/src/par_iter/range.rs
@@ -74,7 +74,6 @@ macro_rules! range_impl {
         impl Producer for RangeIter<$t> {
             type Item = $t;
             type Shared = ();
-            type SeqState = ();
 
             unsafe fn split_at(self, index: usize) -> (Self, Self) {
                 assert!(index <= self.range.len());
@@ -86,14 +85,8 @@ macro_rules! range_impl {
                 (RangeIter { range: left }, RangeIter { range: right })
             }
 
-            unsafe fn start(&mut self, _: &()) {
-            }
-
-            unsafe fn produce(&mut self, shared: &(), state: &mut ()) -> $t {
+            unsafe fn produce(&mut self, shared: &()) -> $t {
                 self.range.next().unwrap()
-            }
-
-            unsafe fn complete(self, _: &(), _: ()) {
             }
         }
     }
--- a/src/par_iter/range.rs
+++ b/src/par_iter/range.rs
@@ -23,8 +23,10 @@ macro_rules! range_impl {
             type Shared = ();
             type State = Self;
 
-            fn drive<C: Consumer<Item=Self::Item>>(self, consumer: C, shared: C::Shared)
-                                                   -> C::Result {
+            fn drive<'c, C: Consumer<'c, Item=Self::Item>>(self,
+                                                           consumer: C,
+                                                           shared: &'c C::Shared)
+                                                           -> C::Result {
                 bridge(self, consumer, &shared)
             }
 
--- a/src/par_iter/range.rs
+++ b/src/par_iter/range.rs
@@ -70,6 +70,32 @@ macro_rules! range_impl {
                 self.range.next()
             }
         }
+
+        impl Producer for RangeIter<$t> {
+            type Item = $t;
+            type Shared = ();
+            type SeqState = ();
+
+            unsafe fn split_at(self, index: usize) -> (Self, Self) {
+                assert!(index <= self.range.len());
+                // For signed $t, the length and requested index could be greater than $t::MAX, and
+                // then `index as $t` could wrap to negative, so wrapping_add is necessary.
+                let mid = self.range.start.wrapping_add(index as $t);
+                let left = self.range.start .. mid;
+                let right = mid .. self.range.end;
+                (RangeIter { range: left }, RangeIter { range: right })
+            }
+
+            unsafe fn start(&mut self, _: &()) {
+            }
+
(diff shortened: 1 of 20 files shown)


src/par_iter/range.rs
```diff
@@ -20,44 +19,62 @@ macro_rules! range_impl {
...
+                                                           consumer: C,
+                                                           shared: &'c C::Shared)
+                                                           -> C::Result {
+                bridge(self, consumer, &shared)
+            }
         }
 
         unsafe impl ExactParallelIterator for RangeIter<$t> {
+            fn len(&mut self) -> usize {
+                self.range.len() as usize
```
cuviper: Unnecessary cast.

src/par_iter/slice.rs
```diff
@@ -26,40 +25,65 @@ impl<'data, T: Sync + 'data> IntoParallelRefIterator<'data> for [T] {
...
+                                                   shared: &'c C::Shared)
+                                                   -> C::Result {
+        bridge(self, consumer, &shared)
     }
 }
 
-unsafe impl<'data, T: Sync> BoundedParallelIterator for SliceIter<'data, T> { }
+unsafe impl<'data, T: Sync> ExactParallelIterator for SliceIter<'data, T> {
+    fn len(&mut self) -> usize {
+        self.slice.len() as usize
```
cuviper: Unnecessary cast.

src/par_iter/slice_mut.rs
```diff
@@ -27,41 +26,66 @@ impl<'data, T: Send + 'data> IntoParallelRefMutIterator<'data> for [T] {
...
+                                                   shared: &'c C::Shared)
+                                                   -> C::Result {
+        bridge(self, consumer, &shared)
     }
 }
 
-unsafe impl<'data, T: Send> BoundedParallelIterator for SliceIterMut<'data, T> { }
+unsafe impl<'data, T: Send> ExactParallelIterator for SliceIterMut<'data, T> {
+    fn len(&mut self) -> usize {
+        self.slice.len() as usize
```
cuviper: Unnecessary cast.

src/par_iter/slice.rs
```diff
@@ -26,40 +25,65 @@ impl<'data, T: Sync + 'data> IntoParallelRefIterator<'data> for [T] {
...
+    }
+}
 
-unsafe impl<'data, T: Sync> ExactParallelIterator for SliceIter<'data, T> { }
+impl<'data, T: Sync> IndexedParallelIterator for SliceIter<'data, T> {
```
cuviper: I need `T: 'data` for this to compile on stable, which makes sense since `SliceProducer` requires it.  Odd that rust nightly doesn't care though...

src/par_iter/slice_mut.rs
```diff
@@ -27,41 +26,66 @@ impl<'data, T: Send + 'data> IntoParallelRefMutIterator<'data> for [T] {
...
+    }
+}
 
-unsafe impl<'data, T: Send> ExactParallelIterator for SliceIterMut<'data, T> { }
+impl<'data, T: Send> IndexedParallelIterator for SliceIterMut<'data, T> {
```
cuviper: Ditto needing `T: 'data`.

cuviper:
Holy rewrite, Batman!  The description in the readme is helpful, and though the concepts are more challenging now, it sounds pretty reasonable.

I don't have time to dig through this in detail right now, but I did notice a few nits, commented inline.

And FWIW, a program I have using range got about 10% faster, so that's nice. :)

nikomatsakis:
@cuviper 

> And FWIW, a program I have using range got about 10% faster, so that's nice. :)

Hey, score! :)

I had some thoughts on the subway ride home about how to simplify things a bit (remove some lifetimes etc). I'll try to experiment it with it next chance I get.

src/par_iter/slice.rs
nikomatsakis: @cuviper 

> I need T: 'data for this to compile on stable, which makes sense since SliceProducer requires it. Odd that rust nightly doesn't care though...

Thanks. I think this is because of RFC 1214, which allows impls to infer that `T: 'data` because `SliceIter` required it (that code only became finalized in nightly).

src/par_iter/range.rs
nikomatsakis: Yeah, I originally was using `u64` but got sick of casting and decided to go with `usize` for now. Not clear if that is the right call though.

nikomatsakis:
OK, so, I thought that I could remove the `Shared` assoc type from `Consumer` and instead just thread the shared state with an `&` reference embedded in the `Consumer` itself. This would have avoided the need for the `'consume`, I think, and also just simplified the fn signatures and so forth. But this idea, sadly, does not work because of the `reduce` method, which combines two items after a split -- at that point, you no longer have a `Consumer` to use, so you need to have the shared state separated out.

nikomatsakis:
OK, I'm going to merge this. I have some more edits I'll maybe open in another PR, but this is a good point to get locked down. :)

bjz:
:+1:

PR #29, opened 2016-01-26
Limit number of  threads Rayon is allowed to use

willi-kappler:
- This allows the user to specify a maximum number of threads that Rayon is allowed to use.

- The user can call the `rayon::initialize(max_num)` function to specify the limit. This breaks the API since this function didn't expect any arguments before. This function should have been used for benchmarking only, so the number of affected users should be small.

- Alternatively a new initialize function could have been created,  for example `rayon::initialize_limit(max_num)` and no API would be broken.

- If the user does not call `rayon::initialize(max_num)` the old behavior of using all the cores available is given.

- If the user specifies 0 as limit, Rayon ends up in a infinite loop. Thus an `assert!` checks the user input. Alternatively instead of a panic Rayon could just use the all the cores available, but this has the disadvantage that the user in not informed about the erroneous input.


src/api.rs
```diff
@@ -4,13 +4,19 @@ use log::Event::*;
...
 
 /// Initializes the Rayon threadpool. You don't normally need to do
 /// this, as it happens automatically, but it is handy for
 /// benchmarking purposes since it avoids initialization overhead in
 /// the actual operations.
-pub fn initialize() {
-    let registry = thread_pool::get_registry();
+/// But you need to call this if you want to specify the maximum number
+/// of threads that Rayon is allowed to use.
+/// If you don't specify it, Rayon uses all the cores avaiable (determined
```
nikomatsakis: Nit: s/avaiable/available (misspelled)

nikomatsakis: I think I would prefer if there was an easy way to get the default behavior; for example, this could take `Option<u32>`, or else we could say that if you say 0, we will autodetect the number of CPUs.

nikomatsakis: My other concern here is that if you call this function *after* a thread-pool has been created (for whatever reason), or if you call it twice, then the value of `num_threads` will be ignored. It's kind of a pain to detect this, though -- so maybe it's enough to just document it.

src/api.rs
```diff
@@ -85,7 +91,7 @@ unsafe fn join_inject<A,B,RA,RB>(oper_a: A,
     let mut latch_b = LockLatch::new();
     let mut job_b = Job::new(&mut code_b, &mut latch_b);
 
-    thread_pool::get_registry().inject(&[&mut job_a, &mut job_b]);
+    thread_pool::get_registry(num_cpus::get()).inject(&[&mut job_a, &mut job_b]);
```
nikomatsakis: (If we had an easy way to specify the default value, I'd prefer that we use that here rather than hardcode `num_cpus::get()` -- or rather, this seems like the wrong place to hardcode that, I'd prefer it's hardcoded in the `get_registry` routine itself, which seems more central.)

nikomatsakis:
Thanks @willi-kappler! I left some comments. Let me know what you think. I'm not worried about breaking the API -- this is still in the `0.x` series, so I think people have to expect some minor adjustments. However, I would like it to be easy for users to get a "default number of threads", rather than being forced to specify a particular number.

lilianmoraru:
I think it's nicer to have the default behavior to use everything it can.
Also, Rust APIs seem to already have a style for this kind of situation: `with_capacity`, `with_len`...
Seems confusing to pass `0`. You will probably see users pass `std::usize::MAX`.

cuviper:
If this is just an upper bound, then internally it should use `cmp::min(num_cpus::get(), max_num)`.

If this is an exact number of threads to use, then don't call it a "limit" or "max".  But I think it rarely makes sense to oversubscribe threads.  Maybe if you know some of the jobs will sleep, e.g. for IO, but rayon is not a great API for blocking work like that.

willi-kappler:
I like `Option<>`, zero seems strange a value for default (`num_cpus::get()`). And @cuviper is right: internally `cmp:min()` should be used if s.o. passes `std::usize::MAX` as @lilianmoraru mentioned.

There is still the issue how to handle zero: I don't like panics in libraries, but imagine a developer uses Rayon for his / her GUI application, where there is an option to specify the number of threads to be used. If the GUI application has a bug and the number of threads gets lost and instead zero is passed to Rayon, the developer could have a hard time to find the reason why Rayon doesn't behave like expected.

The `assert!` will crash the application at runtime but at least the developer gets a notification.

An alternative would be if `initialize()` returns the status (fail or ok) but the developer has to handle this.

nikomatsakis:
Another thought I had was that it might be good to have some kind of struct you can pass in, so that we can be prepared for other options in the future. e.g., `rayon::Configuration::new().set_num_threads(4).initialize();` I don't want to go crazy over-designing things though :) especially since I want Rayon to have as few knobs as possible. Not sure what is the established "best practice" here (if any). 

I'm also not sure if it makes sense to cap the number of threads at the number of cores. There are certainly workloads that perform better when mildly oversubscribed.

As far as surprises go, perhaps what worries me the most is that if you call `initialize` twice, the second one is ignored. I was thinking that we could fix this by having `initialize` check the number of threads that the current thread-pool has and assert that they are as expected. (This way, if an earlier call already initialized it to the wrong number, you will find out.)

nikomatsakis:
@willi-kappler looks like there are travis failures in the demo.

willi-kappler:
Oversubscribing slightly makes sense.

The builder pattern for configuration looks like a nice idea. No need for `Option<>` and error handling of input values could be done in the last initialize() step.

nikomatsakis:
One final thought: it'd be good to add some tests that show what happens in the various error cases. Since this is tinkering with global configuration, I expect we'd want to add those using the compiletest harness, but we'd have to setup the `run-fail` mode. (The compiletest harness is a port of the testing system we use in the compiler; it actually runs distinct processes, rather than just starting different threads like `#[test]` does.) I don't have that setup, but I may set it up for another piece of work I am doing anyway...

lilianmoraru:
Couldn't we have it return a `Result`? Or at least use `debug_assert`?
I detest libraries that decide by themselves to crash my application.
If somebody receives a `Result`, they can handle it at least, and most of the times makes them think what they have to handle(or unwrap() and have the same behavior as having assert).

cuviper:
I agree libraries shouldn't "decide by themselves" to crash, but it's not that unusual to crash on bad input.  Look at Vec, for instance, and all the many ways it will panic on you.  If there's nothing reasonable to ever do with threads=0, then I say document that and panic.

*Maybe* it could mean something though, like threads=0 meaning don't use a thread pool at all, just run serially in the current thread, no matter what.

nikomatsakis:
I agree it'd be better to return a `Result` (both if `0` is provided, and if initialization has already happened with distinct options).

willi-kappler:
Yes, returning a Result (or even a custom error type) is better than a panic. And it also fits nicely into the builder pattern. I'll change the code accordingly.

I can also take a look into the test cases (at least the non-failing ones).

willi-kappler:
Hi @nikomatsakis,
this is a first try. Before I add the documentation, is this s.th. you had in mind ?
I'm not sure about the return type of the builder pattern `initialize()` step: Since we don't want to return the registry, `Result` would have been empty. So I've used a custom enum.

For the `ThreadPool` Result makes more sense. In the `Err()` case it returns one of the custom enum variants. Is this idiomatic ?

@cuviper threads=0 for single (main) threading seems too complicated right now. At least it looks like I have to change a lot of code.

cuviper:
No problem, I admit that was a weird idea anyway. :)

nikomatsakis:
@willi-kappler I will take a look a bit later, but I'd probably use `Result<(), ErrorCode>` vs a custom enum. That way, it fits better with `try!`, and benefits from all the `Result` methods like unwrap etc.

willi-kappler:
@nikomatsakis No need to rush ;-)
Yes of course, that makes sense. I'll change it, sorry!

src/api.rs
```diff
@@ -5,13 +5,54 @@ use job::{Code, CodeImpl, Job};
...
+
+        let registry = thread_pool::get_registry(self.num_threads);
+
+        if let Some(value) = self.num_threads {
+            if value != registry.num_threads() {
+                return Err(InitResult::NumberOfThreadsNotEqual);
+            }
+        }
+
+        if self.bench {
```
nikomatsakis: It might be better to just wait unconditionally. It's not going to be a long wait, I expect.

src/api.rs
```diff
@@ -5,13 +5,54 @@ use job::{Code, CodeImpl, Job};
...
+    pub bench: bool,
+    pub num_threads: Option<usize>
+}
+
+impl Configuration {
+    pub fn new() -> Configuration {
+        Configuration { bench: false, num_threads: None }
+    }
+
+    pub fn set_num_threads(mut self, num_threads: usize) -> Configuration {
```
nikomatsakis: We should add a doc comment here. Something like

```
/// Set the number of threads to be used in the Rayon threadpool.
/// The argument `num_threads` must not be zero. If you do not call this function,
/// Rayon will select a suitable default (currently, the default is one thread per CPU).
```

src/api.rs
```diff
@@ -98,10 +139,20 @@ pub struct ThreadPool {
...
 impl ThreadPool {
-    pub fn new() -> ThreadPool {
-        ThreadPool {
-            registry: Registry::new()
+    /// Constructs a new thread pool
+    /// If you pass None as arguments, the current number of cores is used
+    /// as determinde via num_cpus::get()
+    /// This function returns Error() if the number of threads is zero
+    /// otherwise it returns Ok()
+    pub fn new(num_threads: Option<usize>) -> Result<ThreadPool,InitResult> {
```
nikomatsakis: It's fine for now, but eventually I would expect this to take a `Configuration` as its arguments.

nikomatsakis:
@willi-kappler this is looking good, thanks for refactoring it to the builder interface!

What do you think about removing the `bench` method and just making it wait unconditionally?

Other than that, I'd say it just needs a doc comment and it's good to go.

willi-kappler:
@nikomatsakis Thanks! I'll remove the `bench` method, add the documentation and add `Configuration` to `ThreadPool`.

lilianmoraru:
I actually prefer more the initial `rayon::initialize()`(or `rayon::init()`) style because I think the default behavior(to use everything) will be used almost exclusively and it seems easier to remember.
But I guess it's about choice of style...

nikomatsakis:
@lilianmoraru the main thing is that you actually do not really need to call `initialize` *unless* you want some configuration. Even in benchmarks I don't know that you can observe the difference, actually. That said, having an `initialize()` function that is equivalent to

nikomatsakis:
See https://github.com/nikomatsakis/rayon/pull/38

willi-kappler:
@nikomatsakis No need to worry ;-) Even with such a small patch I learned some things!

nikomatsakis:
@willi-kappler ok, please do take a look at #38 and let me know if it all looks good to you. :) I'll close this PR for now though.

lilianmoraru:
@nikomatsakis Oh, ok, I see. For some reason I was thinking that you need to use `initialize`

Issue #17, opened 2016-01-20
Support moving into paralell iterators


bjz:
At the moment we only have an impl on `&[T]`. This makes it expensive to do anything on large objects - you need to clone, especially if you want to reduce.

A possible way could be to pull apart a `Vec` by way of unsafe code. We could use `split_off`, but that would be expensive for large n-sizes.

PR #30, opened 2016-01-27
Restructure producer to use a callback.

nikomatsakis:
This change was a bit of a pain, but makes sense conceptually, and should enable "moving" iterators quite gracefully (e.g., `Vec::into_par_iter`). The idea is to change `fn into_producer(self)` to instead be:

```rust
    fn with_producer<WP: ProducerContinuation<Self::Item>>(self, wp: WP) -> WP::Output;
```

where `ProducerContinuation` is basically a callback with a specific continuation:

```rust
pub trait ProducerContinuation<ITEM> {
    type Output;
    fn with_producer<'p, P>(self, producer: P, shared: &'p P::Shared) -> Self::Output
        where P: Producer<'p, Item=ITEM>;
}
```

This would be a closure, except that closures can't be generic in the way that `with_producer` is monomorphic. That is, to use a closure, we'd have to specify a particular kind of producer, whereas with the trait, the producer type is not known to the continuation. It's sort of the difference between having `T: Fn(SomeType)` and `T: for<P> Fn(P)` (the latter is not syntax we support).

Anyway, this requires rewriting some really simple code into really horrific explicit continuations. However, the benefits are:

1. We can pass the shared state using only references, which means we never move the `map` closures and the like (before, we used to move them out of a structure and onto the stack, etc). I would expect this to be a perf-win, though how large I have no idea.
    - The reason this is possible is that the produce type `P` can now contain local lifetimes.
2. Later, we can implement `into_par_iter` for `Vec<T>` relatively easily; I guess before it was possible if ownership of the vec was moved into the `Shared` type, but that was both unsafe and unfortunate, since it requires the moves I was complaining about above. Now that the processing of the producer takes place inside the `with_producer` call, we can handle the freeing ourselves more easily.

### DIFF
--- a/src/par_iter/enumerate.rs
+++ b/src/par_iter/enumerate.rs
@@ -64,16 +64,16 @@ impl<M> IndexedParallelIterator for Enumerate<M>
             where CB: ProducerCallback<(usize, ITEM)>
         {
             type Output = CB::Output;
-            fn with_producer<'produce, P>(self,
-                                          base: P,
-                                          shared: &'produce P::Shared)
-                                          -> CB::Output
+            fn callback<'produce, P>(self,
+                                     base: P,
+                                     shared: &'produce P::Shared)
+                                     -> CB::Output
                 where P: Producer<'produce, Item=ITEM>
             {
                 let producer = EnumerateProducer { base: base,
                                                    offset: 0,
                                                    phantoms: PhantomType::new() };
-                self.callback.with_producer(producer, shared)
+                self.callback.callback(producer, shared)
             }
         }
     }
--- a/src/par_iter/enumerate.rs
+++ b/src/par_iter/enumerate.rs
@@ -50,11 +50,27 @@ unsafe impl<M> ExactParallelIterator for Enumerate<M>
 impl<M> IndexedParallelIterator for Enumerate<M>
     where M: IndexedParallelIterator,
 {
-    type Producer = EnumerateProducer<M::Producer>;
+    fn with_producer<WP>(self, wp: WP) -> WP::Output
+        where WP: ProducerContinuation<Self::Item>
+    {
+        return self.base.with_producer(Continuation { wp: wp });
 
-    fn into_producer(self) -> (Self::Producer, <Self::Producer as Producer>::Shared) {
-        let (base, shared) = self.base.into_producer();
-        (EnumerateProducer { base: base, offset: 0 }, shared)
+        struct Continuation<WP> {
+            wp: WP,
+        }
+
+        impl<ITEM, WP> ProducerContinuation<ITEM> for Continuation<WP>
+            where WP: ProducerContinuation<(usize, ITEM)>
+        {
+            type Output = WP::Output;
+            fn with_producer<P:Producer<Item=ITEM>>(self,
+                                                    base: P,
+                                                    shared: P::Shared)
+                                                    -> WP::Output {
+                let producer = EnumerateProducer { base: base, offset: 0 };
+                self.wp.with_producer(producer, shared)
+            }
+        }
     }
 }
 
--- a/src/par_iter/enumerate.rs
+++ b/src/par_iter/enumerate.rs
@@ -51,29 +51,29 @@ unsafe impl<M> ExactParallelIterator for Enumerate<M>
 impl<M> IndexedParallelIterator for Enumerate<M>
     where M: IndexedParallelIterator,
 {
-    fn with_producer<WP>(self, wp: WP) -> WP::Output
-        where WP: ProducerContinuation<Self::Item>
+    fn with_producer<CB>(self, callback: CB) -> CB::Output
+        where CB: ProducerCallback<Self::Item>
     {
-        return self.base.with_producer(Continuation { wp: wp });
+        return self.base.with_producer(Callback { callback: callback });
 
-        struct Continuation<WP> {
-            wp: WP,
+        struct Callback<CB> {
+            callback: CB,
         }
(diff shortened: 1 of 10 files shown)


nikomatsakis:
cc @cuviper -- if you have benchmarks, I'd be curious to see if this change has any affect on them. Also, it's definitely time to start adding benchmarks in the Rayon repo proper, since I'm kind of flying blind on the effects of these changes.

bjz:
cc. #17

src/par_iter/enumerate.rs
```diff
@@ -50,29 +51,50 @@ unsafe impl<M> ExactParallelIterator for Enumerate<M>
 impl<M> IndexedParallelIterator for Enumerate<M>
     where M: IndexedParallelIterator,
 {
-    type Producer = EnumerateProducer<M::Producer>;
+    fn with_producer<WP>(self, wp: WP) -> WP::Output
```
bjz: Would `producer: P` be clearer here?

bjz: Or maybe I am mis-understanding. Anyway, it would be nice if the parameter name was more explicit.

src/par_iter/enumerate.rs
```diff
@@ -50,29 +51,50 @@ unsafe impl<M> ExactParallelIterator for Enumerate<M>
...
 
-pub struct EnumerateProducer<M>
-    where M: Producer,
+pub struct EnumerateProducer<'p, P>
+    where P: Producer<'p>,
 {
-    base: M,
+    base: P,
     offset: usize,
+    phantoms: PhantomType<&'p ()>,
```
bjz: Does this need to be `PhantomType`, or would `PhantomData` suffice?

src/par_iter/internal.rs
```diff
@@ -7,14 +7,20 @@ use join;
 use super::IndexedParallelIterator;
 use super::len::*;
 
+pub trait ProducerContinuation<ITEM> {
+    type Output;
+    fn with_producer<'p, P>(self, producer: P, shared: &'p P::Shared) -> Self::Output
+        where P: Producer<'p, Item=ITEM>;
+}
+
```
bjz: So this trait and `Producer` are in the `internal` module - does that mean that clients would not need to impl their own `Producer`s if they wanted to create custom parallel iterators?

src/par_iter/internal.rs
```diff
@@ -7,14 +7,20 @@ use join;
 use super::IndexedParallelIterator;
 use super::len::*;
 
+pub trait ProducerContinuation<ITEM> {
```
bjz: Is [this what you mean](https://en.wikipedia.org/wiki/Continuation) when you say 'continuation'? Might be handy to have a comment here.

lilianmoraru:
@nikomatsakis 

> if you have benchmarks, I'd be curious to see if this change has any effect on them.

I don't know if you've seen these: https://users.rust-lang.org/t/new-version-of-mandel-rust-uses-rayon-added-benchmark/4403

cuviper:
The code I've got is mostly for Project Euler, and in the spirit of that project I choose not to share those solutions.  Maybe I'll see if some of those can be abstracted into reasonable benchmarks though.  Anyway, performance with this PR looks about the same.

Those mandelbrot results are looking pretty good for rayon. :)

src/par_iter/enumerate.rs
nikomatsakis: No, the `wp` here is the continuation. I should call it `continuation`, I think.

src/par_iter/enumerate.rs
nikomatsakis: Doesn't really matter here, but I used `PhantomType` because this does not represent data that is reachable which the compiler can't see.

src/par_iter/internal.rs
nikomatsakis: No, you would have to do so. The idea of the `internal` module was that this is code you only need *if* you are writing custom parallel iterators. Maybe it needs a better name.

src/par_iter/internal.rs
nikomatsakis: It is roughly what I mean. :) I could perhaps just use the term "callback", which is more accurate I guess and more common.

nikomatsakis:
@lilianmoraru thanks, I had seen those, and I was planning on incorporating that code.

nikomatsakis:
I was also wondering if I should have two names, one for the `with_producer` on parallel iterators, and one for the `with_producer` in the `ParameterCallback`

src/par_iter/internal.rs
nikomatsakis: (I think I also called it `internal` because I don't really want people to implement their own parallel iterators yet, until the interface calms down a bit.)

src/par_iter/enumerate.rs
nikomatsakis: As an aside, with this change, it would be possible to remove the `Shared` type from `Producer` and instead have each producer carry their own shared data internally as `&'p Foo` reference (copying it on each split). I started doing this but stopped for two reasons:

- It was kind of a pain and I'm sick of refactoring this code for a bit. :)
- It didn't seem to be making the code much simpler, somewhat surprisingly to me. Maybe a bit.
- It meant that each split had to copy more data, since instead of copying one shared reference, you copied N, one for each level. OTOH, to *access* the shared data, this version requires more dereferences. I have no idea which is better. Probably it doesn't make any difference at all.

nikomatsakis:
OK, going to land this PR.

PR #31, opened 2016-01-28
Keep rayon stable

cuviper:
Slice's IntoParallelIterator was missing a `T: 'data`, upsetting rust 1.6.

This also adds rust stable to the travis script.  We can't run tests on stable
since compiletest_rs requires nightly features, but we can at least build it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 language: rust
 rust:
+  - stable
   - nightly
 script:
-  - cargo test
+  - cargo build
+  - |
+    [ $TRAVIS_RUST_VERSION != nightly ] ||
+    cargo test
   - cd demo/quicksort
   - cargo build
--- a/src/par_iter/slice.rs
+++ b/src/par_iter/slice.rs
@@ -5,7 +5,7 @@ pub struct SliceIter<'data, T: 'data + Sync> {
     slice: &'data [T]
 }
 
-impl<'data, T: Sync> IntoParallelIterator for &'data [T] {
+impl<'data, T: Sync + 'data> IntoParallelIterator for &'data [T] {
     type Item = &'data T;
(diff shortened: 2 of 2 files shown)


bjz:
We could limit `compiletest_rs` to an `unstable` feature, and conditionally use that for the nightly build...?

.travis.yml
```diff
@@ -1,7 +1,11 @@
 language: rust
 rust:
+  - stable
   - nightly
 script:
-  - cargo test
+  - cargo build
+  - |
+    [ $TRAVIS_RUST_VERSION != nightly ] ||
+    cargo test
```
bjz: Could this just be on the same line, rather than needing the pipe block? ie. `[ $TRAVIS_RUST_VERSION != nightly ] || cargo test`

cuviper:
I was hoping to make `compiletest_rs` an optional dev-dependency, default enabled.  That way `cargo build` still wouldn't need it, but `cargo test` would default to everything yet allow opting out for stable runs.  Apparently cargo doesn't allow optional dev-dependencies though.

.travis.yml
cuviper: I did try that, but `travis lint` complains:

```
Warnings for .travis.yml:
[x] syntax error: (<unknown>): did not find expected comment or line break while scanning a block scalar at line 7 column 41
```
(column 41 being the start of `||`)

cuviper: re: inverting the condition, that leaves the command with a non-zero exit status, which will be read by travis as a failure.

nikomatsakis:
@cuviper ah, nice, thanks.

.travis.yml
bjz: Ah, makes sense! :+1:

PR #32, opened 2016-01-29
Add a few simple benchmarks

cuviper:
See each commit message for benchmark results on my i7-2600 (4 core, 8 threads)
running on Linux, Fedora 23 x86_64.

These aren't terribly interesting computations, nor are they really good demos
of rayon, as a lot of the results don't show good parallelization.  But they do
hammer on rayon pretty hard, so I hope they're still useful for comparing the
effects of any proposed rayon changes.

### DIFF
--- a/benches/fibonacci.rs
+++ b/benches/fibonacci.rs
@@ -1,0 +1,94 @@
+//! Benchmark Fibonacci numbers, F(n) = F(n-1) + F(n-2)
+//!
+//! Recursion is a horrible way to compute this -- roughly O(2ⁿ).
+//! 
+//! It's potentially interesting for rayon::join, because the splits are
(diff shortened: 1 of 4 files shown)


nikomatsakis:
Nice, thanks!

benches/fibonacci.rs
```diff
@@ -0,0 +1,94 @@
...
+        a + b
+    }
+
+    b.iter(|| assert_eq!(fib(test::black_box(N)), FN));
+}
+
+
+#[bench]
+/// Compute the Fibonacci number iteratively, just to show how silly the others
+/// are.  Parallelism can't make up for a bad choice of algorithm.
```
nikomatsakis: Heh :)

PR #34, opened 2016-01-30
Make parallel traits safe by making `collect` check for proper usage

nikomatsakis:
Remove pervasive unsafety and instead restructure `collect` consumer to monitor how it is being used and assert that producer is doing the right thing. I think this is not quite ready to land yet. Needs the following:

- [ ] Measurements of impact (which sort of needs good benchmarks)
- [ ] Some negative tests of evil producers that fail to uphold their end of the bargain in various ways; we would want to be sure that the `collect` consumer doesn't fail under those cases.
    - [ ] fail to produce correct number of items
    - [ ] produce too many items
    - [ ] abuse consumer protocol by splitting after starting consume phase
    - [ ] abuse consumer protocol by never calling `complete`
    - [ ] abuse consumer protocol by leaking calling
    - [ ] abuse consumer protocol by mixing up the shared state (*)
    - [ ] not sure what else

It might also be worth restructuring the consumer protocol so that the sequential phase is a "step transition" (e.g., `fn start_seq_phase(self) -> Self::Seq`). This would ensure statically that once sequential execution begins, the producer can no longer split the consumer, which would avoid one assertion in the `collect` code.

(*) Hmm, collect might be vulnerable to this actually, now that I think about it. It could be solved by moving the `Shared` data internal to the consumer, which might be a good change actually (I've also thought about it for the `Producer` trait). Maybe also by converting `Shared` from `AtomicUsize` to `&'p Shared`, since then the two `Shared` types would not be equal and hence I don't think you could write a producer that can mix them up.


nikomatsakis:
It occurs to me that the collect consumer could also use the `Producer` interface, which would be somewhat easier since it would be in control of *pulling* elements -- but I wanted to generalize it someday to support producers of unknown size. Have to think about it.

nikomatsakis:
This is now obviated by #47.

Issue #35, opened 2016-01-30
Benchmark suite


nikomatsakis:
We need to have a start at a standard Rayon benchmark suite. @cuviper got us started with <https://github.com/nikomatsakis/rayon/pull/32>. But we need a bit more.

- [ ] NBody -- I've started porting NBody from [the ParallelJS benchmarks][pjs] in the [nbody][nbody] branch.
- [ ] Mandelbrot -- @willi-kappler has [implemented Mandelbrot](https://github.com/willi-kappler/mandel-rust), we could freeze a version (I'd want to have something constant for comparisons).
- [ ] ProceduralTerrain from [the ParallelJS benchmarks][pjs]
- [ ] liquid-resize from [the ParallelJS benchmarks][pjs]
- [ ] A gigantic dot-product implemented using `par_iter`
- [ ] Naive matrix multiplication implemented using `par_iter` perhaps?

More suggestions or pointers welcome.

[pjs]: https://github.com/IntelLabs/RiverTrail/tree/master/examples

willi-kappler:
I think mandel-rust v0.4 should be ready for this (will be release soon). I plan to do some clean up and add script-collectable output. Should I remove the other crates and just leave Rayon for the benchmark suite?

Matklad wrote a [raytracer](https://github.com/matklad/rustraytracer) in Rust which could be added.
Maybe also audio en-/decoder (ogg, flac) and compression / decompression algorithms. IIRC I saw some of them written in 100% Rust.

There should always be a way to verify (automatically) that the results are correct (compare images, compressed files, etc.).

nikomatsakis:
@willi-kappler 

> I think mandel-rust v0.4 should be ready for this (will be released soon). I plan to do some clean up and add script-collectable output. Should I remove the other crates and just leave Rayon for the benchmark suite?

If you could prepare a kind of stripped down version to include as a benchmark/demo, that'd be great.

> Matklad wrote a raytracer in Rust which could be added.  Maybe also audio en-/decoder (ogg, flac) and compression / decompression algorithms. IIRC I saw some of them written in 100% Rust.

Good ideas. The raytracer should be imminently actionable.

> There should always be a way to verify (automatically) that the results are correct (compare images, compressed files, etc.).

Yes. I need to add this to the nbody demo that I wrote, though I'm not quite sure how to do it. I mean I guess I can just hardcode some results. (I was considering running the original JS code for some time and comparing the results, as well.)

lilianmoraru:
[...]

PR #36, opened 2016-01-31
Add an nbody demo

nikomatsakis:
This can serve as a benchmark. It also includes a (pretty crude) visualizer currently based on SDL. It should maybe be ported to use OpenGL or something fancier. The visualizer should also report FPS and be switchable between par and seq and a few other things :)

cc #35

### DIFF
--- a/demo/nbody/Cargo.toml
+++ b/demo/nbody/Cargo.toml
@@ -7,3 +7,4 @@ authors = ["Niko Matsakis <niko@alum.mit.edu>"]
 rand = "0.3"
 rayon = { path = "../../" }
 time = "0.1"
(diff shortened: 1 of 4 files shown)


bjz:
Nice! I tried it on OS X, and the window didn't show up for me :(

I'm having a go at implementing the visualiser in glium, I hope that's ok!

nikomatsakis:
@bjz that'd be awesome. I was sort of hoping somebody would do that. ;) I think I will land this PR in the meantime.

(That said, I am testing this on OS/X, so I'm surprised the window didn't show up for you.)

bjz:
:+1:

I'll also try to clean up the argument parser ;)

PR #37, opened 2016-02-01
Propagate panics across joins instead of hanging

### LINKED ISSUE #10: join hangs if a worker thread panics
If a worker thread panics, I would expect rayon to either return an error result or simply propagate that panic.  Instead, it just hangs waiting for a thread that will never finish.  Example:

```rust
extern crate rayon;

fn main() {
    let s = std::time::Duration::new(1, 0);
    rayon::join(|| std::thread::sleep(s),
                || unimplemented!());
}
```

The unimplemented panic will be reported, but then the process just spins forever waiting for it.

In #9, I toyed with a `Mutex` and `Condvar`.  If that mutex were held throughout the stolen job execution, then a panic would poison it.  I think this will wake the condvar wait with the poisoned result, but I haven't tried it.  We'd also need the `steal_until` loop to detect panics.

Amanieu:
Fixes #10

### DIFF
--- a/src/job.rs
+++ b/src/job.rs
@@ -1,5 +1,4 @@
 use latch::Latch;
-use std::mem;
 #[cfg(feature = "nightly")]
 use std::any::Any;
 #[cfg(feature = "nightly")]
@@ -17,9 +16,19 @@ enum JobResult<T> {
 /// arranged in a deque, so that thieves can take from the top of the
 /// deque while the main worker manages the bottom of the deque. This
 /// deque is managed by the `thread_pool` module.
-pub trait Job<L:Latch> {
-    unsafe fn execute(&mut self);
+///
+/// Note that we don't use a trait here due to `&mut` aliasing issues. A thief
+/// needs an `&mut` to access the func and result fields, while the original
+/// thread uses a `&` to access the latch. We solve this by simply using a raw
+/// pointer which allows us to get an &mut for func and result while getting a
+/// `&` for the latch.
+#[derive(Copy, Clone)]
+pub struct Job {
+    pub execute: unsafe fn(*mut ()),
+    pub data: *mut (),
 }
+unsafe impl Send for Job { }
+unsafe impl Sync for Job { }
 
 pub struct JobImpl<L:Latch,F,R> {
     pub latch: L,
@@ -38,8 +47,29 @@ impl<L:Latch,F,R> JobImpl<L,F,R>
         }
     }
 
-    pub unsafe fn as_static(&mut self) -> *mut Job<L> {
-        mem::transmute(self as *mut Job<L>)
+    pub unsafe fn as_job(&mut self) -> Job {
+        unsafe fn execute<L:Latch,F,R>(data: *mut ())
+            where F: FnOnce() -> R
+        {
+            let job = data as *mut JobImpl<L,F,R>;
+
+            // Use a guard here to ensure that the latch is always set, even if
+            // the function panics.
+            struct PanicGuard<'a,L:Latch + 'a>(&'a L);
+            impl<'a,L:Latch> Drop for PanicGuard<'a,L> {
+                fn drop(&mut self) {
+                    self.0.set();
+                }
+            }
+            let _guard = PanicGuard(&(*job).latch);
+            let func = (*job).func.take().unwrap();
+            (*job).result = JobImpl::<L,F,R>::run_result(func);
+        }
+
+        Job {
+            execute: execute::<L,F,R>,
+            data: self as *mut _ as *mut (),
+        }
     }
 
     pub fn run_inline(self) -> R {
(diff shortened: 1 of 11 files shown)


nikomatsakis:
Nice, thanks! This seems like a good start, but I don't think it's quite there. For example, if a panic occurs here, it will propagate all the way up and kill the worker thread -- but since it doesn't affect the core thread pool, it means we'll just continue with one less thread.

What I had in mind for a stopgap measure was to add another guard into the worker thread that will poison the central thread pool. Then at various points (e.g., when a job is stolen, or when a new job is submitted from a non-worker-thread) we can check for poisoning and propagate errors. This will at least eagerly bring rayon down once a panic occurs. It's not great, but it's better than hanging.

Another alternative might be to send a message to the main pool that will cause it to restart the thread instead, so that some jobs go down, but eventually the pool recovers. This would probably be easier/better to do based on the `recover` API.

Amanieu:
I've added a second guard to terminate the thread pool if a worker thread panics. I also fixed the handling of terminate to properly wake up all threads.

I'm planning on implementing proper panic propagation based on `recover` behind a `cfg(nightly)` switch. This would send panics back the caller of `join`, but it will always keep the pool alive.

nikomatsakis:
@Amanieu 

> I've added a second guard to terminate the thread pool if a worker thread panics. I also fixed the handling of terminate to properly wake up all threads.

Great! I'll take a look shortly.

> I'm planning on implementing proper panic propagation based on recover behind a cfg(nightly) switch. This would send panics back the caller of join, but it will always keep the pool alive.

Even better :)

src/test.rs
```diff
@@ -107,3 +107,9 @@ fn positive_test_run_pass() {
 
     compiletest::run_tests(&config);
 }
+
+#[test]
+#[should_panic]
+fn panic() {
+    join(|| {}, || panic!("should panic"));
```
nikomatsakis: I am vaguely concerned that having a panic in a test like this will infect the global schedule and affect other tests -- we may want to move these kinds of panic-y tests to run-fail tests using the `compiletest` infrastructure. We'd have to kind of copy-and-paste the `positive_tests_run_pass` test right above and change the `mode` variable to `fun-fail` and the directory to `neg-tests-run`. Then you can make `.rs` files that are expected to panic. I think they have to have a comment like `// error-pattern: foo`, where `foo` is a string we expected to see in the output.

src/thread_pool.rs
```diff
@@ -19,10 +18,10 @@ pub struct Registry {
     thread_infos: Vec<ThreadInfo>,
     state: Mutex<RegistryState>,
     work_available: Condvar,
-    terminate: AtomicBool,
 }
 
 struct RegistryState {
+    terminate: bool,
```
nikomatsakis: Nice, I like adding this here. Ensures we will see it eventually, but it doesn't interfere with the hot paths.

src/thread_pool.rs
```diff
@@ -112,11 +116,14 @@ impl Registry {
     pub unsafe fn inject(&self, injected_jobs: &[*mut Job<LockLatch>]) {
         log!(InjectJobs { count: injected_jobs.len() });
         let mut state = self.state.lock().unwrap();
+        if state.terminate {
+            panic!("attempted to inject a job after thread pool shut down");
```
nikomatsakis: This error message will be seen by end-users, right? That is, if a panic occurs in one place, it would get propagated to later jobs?

Perhaps we should make it a bit more readable, something like: 

"rayon thread pool is contaminated by a previous panic; recovery is currently unimplemented, see <https://github.com/nikomatsakis/rayon/issues/10> for details"

nikomatsakis:
OK, this looks nice. The main thing that would be nice is if you could adapt the tests to use `run-fail` and add a few more detailed tests. For example:

0. The existing "shallow panic" test you wrote.
1. Use par_iter with `weight_max` (to ensure fine-grained jobs) and map over a large array, and then panic in one of the maps. Something like `(0..2_000_000).into_par_iter().weight_max().foreach(|i| if i == 1_350_000 { panic!("boom") })`. I would of course expect this to propagate.
2. Start one thread that does a job and panics while using rayon. After it finishes, start another that does the same. The panic should propagate.

I guess that's all I can think of. :)

src/test.rs
Amanieu: Oops, I thought I removed that test. Yes it does have undesirable effects on the other tests. I'll look into adding some run-fail tests.

Amanieu:
I've pushed an updated version that fixes your comments. However there is still a big issue: the `iter_panic.rs` test sometimes segfaults, but I haven't been able to figure out the cause. It seems this is due to a thread having pointers into the stack of another thread while that thread is being unwound.

nikomatsakis:
@Amanieu hmm, yes, that makes sense. We will probably want to add some guards that ensure that we block on latches when unwinding. Can't think about this just now though.

An alternative of course is to stop using the stack, but I'm pretty reluctant to do that.

Amanieu:
I've refactored the code that deals with job objects, which seems to have fixed the segfault. It should also slightly improve benchmark numbers :)

I've also added panic propagation under the "nightly" cargo feature.

nikomatsakis:
@Amanieu awesome, sorry I havent' had a chance to take a look yet :) I'll try to do it today!

src/job.rs
```diff
@@ -1,56 +1,99 @@
...
+    #[cfg(not(feature = "nightly"))]
+    fn run_result(func: F) -> JobResult<R> {
+        JobResult::Ok(func())
+    }
+
+    #[cfg(feature = "nightly")]
+    fn run_result(func: F) -> JobResult<R> {
+        // RecoverSafe is a bit ugly to deal with...
+        // Ideally this would just be panic::recover(func)
+        let mut wrapper = AssertRecoverSafe::new(Some(func));
```
nikomatsakis: It'd be good to add a comment explaining why we think it's legit to call this function "recover safe" despite having no idea what kind of data it touches. (The answer of course is because we propagate panics.) So perhaps just write: "// We can call this recover safe because we propagate panics."

src/job.rs
```diff
@@ -1,56 +1,99 @@
...
+        match panic::recover(move || (*wrapper).take().unwrap()()) {
```
nikomatsakis: Can you add `FIXME(rust-lang/rust#XXX)` where XXX is the number of your PR?

src/thread_pool.rs
```diff
@@ -109,14 +113,18 @@ impl Registry {
         self.work_available.notify_all();
     }
 
-    pub unsafe fn inject(&self, injected_jobs: &[*mut Job<LockLatch>]) {
+    pub unsafe fn inject(&self, injected_jobs: &[*mut (Job<LockLatch> + 'static)]) {
```
nikomatsakis: Is the explicit `'static` needed here?

Amanieu: Yes, it fails to compile otherwise. It's automatically added by lifetime inference in other places, but it needs to be explicitly specified here.

nikomatsakis: Huh. I'm a bit surprised, but I guess I forget what the defaulting rules are around `*mut`.

Amanieu: It's because it's defaulting to the lifetime of the outer `&[]`.

nikomatsakis: Yes, I just expected it to be overridden by the `*mut`.

nikomatsakis:
So, as I said on IRC, this looks great!

In later PRs, it'd be nice to play with the setup of `Job` and see if we can't go back to using a trait somehow (or maybe we'll eventually fix stuff like `self: *mut Self`...), and of course it'd be nice to write more exhaustiv tests, but this is obviously far better than what's there now.

cuviper:
Nice work, @Amanieu, thanks!

PR #38, opened 2016-02-01
Tweaks to the limit number of threads PR

nikomatsakis:
This is a tweaked version of PR #29 with some additional commits.


nikomatsakis:
D'oh, I have to fix the demos.

willi-kappler:
It's just these three things: `InitResult`, `NumberOfThreadsZero`, `NumberOfThreadsNotEqual`. Besides that it looks really good! Thanks for "mentoring" and letting me work on this ;-)

lilianmoraru:
Note the panic message "expected InitOk" - but there is no more InitOk in the enum(that was InitResult and now InitError).

nikomatsakis:
it occurred to me that the error type should really implement [the Error trait](http://doc.rust-lang.org/std/error/index.html#the-error-trait)

nikomatsakis:
But I'll leave that for another PR. :)

PR #41, opened 2016-02-02
Nbody demo improvements

### LINKED ISSUE #40: Use OpenGL for nbody demo
#36 added an nbody demo. It would be nice to glium instead of SDL though.

bjz:
Uses `cgmath` for maths, `glium` for graphics, and `docopt` for arg parsing.

Fixes #40.

### DIFF
--- a/demo/nbody/src/main.rs
+++ b/demo/nbody/src/main.rs
@@ -18,9 +18,9 @@ use self::visualize::visualize_benchmarks;
 use self::nbody::NBodyBenchmark;
 
 const USAGE: &'static str = "
-Usage: nbody bench [--no-par | --no-seq] [--visualize --bodies N --ticks N]
+Usage: nbody bench [--no-par | --no-seq] [--bodies N --ticks N]
        nbody visualize [--mode MODE --bodies N]
-       nbody (--help | --version)
+       nbody --help
 
 Commands:
     bench              Run the benchmark and print the timings.
--- a/demo/nbody/src/main.rs
+++ b/demo/nbody/src/main.rs
@@ -22,7 +22,7 @@ use self::nbody::NBodyBenchmark;
 
 const USAGE: &'static str = "
 Usage: nbody bench [--no-par | --no-seq] [--visualize --bodies N --ticks N]
-       nbody visualize [--bodies N]
+       nbody visualize [--mode MODE --bodies N]
        nbody (--help | --version)
 
 Commands:
@@ -31,18 +31,35 @@ Commands:
 
 Options:
     -h, --help         Show this message.
-    --no-par           Skip parallel execution.
-    --no-seq           Skip sequential execution.
+    --no-par           Skip parallel execution in the benchmark.
+    --no-seq           Skip sequential execution in the benchmark.
+    --mode MODE        Execution mode for the visualizer [default: par].
     --bodies N         Use N bodies [default: 4000].
     --ticks N          Simulate for N ticks [default: 100].
 ";
 
+#[derive(Copy, Clone, RustcDecodable)]
+pub enum ExecutionMode {
+    Par,
+    Seq,
+}
+
+impl fmt::Display for ExecutionMode {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ExecutionMode::Par => writeln!(f, "par"),
+            ExecutionMode::Seq => writeln!(f, "seq"),
+        }
+    }
+}
+
 #[derive(RustcDecodable)]
-struct Args {
+pub struct Args {
     cmd_bench: bool,
     cmd_visualize: bool,
     flag_no_par: bool,
     flag_no_seq: bool,
+    flag_mode: ExecutionMode,
     flag_bodies: usize,
     flag_ticks: usize,
 }
@@ -56,6 +73,7 @@ impl fmt::Display for Args {
         if self.flag_no_par     { try!(writeln!(f, "  --no-par")); }
         if self.flag_no_seq     { try!(writeln!(f, "  --no-seq")); }
 
+        try!(writeln!(f, "  --mode {}", self.flag_mode));
         try!(writeln!(f, "  --bodies {}", self.flag_bodies));
         try!(writeln!(f, "  --ticks {}", self.flag_ticks));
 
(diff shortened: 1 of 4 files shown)


demo/nbody/src/visualize.rs
```diff
@@ -1,62 +1,180 @@
...
+        3,  7, 11,
+        3,  9,  6,
+        4,  8,  9,
+        5, 11, 10,
+        6,  9,  8,
+        7, 10, 11,
+    ];
+
+    (vertices, indices)
+}
```
bjz: Couldn't be bothered making a sphere function - had this code lying around....

demo/nbody/src/visualize.rs
```diff
@@ -1,62 +1,180 @@
...
+        };
+
+        let mut target = display.draw();
+
+        let (width, height) = target.get_dimensions();
+        let aspect = width as f32 / height as f32;
+
+        let proj = cgmath::perspective(Rad::full_turn() / 6.0, aspect, 0.1, 3000.0);
+        let view = Matrix4::look_at(Point3::new(10.0, 10.0, 10.0), Point3::origin(), Vector3::unit_z());
+        let view_proj: [[f32; 4]; 4] = (proj * view).into();
```
bjz: The camera could be tweaked more, but you can see most of it.

demo/nbody/src/main.rs
```diff
@@ -1,111 +1,123 @@
...
-
-    let par_time = {
+fn run_benchmarks(run_par: bool, run_seq: bool, bodies: usize, ticks: usize) {
+    let par_time = if run_par {
         let mut rng = XorShiftRng::from_seed([0, 1, 2, 3]);
-        let mut benchmark = NBodyBenchmark::new(num_bodies, &mut rng);
+        let mut benchmark = NBodyBenchmark::new(bodies, &mut rng);
         let par_start = time::precise_time_ns();
+
+        let mut progress = ProgressBar::new(ticks);
```
lilianmoraru: Isn't this also benchmarking the ProgressBar?
Output takes a lot of time(Note that I don't exactly know how `pbr` works but my guess is that when you call `progress.inc()` it blocks to write to the output).

demo/nbody/src/main.rs
```diff
@@ -1,111 +1,123 @@
...
+
+        try!(writeln!(f, "  --bodies {}", self.flag_bodies));
+        try!(writeln!(f, "  --ticks {}", self.flag_ticks));
+
+        Ok(())
+    }
+}
 
 fn main() {
-    let mut num_bodies = DEFAULT_NUM_BODIES;
```
lilianmoraru: I like how `docopt` makes all of this look so much better :)

lilianmoraru:
I haven't read the book for a long time(I think it was mentioned there) that I already forgot you can do things like `'main: loop { ... code ...; break 'main; }` :)

demo/nbody/src/main.rs
bjz: Hmm, you might be right. Although `ticks` is quite small. I can remove it though, just felt nicer to have something to look at :)

bjz:
Yeah, I only remember it because it is useful for breaking out of nested game/input event loops. :)

nikomatsakis:
Hmm, building locally I get errors like this:

```
/Users/nmatsakis/.multirust/toolchains/nightly/cargo/registry/src/github.com-88ac128001ac3a9a/glutin-0.4.7/src/api/cocoa/mod.rs:148:18: 148:62 error: the trait `objc::encode::Encode` is not implemented f\
or the type `*mut libc::c_void` [E0277]
/Users/nmatsakis/.multirust/toolchains/nightly/cargo/registry/src/github.com-88ac128001ac3a9a/glutin-0.4.7/src/api/cocoa/mod.rs:148             decl.add_ivar::<*mut libc::c_void>("glutinState");
                                                                                                                                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

any idea what is going on?

nikomatsakis:
cargo update fixed it.

nikomatsakis:
@bjz ok, to get this to work, I had to `s/cmd_benchmark/cmd_bench`, but once I did that, it looks pretty awesome. I've not looked at the actual diff yet -- but I have one more request. Can you make it so that the visualizer reports FPS, and can be changed from par to seq mode? I mean, I'll merge it either way, but that'd be the icing on the cake. ;)

demo/nbody/src/main.rs
nikomatsakis: Yeah, I'm not sure about benchmarking the progress bar.

demo/nbody/src/main.rs
```diff
@@ -1,111 +1,123 @@
+ #![feature(augmented_assignments)]
```
nikomatsakis: Hmm, I'm a bit disappointed to take this off of stable. But it's just a demo I guess.

demo/nbody/src/nbody.rs
```diff
@@ -230,22 +223,22 @@ fn next_velocity(time: usize, prev: &Body, bodies: &[Body]) -> (Vector, Vector)
                 // normalize the distance vector
                 let d = (r / length) * c;
 
-                acc.increment(d);
-                acc2.decrement(d);
+                acc += d;
+                acc2 -= d;
```
nikomatsakis: And this IS so much prettier. :)

demo/nbody/src/main.rs
nikomatsakis: I definitely saw a drop in parallel speedup locally when using the progress bar. Makes sense. Amdahl's law and all that. (3.9x to 3.5x on a 4core)

demo/nbody/src/main.rs
bjz: Only thing is the `*=` operators from cgmath. I could just use regular re-assignments, which I'm guessing should optimize ok, which mean we could drop cgmath's `"unstable"` feature flag.

bjz:
FPS counter would take a bit more work (me figuring out fonts and stuff).

demo/nbody/src/nbody.rs
bjz: At the expense of `#![feature(augmented_assignments)]` :(

nikomatsakis:
@bjz can you make it switch from par to seq by typing `p` or `s` ?

nikomatsakis:
@bjz yeah let's keep the augmented-assignments, this is just a demo, and they are on their way to being stable anyway.

nikomatsakis:
Ah, good enough. Thanks @bjz, this is great. :)

PR #42, opened 2016-02-02
Adapt benchmarks to the changes in the initialize function

lilianmoraru:
I wanted to benchmark the changes in #37, but the benchmarks were broken due to the latest changes to support a way to indicate a limit on the number of threads that rayon can use.

This adapts to the new changes.

### DIFF
--- a/benches/factorial.rs
+++ b/benches/factorial.rs
@@ -8,8 +8,10 @@ extern crate test;
 
 use num::{One, BigUint};
 use rayon::par_iter::*;
+use rayon::Configuration;
 use std::ops::Mul;
 
+const INIT_FAILED: &'static str = "Rayon failed to initialize";
 const N: u32 = 9999;
 
 /// Compute the Factorial using a plain iterator.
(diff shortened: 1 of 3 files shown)


lilianmoraru:
Not directly related to this PR but seems like `fibonacci_iterative` is optimised away.
```sh
running 4 tests
test fibonacci_iterative ... bench:          15 ns/iter (+/- 4)
test fibonacci_join_1_2  ... bench: 127,917,694 ns/iter (+/- 50,380,269)
test fibonacci_join_2_1  ... bench: 128,785,715 ns/iter (+/- 53,188,061)
test fibonacci_recursive ... bench:  54,872,145 ns/iter (+/- 5,577,286)
```

cuviper:
@lilianmoraru FWIW that `fibonacci_iterative` result is an intentional nod to the fact that the recursive and join methods are horrible ways to compute this. :)
(Perhaps we should `#[ignore]` it if this causes confusion)

lilianmoraru:
@cuviper Ow, you did mention this in the pull request. I thought that laptop was at fault...

PR #43, opened 2016-02-03
fix for #39 implement Error and Display for InitError

### LINKED ISSUE #39: Error type should implement `Error` trait
We recently introduced the `InitError` for "configuration gone wrong". This enum should implement the `Error` trait.

willi-kappler:
I would like to fix #39, so I followed the example in TRPL chapter 5.7.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -3,6 +3,8 @@ use latch::{LockLatch, SpinLatch};
 use log::Event::*;
 use job::{Code, CodeImpl, Job};
 use std::sync::Arc;
+use std::error::Error;
+use std::fmt;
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Great, thanks! Maybe add a test that just coerces to `Box<Error>`, to make sure everything is setup right? Something like:

```rust
#[test]
fn coerce_to_box_error() {
    let _: Box<Error> = From::from(InitError::NumberOfThreadsZero); // check that coercion succeeds
}
```

willi-kappler:
Nice, I'll remember that. The test case passes, hope everything else is OK.

nikomatsakis:
Thanks!

PR #47, opened 2016-02-05
Simplify Producer and Consumer traits further

nikomatsakis:
This PR combines a number of refactorings. The general goal is to increase safety and to lower the number of esoteric annotations (lifetime parameters, phantom types) as well as general...clumsiness of the parallel iterator internal traits.

- Remove the `Shared` associated types from `Producer` and `Consumer`
    - instead, carry the shared pointers inside the implementations
    - this is possible because we are always obtaining instances of these types via callbacks now
- Remove the lifetimes from `Producer` and `Consumer`
    - they were only ever needed to deal with `Shared`
- Make `Item` an *input type* (not an associated type) of `Consumer`
    - this makes sense because items flow *in* to a consumer, so in a sense it is not the one that chooses the kinds of items it will be provided
    - of course, in some cases, consumers can only consume compatible kinds of items, but nobody says they have to have an impl that applies to all input types
    - this in turn allows removing a number of phantom types that only served to "remember" the type of input we were expecting; this is now specified on the trait reference `Consumer<Item>`
- Refactor the producer to extend `IntoIterator`
    - when we start sequential execution, we convert to an iterator
    - this lets us leverage existing iterator implementations
- Refactor the consumer to have more distinct phases
    - a consumer now has a clear protocol: either it gets split into two parallel consumes (whose results are then reduced), or else it gets converted into a folder and processes items sequentially
    - this protocol shift is needed to support removing shared as well
    - it also helps with making things safe 
- Make all traits safe
    - the unsafety is now concentrated in the collector. Having the more refined protocol helps us to remove some assertions from this path; it's now fairly minimal. I think we could actually remove some of the assertions we have by having the collector call `bridge` directly instead of calling `drive` (in fact, arguably, we don't need `drive` at all).

Some things that might be nice to do before landing:

- [ ] Measurements of impact (which sort of needs good benchmarks)
- [ ] Some negative tests of evil producers that fail to uphold their end of the bargain in various ways; we would want to be sure that the `collect` consumer doesn't fail under those cases.
    - [ ] fail to produce correct number of items
    - [ ] produce too many items
    - [ ] abuse consumer protocol by never calling `complete`
    - [ ] abuse consumer protocol by leaking calling
    - [ ] not sure what else


nikomatsakis:
cc @cuviper @bjz -- these changes may interest you :)

src/par_iter/enumerate.rs
```diff
@@ -82,38 +74,32 @@ impl<M> IndexedParallelIterator for Enumerate<M>
...
+    type Item = (usize, P::Item);
+    type IntoIter = iter::Enumerate<P::IntoIter>;
 
-    unsafe fn produce(&mut self, shared: &Self::Shared) -> (usize, P::Item) {
-        let item = self.base.produce(shared);
-        let index = self.offset;
-        self.offset += 1;
-        (index, item)
+    fn into_iter(self) -> Self::IntoIter {
+        self.base.into_iter().enumerate()
```
cuviper: Surely this needs to take `self.offset` into account?
Perhaps use instead: `(self.offset..).zip(self.base)`

Are the current tests big enough to actually exercise splits and run in parallel?
We may need to repeat them at different weights, say 0, 1, 100, inf.

src/par_iter/collect/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+    assert!(len < (isize::MAX) as usize);
+
+    v.truncate(0); // clear any old data
+    v.reserve(len); // reserve enough space
+    let target = v.as_mut_ptr(); // get a raw ptr
+    let writes = AtomicUsize::new(0);
+    let consumer = unsafe { // assert that target..target+len is unique and writable
+        CollectConsumer::new(&writes, target, len)
+    };
+    pi.drive(consumer);
```
cuviper: Instead of this manual consumer, what if you use `from_raw_parts_mut` for an uninitialized slice, then
`slice.par_iter_mut().zip(pi).for_each(/* atomic inc, assign */)`

I'm not at home, so I can't try yet, but if this works it would simplify a lot...

src/par_iter/enumerate.rs
bjz: I would love to see some quickcheck tests as well to exercise some of these properties.

src/par_iter/internal.rs
```diff
@@ -9,133 +9,126 @@ use super::len::*;
...
-    /// `0..index` and one processing items from `index..`.
-    unsafe fn split_at(self, shared: &Self::Shared, index: usize) -> (Self, Self);
+    /// `0..index` and one processing items from `index..`. Also
+    /// produces a reducer that can be used to reduce the results at
+    /// the end.
+    fn split_at(self, index: usize) -> (Self, Self, Self::Reducer);
+
+    /// Convert the consumer into a folder that can consume items
+    /// sequentially, eventually producing a final result.
+    fn fold(self) -> Self::Folder;
```
bjz: `fold` is kind of a confusing name, given the type signature. How about constraining `Consumer` to also implement `Into<Self::Folder>`?

src/par_iter/util.rs
```diff
@@ -1,25 +0,0 @@
-use std::marker::PhantomData;
-
-/// A little structure used to mark some type. This is often used when
-/// we have some phantom type we wish to remember as part of a struct.
-/// This doesn't represent "reachable data", and hence it is
-/// considered sendable regardless of `T`, unlike `PhantomData`.
-pub struct PhantomType<T> {
-    data: PhantomData<T>
-}
```
bjz: I'm really glad to see the end of this. Seemed like a dangerous subversion of the type system.

src/par_iter/internal.rs
```diff
@@ -9,133 +9,126 @@ use super::len::*;
...
+
+pub trait Reducer<Result> {
 
     /// Reduce two final results into one; this is executed after a
     /// split.
-    unsafe fn reduce(shared: &Self::Shared,
-                     left: Self::Result,
-                     right: Self::Result)
-                     -> Self::Result;
+    fn reduce(self, left: Result, right: Result) -> Result;
```
bjz: Like with `fold`, this doesn't really match the type signature I would expect for `reduce`. Not sure what a better name would be though.

src/par_iter/internal.rs
```diff
@@ -9,133 +9,126 @@ use super::len::*;
...
 }
 
 /// A stateless consumer can be freely copied.
-pub trait UnindexedConsumer<'c>: Consumer<'c> {
+pub trait UnindexedConsumer<ITEM>: Consumer<ITEM> {
     fn split(&self) -> Self;
```
bjz: Since this does not have the type signature `fn split(self) -> (Self, Self)`, perhaps `split_off` would be better?

src/par_iter/internal.rs
```diff
@@ -9,133 +9,126 @@ use super::len::*;
...
+    fn reducer(&self) -> Self::Reducer;
```
bjz: `get_reducer`?

src/par_iter/internal.rs
```diff
@@ -9,133 +9,126 @@ use super::len::*;
 
 pub trait ProducerCallback<ITEM> {
     type Output;
-    fn callback<'p, P>(self, producer: P, shared: &'p P::Shared) -> Self::Output
-        where P: Producer<'p, Item=ITEM>;
+    fn callback<'p, P>(self, producer: P) -> Self::Output
```
bjz: Do we still need the `'p`?

src/par_iter/collect/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+mod consumer;
+use self::consumer::CollectConsumer;
+
+pub fn collect_into<PAR_ITER,T>(mut pi: PAR_ITER, v: &mut Vec<T>)
+    where PAR_ITER: ExactParallelIterator<Item=T>,
+          PAR_ITER: ExactParallelIterator,
+          T: Send,
+{
+    let len = pi.len();
+    assert!(len < (isize::MAX) as usize);
```
bjz: Do we need parens around the `isize::MAX`?

src/par_iter/map.rs
```diff
@@ -21,17 +21,15 @@ impl<M, MAP_OP, R> ParallelIterator for Map<M, MAP_OP>
...
+    {
+        let consumer1 = MapConsumer::new(consumer, &self.map_op);
+        self.base.drive_unindexed(consumer1)
     }
 }
 
-unsafe impl<M, MAP_OP, R> BoundedParallelIterator for Map<M, MAP_OP>
+impl<M, MAP_OP, R> BoundedParallelIterator for Map<M, MAP_OP>
     where M: BoundedParallelIterator,
           MAP_OP: Fn(M::Item) -> R + Sync,
```
bjz: Just checking - does this mean that `MAP_OP` is `Sync`, or just its return value?

bjz: Oh wait, dw - it must be the return value, because `R` is a type parameter, not a trait. :P

src/par_iter/internal.rs
nikomatsakis: Conventions vary, but I think reduce is the right name for this. This is my convention:

- fold: starts with an initial value of type U, and hence the closure has signature `(U x T) -> U`
- reduce: draws its initial value from the first element, and hence the closure has signature `(T x T) -> T`

This fits the convention that [scheme established](http://www.gnu.org/software/mit-scheme/documentation/mit-scheme-ref/Reduction-of-Lists.html), but it also has other precedent. And certainly lots of parallel projections talk about "reduction" operations.

src/par_iter/map.rs
nikomatsakis: It means `MAP_OP` is `Sync`.

src/par_iter/enumerate.rs
nikomatsakis: The current tests are clearly not enough in this case. :) 

For the most part I haven't stressed about testing them in parallel because the producers themselves were essentially stateless, and the type system ensures that we don't screw up too badly, but I think some targeted tests would make a lot of sense. (ie., manually invoking the relevant methods)

src/par_iter/collect/mod.rs
nikomatsakis: Hmm, interesting thought! That would indeed simplify a lot. I have a few reservations though:

1. I've been planning to extend `collect` to work for bounded parallel iterations (e.g., with filter). The idea would be to track where writes have occurred and then do a compression step (if needed). But then again, I guess that would work fine with your approach.

2. I wanted to drop partially initialized results on panic. This would be fairly easy with the current approach (when the "Folder" gets dropped, if it hasn't finished, it can drop those values its written thus far; similarly, it can return a dummy value as its Result. When this value is reduced, it's a no-op, but if it gets dropped, it drops the completed values). However, if we adopt the "track where writes occurred" approach, that can be handled too.

3. We're relying on various bits of safe code, but it's all safe code we control I suppose, so that seems ok.

Hmm so maybe it would work out. That'd be very cool.

nikomatsakis: Along those lines, I've noticed I don't have many actual use cases for `collect`. It *seems* handy, but all the tests I've written tend to use other things, or just use `for_each` and write directly into the output buffer (as you suggested). Basically the only advantage `collect` brings (right now) is handling uninitialized memory for you; I guess if it supported `filter` it would be mildly more interesting.

src/par_iter/internal.rs
nikomatsakis: Yeah, `into_folder` is the right name.

src/par_iter/util.rs
nikomatsakis: I don't know that it was a *subversion*, but I'm glad to see it go, because it was annoying and made the code seem complex. I actually found the process of removing it quite enlightening, it suggested principles for when to choose an associated type vs a type parameter on traits that I hadn't fully understood.

src/par_iter/internal.rs
nikomatsakis: Yes.

src/par_iter/internal.rs
nikomatsakis: I have some kind of aversion to `get_` prefixes. ;)

nikomatsakis: I guess `as_reducer` or `to_reducer` is our usual convention here. I'll go with `to`.

nikomatsakis:
OK, addressed the various nits and bugs, the only thing I haven't investigated yet is simplifying `collect` per @cuviper's suggestion. We could definitely use some more tests targeting splits. :)

src/par_iter/collect/mod.rs
nikomatsakis: So, what I was saying about filter made no sense -- well, in particular, we couldn't adapt the approach you suggested to work with bounded parallel iterations, unless we added some sort of enumerated for that case which gave you the kind of "original index" (also in the case of a hand-written collect, I'd have to extend the protocol so that there is a version like `BoundedConsumer` where, for each split, you find out the bound at each step).

src/par_iter/internal.rs
bjz: It's more that I would have expected something like: `fn reduce(Self<T>, U, impl Fn(U, T) -> U) -> U`.

src/par_iter/internal.rs
bjz: Apparently the convention is to not use `get_`! https://www.reddit.com/r/rust/comments/44gmmo/some_thoughts_on_properties/czq5vvt

nikomatsakis:
OK, given my dearth of time, I've decided to defer the "negative tests" and make an issue for them. I'll try to do some benchmarking of what we have now -- even though the existing benchmarks don't really stress test much of the par iter infrastructure. But that's ok.

nikomatsakis:
OK, I did measurements and found basically no performance difference (as expected).

PR #52, opened 2016-02-13
Add Vec::into_par_iter

### LINKED ISSUE #17: Support moving into paralell iterators

cuviper:
This is modeled similarly to what &mut[] does, except that values are
moved out with ptr::read, and any remains are dropped.

Fixes #17.

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -40,6 +40,7 @@ pub mod map;
 pub mod weight;
 pub mod zip;
 pub mod range;
+pub mod vec;
 mod util;
 
 #[cfg(test)]
--- a/src/par_iter/test.rs
+++ b/src/par_iter/test.rs
@@ -104,6 +104,58 @@ pub fn check_increment() {
 }
 
 #[test]
+pub fn check_move() {
+    let a = vec![vec![1, 2, 3]];
+    let ptr = a[0].as_ptr();
+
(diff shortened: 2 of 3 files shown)


cuviper:
FWIW, I also implemented this on top of #47 at [cuviper:vec-into_par_iter-pr47](https://github.com/cuviper/rayon/tree/vec-into_par_iter-pr47).  Actually, I did that first.  I do think that abstraction looks nicer, although it required me to write a custom `Iterator` to handle moving out of a slice, much like a `Drain`.

bjz:
:+1: - makes Rayon much more useful now!

nikomatsakis:
Ah, I should probably land #47 first. Sorry, I've been rather slow about landing Rayon PRs. But this looks great!

nikomatsakis:
(Well, I've not read it in detail yet.)

cuviper:
I can send a PR into #47's branch if you like.  Or just wait for it - I'm
not in a hurry.

nikomatsakis:
OK, I landed #47, so if you don't mind rebasing, that'd be great. (I'm going to close this PR for now.)

PR #54, opened 2016-02-16
Add Vec::into_par_iter

### LINKED ISSUE #17: Support moving into paralell iterators

cuviper:
This is modeled similarly to what &mut[] does, except that values are
moved out with ptr::read, and any remains are dropped.

Fixes #17.

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -40,6 +40,7 @@ pub mod map;
 pub mod weight;
 pub mod zip;
 pub mod range;
+pub mod vec;
 
 #[cfg(test)]
 mod test;
--- a/src/par_iter/test.rs
+++ b/src/par_iter/test.rs
@@ -126,6 +126,54 @@ pub fn check_increment() {
 }
 
 #[test]
+pub fn check_move() {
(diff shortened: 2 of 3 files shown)


nikomatsakis:
This looks good to me: my only nitpick is <https://github.com/nikomatsakis/rayon/pull/54#commitcomment-16171143>. At first I expected to use a `*mut` in the iterator in place of `&mut`, but actually the way you are using it seems correct to me, and results in much cleaner code.

nikomatsakis:
@cuviper 

> I can't call it right here, because the slice has borrowed self.vec. And for exception safety, I didn't want to wait until the producer is done, else a panic unwind would miss the truncation and double-drop the contents.

I see, I see. Yes, you could do something with `from_raw_parts` or whatever but I agree this is cleaner and simpler. I really like minimizing unsafe code code where possible :)

nikomatsakis:
Nice :)

PR #56, opened 2016-02-19
Add a reduce version of nbody benchmark, some new combinators

nikomatsakis:
This updates the nbody benchmark in two ways:

1. Adds a mode (par-reduce) that uses reduction internally. Right now this is kind of clumsily done, resulting in code duplication. This version runs more slowly than the original but exercises more of the parallel iterators API (and might provide a good target for optimization).
    - That said, I'd like to work on it a bit more to use parallel iterators better and remove DRY before we go too crazy.
2. Updates the visualizer to let you press `P`, `R`, or `S` during execution.
3. Updates the CLI to just use `--mode` universally.

cc @bjz

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,7 @@ num = "0.1.30"
 
 [features]
 nightly = []
+
+# Unstable APIs that have not yet
+# proven their utility.
+unstable = []
--- a/demo/nbody/Cargo.toml
+++ b/demo/nbody/Cargo.toml
@@ -8,6 +8,6 @@ cgmath = { version = "0.7", features = ["unstable"] }
 docopt = "0.6"
 glium = "0.13"
 rand = "0.3"
-rayon = { path = "../../" }
+rayon = { path = "../../", features = ["unstable"] }
 rustc-serialize = "0.3"
 time = "0.1"
--- a/demo/nbody/src/main.rs
+++ b/demo/nbody/src/main.rs
@@ -18,7 +18,7 @@ use self::visualize::visualize_benchmarks;
 use self::nbody::NBodyBenchmark;
 
 const USAGE: &'static str = "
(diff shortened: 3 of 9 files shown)


nikomatsakis:
Results for nbody running on my 8x2 core (2 hyperthreads per core) machine:

```
Parallel time   : 887938949
ParReduce time  : 1127716606
Sequential time : 8229587355
Parallel speedup: 9.26819
Parallel reduce speedup: 7.297567
```

nikomatsakis:
And on my 4-core macbook pro:

```
Parallel time    : 1120228393
ParReduce time   : 1474255482
Sequential time  : 4500234184
Parallel speedup : 4.017247
ParReduce speedup: 3.052547
```

lilianmoraru:
@nikomatsakis Could it also indicate what those numbers are?

nikomatsakis:
@lilianmoraru you mean like a unit? I left it off intentionally since the ratio is the only thing that really matters... but I think maybe they are ns? We could print it I guess

nikomatsakis:
@lilianmoraru done :)

lilianmoraru:
@nikomatsakis Yes, I meant the time, thanks.

nikomatsakis:
Going to land this for now. I'd like to do some more work making the variants DRY -- should be possible to at least extra subexpressions. But I'd also like to expose the new `reduce_with_identity` and (unstable) `fold` operations, and I like having some benchmarks that expose more interesting iterator functionality, even if I suspect more tuning is in order.

Cargo.toml
```diff
@@ -17,3 +17,7 @@ num = "0.1.30"
 
 [features]
 nightly = []
+
+# Unstable APIs that have not yet
+# proven their utility.
+unstable = []
```
bjz: You need to set up your editor/gitconfig to insert new lines at the end of your files :3

PR #60, opened 2016-02-23
Introduce prelude

### LINKED ISSUE #59: Using `par_iter::*` is too broad
The docs recommend `use rayon::par_iter::*;` -- but this includes every `pub mod`, and those have very generic names.  So something simple like this fails:

```rust
use rayon::par_iter::*;
use std::slice;
```

```error: a module named `slice` has already been imported in this module```

If those submodules really need to be public, perhaps there should be a prelude to glob use instead?

### PR DESCRIPTION
**Pull request by nikomatsakis**
Fixes #59.

cc @cuviper

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ example, to compute the sum of the squares of a sequence of integers,
 one might write:
 
 ```rust
-use rayon::par_iter::*;
+use rayon::prelude::*;
 fn sum_of_squares(input: &[i32]) -> i32 {
     input.par_iter()
          .map(|&i| i * i)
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by nikomatsakis** [comment]
But I wonder if we should include an alias for rayon itself, so that one can call `rayon::join` without having to do:

```
use rayon;
use rayon::prelude::*;
```

which is annoying.

**Comment by nikomatsakis** [comment]
Could do something hacky like

```
pub mod rayon {
    pub use super::super::*;
}
```

to re-export everything from the top-level in the prelude :)

**Comment by nikomatsakis** [comment]
But I guess that would make simple programs like

```rust
extern crate rayon;
use rayon::prelude::*;
```

fail to compile, so I think I will not do it for now.

cuviper:
Did you forget to add your `prelude.rs`?

nikomatsakis:
@cuviper yes :)

cuviper:
:+1:

reem:
Personally I just use `::rayon` for accessing the top-level functions except in the crate root where it's just `rayon`. I don't think it's a big deal to ask people to `use rayon;` if they really don't want to have the leading `::`.

PR #62, opened 2016-02-24
Added `fn par_chunks` and `fn par_chunks_mut` methods.

pnkfelix:
Added `fn par_chunks` and `fn par_chunks_mut` methods.

These for iteration over chunks of slice in parallel. (The particular use case I am thinking of are things like processing the rows of a matrix that uses a slice in row-major order for its representation.)

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -68,6 +68,36 @@ pub trait IntoParallelRefMutIterator<'data> {
     fn par_iter_mut(&'data mut self) -> Self::Iter;
 }
 
+pub trait ToParallelChunks<'data> {
+    type Iter: ParallelIterator<Item=&'data [Self::Item]>;
+    type Item: Sync + 'data;
+
+    /// Returns a parallel iterator over at most `size` elements of
+    /// `self` at a time. The chunks do not overlap.
+    ///
+    /// The policy for how a collection is divided into chunks is not
+    /// dictated here (e.g. a B-tree-like collection may by necessity
+    /// produce many chunks with fewer than `size` elements), but an
+    /// implementation should strive to maximize chunk size when
+    /// possible.
+    fn par_chunks(&'data self, size: usize) -> Self::Iter;
+}
+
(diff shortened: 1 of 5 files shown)


nikomatsakis:
@pnkfelix 

> hmm I wonder if the tests are passing because I'm not actually ever dividing the work...

I think so.

cuviper:
Does it make sense to have distinct traits for each of these?  Do you really think other types of collections will be able to implement this?

I think I'd be more inclined to make a single extension trait like `ParallelSlice` implemented only for `[T]`, so it's a bit closer to how primitive slice methods are arranged.  I could also see adding `par_windows`.

nikomatsakis:
@cuviper it seems like @pnkfelix was anticipating implementing these methods for tree-shaped data structures too, based on his comments.

nikomatsakis:
(which might make sense, I agree, although one probably wouldn't be able to control the chunk size very well in that scenario)

nikomatsakis:
(iow they might want to use different methods that don't even take a chunk size)

nikomatsakis:
Eh, let's take it with the current trait names. We can always rearrange later. Thanks @pnkfelix ! :)

pnkfelix:
Yeah as an afterthought when writing the doc comment I decided to loosen the spec (compared to what slice.chunks() guarantees).

But the test I wrote, as well as my stated matrix multiplication use case, definitely assumed the specific semantics used here which really only works for slice-like things ...

So it seems like there's a disconnect somewhere in my mental model for how this could actually be used in the general case.  Maybe I should indeed just move it all into a trait dedicated to slices

nikomatsakis:
I'd be good with narrowing this to slices and offering a precise guarantee. I agree that many use cases require precise division. 

I am not sure if the binary tree use case is real yet. Certainly I would expect they may have a very particular variant on how they drive consumers, where they choose the split points according to their internal structure rather than just picking a half way mid point, but it's not clear that the end user will want to care about this. :)

PR #63, opened 2016-02-25
Fixes for panic propagation

### LINKED ISSUE #58: Periodic crashes with panic propagation
I have been seeing travis builds fail unreliably in the negative panic tests. I'll try to pin this down at some point.

cc @Amanieu

Amanieu:
The key part is the missing guard in `steal_until`, which was causing the crashes because unwinding would overwrite the job data.

Another issue was that some injected jobs could be left hanging after the thread pool was terminated. This is resolved by aborting all injected jobs on termination.

I also changed the code back to using a trait and just wrapped the mutable values in `UnsafeCell`.

Fixes #58

### DIFF
--- a/src/thread_pool.rs
+++ b/src/thread_pool.rs
@@ -1,7 +1,7 @@
 use Configuration;
 use deque;
 use deque::{Worker, Stealer, Stolen};
-use job::Job;
+use job::JobRef;
 use latch::{Latch, LockLatch, SpinLatch};
 #[allow(unused_imports)]
 use log::Event::*;
@@ -9,6 +9,8 @@ use rand;
 use std::cell::Cell;
 use std::sync::{Arc, Condvar, Mutex, Once, ONCE_INIT};
 use std::thread;
+use std::collections::VecDeque;
+use std::mem;
 use util::leak;
 use num_cpus;
 
@@ -23,7 +25,7 @@ pub struct Registry {
 struct RegistryState {
     terminate: bool,
     threads_at_work: usize,
-    injected_jobs: Vec<Job>,
+    injected_jobs: VecDeque<JobRef>,
 }
 
 unsafe impl Send for Registry { }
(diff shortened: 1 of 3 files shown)


nikomatsakis:
:fireworks: I'll take a look soon, sorry...lots going on! :)

src/thread_pool.rs
```diff
@@ -263,13 +268,26 @@ impl WorkerThread {
 
     #[cold]
     pub unsafe fn steal_until(&self, latch: &SpinLatch) {
+        // If another thread stole our job when we panic, we must halt unwinding
+        // until that thread is finished using it.
+        struct PanicGuard<'a>(&'a SpinLatch);
+        impl<'a> Drop for PanicGuard<'a> {
+            fn drop(&mut self) {
+                while !self.0.probe() {
```
nikomatsakis: Maybe worth writing a helper fn for this, since we also do a similar spin loop in api.rs?

nikomatsakis:
Looks good -- so the problem was a missing block in `steal_until`?

I definitely prefer the trait + unsafe-cell variant.

src/thread_pool.rs
Amanieu: We could use the [scopeguard](https://crates.io/crates/scopeguard) crate which makes panic guards easier to write.

nikomatsakis: Hmm, perhaps. scopeguard offers `defer!`, which works in this instance, but not the other case, since we wanted the ability to `mem::forget`...

nikomatsakis:
Well, I'll merge this now, we can consider further cleanup later. I'd like to get a fix in for #58.

src/thread_pool.rs
Amanieu: It also allows you to create a guard without the macro, which *does* allow you to `mem::forget` it later.

Amanieu: The main question is whether this is worth adding an additional dependency.

PR #64, opened 2016-02-26
Add simple `cloned()` and `inspect()` combinators

cuviper:
These work like `Iterator::cloned()` and `Iterator::inspect()`.

### DIFF
--- a/src/par_iter/map.rs
+++ b/src/par_iter/map.rs
@@ -1,7 +1,25 @@
 use super::*;
 use super::len::*;
 use super::internal::*;
-use std::iter;
+
+pub trait MapOp<In>: Sync {
+    type Output: Send;
+    fn map(&self, value: In) -> Self::Output;
+}
+
+pub struct MapFn<F>(pub F);
+
+impl<F, In, Out> MapOp<In> for MapFn<F>
+    where F: Fn(In) -> Out + Sync,
+          Out: Send,
+{
+    type Output = Out;
+    fn map(&self, value: In) -> Out {
+        (self.0)(value)
+    }
+}
+
+///////////////////////////////////////////////////////////////////////////
 
 pub struct Map<M, MAP_OP> {
     base: M,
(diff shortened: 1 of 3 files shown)


cuviper:
Added the weight, but held off on trying `Cloned` for now.  Let me know if you want it after all.

I'm also thinking of adding `fn inspect`, but that really will need a new type AFAICT.

nikomatsakis:
@cuviper well, I have no problem with implementing `Cloned` as its own specialized iterator type, but I'm also happy to stick with this version for now and implement that later. Another option that occurred to me is that we might add a "layer of indirection" to make this easier in the future.

Specifically, instead of having the "map" iterators take a `MAP_OP: Fn(In) -> Out` directly, they could take a `MAP_OP: MapFn<In, Out>` and we could have wrapper type `ClosureMapFn(F)` that implements MapFn<In,Out>` for any `F: Fn(In) -> Out`. The reason for this indirection is that then we could have a type `struct Cloner` that implements `MapFn<&'a T, T>` for all T: Clone, basically.

nikomatsakis:
I think I like this `MapOp` idea. I suspect there will be other cases of convenience style functions that can be implemented on top of `Map`.

nikomatsakis:
OK, so this branch seems like a good first step, so I'm happy to rebase/merge it and then try to do this `MapOp` idea I had as a follow-up -- or just do it all at once. Let me know. (But the branch needs a rebase anyhow.)

cuviper:
I'd like to try your idea, if you're not already working on it.

nikomatsakis:
No I am not likely to have much time for coding this next few days

cuviper:
OK, done -- let me know what you think of this...

src/par_iter/map.rs
```diff
@@ -14,12 +56,11 @@ impl<M, MAP_OP> Map<M, MAP_OP> {
     }
 }
 
-impl<M, MAP_OP, R> ParallelIterator for Map<M, MAP_OP>
```
nikomatsakis: Oh, nice that we can drop the R parameter. That always bugged me.

src/par_iter/map.rs
```diff
@@ -112,16 +148,34 @@ impl<'m, P, MAP_OP, RET> Producer for MapProducer<'m, P, MAP_OP>
...
-    type IntoIter = iter::Map<P::IntoIter, &'m MAP_OP>;
+    type Item = MAP_OP::Output;
+    type IntoIter = MapIter<'m, P::IntoIter, MAP_OP>;
 
     fn into_iter(self) -> Self::IntoIter {
-        self.base.into_iter().map(self.map_op)
+        MapIter {
+            base: self.base.into_iter(),
+            map_op: self.map_op,
+        }
```
nikomatsakis: I guess having to write our own version of `map` is the casualty here, huh? Seems like a good trade-off though.

nikomatsakis:
Looks great!

PR #65, opened 2016-02-26
Fix the copypasta docs of map/filter/filter_map/flat_map

cuviper:


cuviper:
I'm no wordsmith, but this has to be better than what was there... :)

nikomatsakis:
Indeed

PR #68, opened 2016-03-06
demo: add a prime sieve

cuviper:


cuviper:
On my i7-4600U, 4 logical CPUs, I get:

```
  serial: 4579387556 ns
  chunks: 1381673098 ns -> 3.31x speedup
parallel:  715523031 ns -> 1.93x speedup
```

demo/sieve/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+// https://oeis.org/A006880
+const NUM_PRIMES: usize = 50_847_534;
+
+// For all of these sieves, sieve[i]==true -> 2*i+1 is prime
+
+/// Sieve odd integers for primes < max.
+fn sieve_serial(max: usize) -> Vec<bool> {
+    let mut sieve = vec![true; max / 2];
+    sieve[0] = false; // 1 is not prime
+    for i in 1 .. {
```
lilianmoraru: I didn't know that you can do this in Rust. `1 ..`
Won't this force bounds checking(I don't know if the compiler gets the information it needs in this situation)?

demo/sieve/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+//! multiple chunks, for the separate jobs in each thread.
+
+extern crate itertools;
+extern crate rayon;
+extern crate time;
+
+use itertools::StrideMut;
+use rayon::prelude::*;
+
+const CHUNK_SIZE: usize = 100_000;
```
lilianmoraru: I wonder if chunks <= 64 won't be better for cache.

demo/sieve/src/main.rs
cuviper: Yeah, you could do a square root and determine that break point ahead of time, but I suspect it won't matter much to performance.  Those branches should be very easy for the CPU branch predictor, and this is not a hotspot.  The code in `clear_stride` has more impact.  But please feel free to try some optimizations yourself!

demo/sieve/src/main.rs
cuviper: Are you thinking to fit the chunk in a single cache line?  Remember that the small-prime iteration in `update_chunk` is repeated for every chunk, so there's a balance to amortize that cost.  If you work too small, the larger primes will skip right over a lot of chunks, making that wasted effort.

At 100,000, I can fit the chunks from two hyperthreads into their shared 256K L2 cache, with room to spare.  I think I tried smaller like 10,000 to fit in the 32K L1 cache, and that was a little slower overall, but I don't recall by how much.

lilianmoraru: Aham, I guess it is right that you should always measure :)
Good idea on splitting chunks so that they can fit in L1 or L2.
Could you try big chunks to fit in L2 but a multiplier of 64?
I think I will take my ARM boards out and try it myself.

cuviper:
FWIW, that i7-4600U is running RHEL7.2.  Here are a few more machines I have handy:

Fedora 23 i7-2600 (quad core + HT = 8 threads):
```
  serial: 5608365452 ns
  chunks: 1402309271 ns -> 4.00x speedup
parallel:  435526211 ns -> 3.22x speedup
```

Windows 10 i7-3770K (quad core + HT = 8 threads):
```
  serial: 4429790076 ns
  chunks: 1365959939 ns -> 3.24x speedup
parallel:  407180117 ns -> 3.35x speedup
```

So it seems to scale closer to the number of cores, with hyperthreading not adding much.  Meh.

demo/sieve/src/main.rs
lilianmoraru: Btw, it actually makes sense to me now. That's the main idea, you want to get chunks from RAM as rare as you could. So, basically, if you know your L2/L3 cache, you want to get the biggest chunk possible that can fit in there(taking in consideration that there are also other things happening on the system that also uses the cache).

nikomatsakis:
Very nice.

nikomatsakis:
Interesting. Running on my 8-core machine, I get the following speedups:

```
     Running `target/release/sieve`
  serial: 4696420262 ns
  chunks: 1735826028 ns -> 2.71x speedup
parallel:  541524724 ns -> 3.21x speedup
lunch-box. cargo run --release
     Running `target/release/sieve`
  serial: 4701838510 ns
  chunks: 1770446016 ns -> 2.66x speedup
parallel:  543002545 ns -> 3.26x speedup
```

nikomatsakis:
Might be nice to extend the demo to allow specifying the various parameters (e.g., chunk size, number of primes to find, maybe the weight) using command line options.

nikomatsakis:
Perhaps I merged a bit early. I found that changing the CHUNK_SIZE to 1000 completely changes the equation:

```
     Running `target/release/sieve`
  serial: 4602140775 ns
  chunks: 27581629316 ns -> 0.17x speedup
parallel: 2839105086 ns -> 9.71x speedup
```

so I think it'd be useful to have two distinct chunk sizes, one for sequential and one for parallel.

nikomatsakis:
This seems like a good motivator, actually, for a "chunk" operation where the size of the chunks is determined dynamically (i.e., based on work stealing). I've been debating about trying to expose an operation like this. Where basically rayon tries to subdivide the work for you a bit but always gives you a series of sequential items to reproduce. 

I should add that I didn't look SUPER closely at the code so I dont' really know what the chunk size is influencing. I should read it a bit more carefully! :)

cuviper:
I'll work on some tuning options for this.

Note that the stated speedups are stacked - your surprising result with 1000 is not even 2x faster than the serial version.  It's just that chunks ran so slowly that time, and I wonder if that was a fluke.  We probably need an option for repetition, and should print the overall speedup too.

nikomatsakis:
Ah, I didn't realize they were stacked actually :)

nikomatsakis:
would be good to say "speed vs X" or something I suppose

lilianmoraru:
On a relatively busy system, i7-2630qm(4 cores, 8 threads):
```
  serial: 6861984667 ns
  chunks: 2346635775 ns -> 2.92x speedup
parallel:  948559069 ns -> 2.47x speedup
```
I am curious how ARM performs(but I have to cross-compile the new stable Rust for that)...

PR #69, opened 2016-03-06
small readme doc fix saying f32 twice instead of f32 and f64.

LiamClark:
The README.md in the section  about atomicity metions:

A Cell offers a mutable slot with just two methods, get and set. Cells can only be used for Copy types that are safe to memcpy around, such as i32, **f32,** or even something **bigger** like (usize, usize, **f32**).

it mentions the f32 primitive and something bigger than that but again mentioning f32.

i'm assuming the intent was to say f64, this micro pr fixes that.

Liam Clark

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ serve a common purpose:
 1. A `Cell` offers a mutable slot with just two methods, `get` and
    `set`.  Cells can only be used for `Copy` types that are safe to
    memcpy around, such as `i32`, `f32`, or even something bigger like
-   `(usize, usize, f32)`.
+   `(usize, usize, f64)`.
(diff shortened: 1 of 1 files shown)


cuviper:
I think "bigger" here is referring to the whole tuple of three items.
The size of the float is not important to that point.

LiamClark:
ah i didn't read it as a tuple for some reason if that's the case then indeed it doesn't matter at all

cuviper:
Maybe you could add that emphasis? Perhaps "... bigger *like a tuple of* (...)"

nikomatsakis:
Yes, @cuviper is correct. And I agree we could maybe make it a bit clearer to address the underlying confusion.

nikomatsakis:
Thanks!

PR #71, opened 2016-03-29
[WIP] introduce explicit scope API

nikomatsakis:
This PR introduces an explicit scope API to complement `join`:

```rust
scope(|scope| {
    for i in 0 .. 10 {
        scope.spawn(|scope| {
            // the parameter `scope` can be used to spawn yet more work within the same overall scope
            process(scope, i);
        })
})
```

In general, `join` is to be preferred, both because it integrates more smoothly (you don't have to define the "scope") and because it leverages stack allocation instead of heap allocation. However, there are some cases where `join` is not a good fit. For example, if you are processing a worklist of items, where each item may yield more items, scope can be very convenient:

```rust
scope(|scope| {
    for item in initial_worklist {
        scope.spawn(|scope| process(scope, item));
    }
})

fn process<'scope>(scope: &Scope<'scope>, item: Item) {
    for item in item.dependent_items {
        scope.spawn(|scope| process(scope, item));
    }
}
```

Note that this is basically a flat-map, but the number of flat-maps being performed depends on the shape of the data. 

Anyway, this is a bit of WIP: there aren't many tests and I suspect that the code is buggy in more than a couple of ways. But I thought I'd toss it up and see what people thought.


nikomatsakis:
Note that this version of the code does not have any way to "join" a task -- that'd be an obvious extension, at which point we've basically entered into "scoped futures" territory. I'm not sure what sort of contortions that might introduce internally -- quite likely it'd be fairly straight-forward though.

nikomatsakis:
OK, I updated this PR and it seems to pass tests locally. I'm inclined to merge it since it's been sitting around so long and the API is unstable anyhow.

nikomatsakis:
cc @Amanieu -- this PR series modifies some of the unwinding support. In particular, I removed the nightly feature since it is no longer needed, and I modified our behavior if an unexpected panic occurs in a worker thread (i.e., not due to the user's callback). In that case, I am now just aborting rather than attempting to terminate the thread pool. My reasoning is that terminating the thread pool did not permit a meaningful recovery, since it could never be used again, and it is a global resource. And of course the logic is simpler this way (the code doesn't change all that much, but it's easier to think about).

nikomatsakis:
Heh, I just realized that there are two big FIXMEs with this PR. First, the scope jobs don't help when they are blocked, and second, the heap jobs are leaked. Interestingly, the fix I would want to do is kind of 'sketchy' from the point of view of possible unsafe code guidelines (that is, I'd like to transmute the `&self` into a `Box<Self>`). To do the interface I really want would require supporting `*mut Self` in traits (which we ought to do, but anyhow).

nikomatsakis:
@emilio ok, I refactored things now; take a look at the [`HeapJob` struct](https://github.com/nikomatsakis/rayon/blob/9b2e682628b73e745b492fa01a6f3c6f8c94f65f/src/scope/mod.rs#L336-L341) introduced in the last commit. I think this is a better way to do it than introducing an enum (fewer branches).

nikomatsakis:
I have yet to add any scope-specific benchmarks, but just FYI the various refactorings here don't seem to have affected overall performance at all. This is the result of using `cargo-benchcmp` against our benchmarks (running on my linux machine; my mac came out comparable):

```
lunch-box. cargo-benchcmp --threshold 1 killme.before2 killme.scope
 name                                    killme.before2 ns/iter  killme.scope ns/iter  diff ns/iter   diff %
 factorial::factorial_par_iter           14,533,952              15,024,386                 490,434    3.37%
 pythagoras::euclid_parallel_full        7,097,000               7,247,884                  150,884    2.13%
 pythagoras::euclid_parallel_outer       1,322,062               1,301,973                  -20,089   -1.52%
 quicksort::bench::quick_sort_par_bench  11,093,168              10,909,666                -183,502   -1.65%
 sieve::bench::sieve_parallel            4,687,502               4,114,703                 -572,799  -12.22%
```

There are some changes, but they seem to be "in the noise" based on some repeated runs (I also screened out things less than 1%, which is **definitely** noise). Not super scientific but at least this doesn't make things grossly slower.

nikomatsakis:
Bah. The travis failure is because my home-rolled leak-check is racy. Have to think about the best fix for that.

emilio:
Sorry for not getting to see this until now. Haven't looked with too much detail, but this looks great!

Let me know if I can help somehow :)

nikomatsakis:
@emilio 

> Sorry for not getting to see this until now. Haven't looked with too much detail, but this looks great!
> Let me know if I can help somehow :)

So I was reading over https://github.com/servo/servo/pull/13641/ and it seems to me that this PR actually subsumes #103, right? That is, it is a better watch for what you are trying to do I think. In particular, [this main loop](https://github.com/emilio/servo/blob/1ddace2197e971c8842c9fdfe44a7753b899788d/components/layout/parallel.rs#L262-L271):

```rust
 queue.push_and_forget(move || {
        profile(time::ProfilerCategory::LayoutParallelWarmup, profiler_metadata,
                time_profiler_chan, move || {
            unsafe {
                assign_inline_sizes(data.nodes, &*data.context, &*data.queue);
            }
        });
    });

queue.wait_all();
```

would become something like:

```rust
queue.install(|| {
    rayon::scope(move |s| {
        profile(time::ProfilerCategory::LayoutParallelWarmup, profiler_metadata,
                time_profiler_chan, move || {
            unsafe {
                assign_inline_sizes(s, data.nodes, &*data.context);
            }
        });
    });
});
```

Note that instead of passing `data.queue` I changed it to thread the scope `s` down.

Also note that I have been wanting to change the `queue.install` mechanism for some time (currently, it moves you to a worker thread) -- but in this case it's fine, since I was thinking of altering `rayon::scope` to move you to a worker thread anyway, so it's kind of "six in one, half-a-dozen in the other", as they say.

emilio:
As far as I understand, yes, instead of passing the queue and using
`push_and_forget`, we'd be passing the scope down and using
`scope.spawn` to trigger new jobs without waiting, is that right?

If it is, yes, it completely subsumes my PR :)

nikomatsakis:
@emilio 

> As far as I understand, yes, instead of passing the queue and using
`push_and_forget`, we'd be passing the scope down and using
`scope.spawn` to trigger new jobs without waiting, is that right?

Yes.

> If it is, yes, it completely subsumes my PR :)

OK, then I will close that PR in favor of this one, if that's ok w/ you.

nikomatsakis:
@emilio 

> Let me know if I can help somehow :)

I'm still not sure what's the best way to handle the leak checking. I'm tempted to just rip it out and leave a FIXME for it. The other changes are relatively straightforward; if you want to try your hand at hacking on them, I'd be happy to merge this and then we can develop together a bit. I'd also like to get some realistic benchmarks: I was thinking already of doing a "tree traversal" one, which seems it's precisely the pattern that Servo is doing.

emilio:
> I'm still not sure what's the best way to handle the leak checking. I'm tempted to just rip it out and leave a FIXME for it.

It seems like a proper "scope does not leak any job" test could be added outside of the core of rayon, and should be more reliable that checking the leak counter in `block_till_jobs_complete`.

> if you want to try your hand at hacking on them, I'd be happy to merge this and then we can develop together a bit. I'd also like to get some realistic benchmarks: I was thinking already of doing a "tree traversal" one, which seems it's precisely the pattern that Servo is doing. 

Yep, sounds great! I already have a few patches that I made based on profiling Servo and improvements Servo had in its custom work queue, so I can post them as soon as this is merged and I can find the time to rebase on top :)

src/scope/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+    };
+    if false { scope.fool_dead_code(); }
+    let result = op(&scope);
+    scope.job_completed_ok(); // `op` counts as a job
+    scope.block_till_jobs_complete();
+    result
+}
+
+impl<'scope> Scope<'scope> {
+    /// just here to full the dead_code lint
```
emilio: nit: While re-checking the `leak_counter` code, you probably meant `fool` here :P

nikomatsakis:
> It seems like a proper "scope does not leak any job" test could be added outside of the core of rayon, and should be more reliable that checking the leak counter in block_till_jobs_complete.

Yeah, I thought this, and in fact attempted it, but it's a bit trickier thank it looks. In particular, the *closure* gets dropped always, even if we never free the underlying box. So if you just check that user-created objects don't leak, that works, even if there is a "box per spawn" being leaked (as there was in the first version of the code, since I forgot to handle it).

My ideal would be to test with valgrind, really, but unfortunately that doesn't quite seem to work with Rust at the moment. =(

nikomatsakis:
@emilio 

> Yep, sounds great! I already have a few patches that I made based on profiling Servo and improvements Servo had in its custom work queue, so I can post them as soon as this is merged and I can find the time to rebase on top :)

OK, I'll rip out the leak detection for now I think and open a followup issue or 3.

emilio:
> In particular, the closure gets dropped always, even if we never free the underlying box. So if you just check that user-created objects don't leak, that works, even if there is a "box per spawn" being leaked (as there was in the first version of the code, since I forgot to handle it).

Oh, I see, it's really tricky, hmm... I'm not sure I can find a way to do it properly off-hand :/

emilio:
> My ideal would be to test with valgrind, really, but unfortunately that doesn't quite seem to work with Rust at the moment. =(

Maybe with alloc-system or something?

nikomatsakis:
OK, I opened fixme issues for the remaining work. If travis is happy with the latest commit, I'm inclined to merge (the api is unstable, after all).

nikomatsakis:
@emilio 

> Maybe with alloc-system or something?

Yeah, something like that. I left some notes in #107.

nikomatsakis:
@emilio merged

PR #72, opened 2016-03-30
Add link to rustdoc in README

bluss:
I often come past here, and a quick link to the docs would be helpful!


nikomatsakis:
Thanks!

PR #73, opened 2016-04-09
remove unnecessary attribute: #[allow(dead_code)]

SharplEr:
Attribute: #[allow(dead_code)] in src/log.rs is unnecessary now and can be safely removed.


nikomatsakis:
Thanks!

PR #74, opened 2016-04-15
Update panic propagation for latest nightly

Amanieu:
Eventually this can be unconditionally enabled, once `catch_panic` is in stable Rust.


nikomatsakis:
Thanks!

PR #83, opened 2016-08-16
Add MIT/Apache license files.

edre:
Matches the cargo license description, and the rust-lang license scheme.


nikomatsakis:
Thanks :)

PR #85, opened 2016-08-18
demo: add mergesort

edre:
It's a lot more parallelizable than quicksort. The critical path is O(log^3 n) versus O(n).

### DIFF
--- a/demo/mergesort/src/main.rs
+++ b/demo/mergesort/src/main.rs
@@ -128,6 +128,52 @@ fn test_merge_sort() {
     assert_eq!(sorted, v);
 }
 
+pub fn seq_merge_sort<T: Ord + Copy>(v: &mut [T]) {
+    let n = v.len();
+    let mut buf = Vec::with_capacity(n);
+    // We always overwrite the buffer before reading to it, and we want to
+    // duplicate the behavior of parallel sort.
+    unsafe {
+        buf.set_len(n);
+    }
+    seq_sort(v, &mut buf[..]);
+}
+
+// Sort src, possibly making use of identically sized buf.
+fn seq_sort<T: Ord + Copy>(src: &mut [T], buf: &mut [T]) {
+    if src.len() <= SORT_CHUNK {
+        src.sort();
+        return;
+    }
+
+    // Sort each half into half of the buffer.
+    let mid = src.len() / 2;
+    let (bufa, bufb) = buf.split_at_mut(mid);
+    {
+        let (sa, sb) = src.split_at_mut(mid);
+        seq_sort_into(sa, bufa);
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Looks nice; it'd be great if we could have a seq/par version for comparison purposes.

edre:
Would the purpose of the seq version be to assess the rayon overhead? This algorithm is strictly less efficient than a normal sequential mergesort, because of the searching to see where it can evenly divide up merging work for subtasks.

I could write a similar version that just replaces parallel merge with sequential merge. There would be less splitting and joining than the current algorithm.

nikomatsakis:
Personally, I'd prefer to see all 3 -- one for assessing rayon overhead, one for parallel, and one for seq overhead. But the first two are most important to me. The reason is yes that we can assess overhead and also use as the basis for evaluating optimizations to `join()`.

edre:
Added sequential version.

nikomatsakis:
Argh, sorry, been busy and didn't notice the update here! I'll give this a run locally.

demo/mergesort/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+        let (d1, d2) = dest.split_at_mut(mid);
+        rayon::join(|| rsort(s1, d1), || rsort(s2, d2));
+    }
+
+    // Merge the halves into dest.
+    rmerge(s1, s2, dest);
+}
+
+// Merge sorted inputs a and b, putting result in dest.
+// TODO: Figure out how to get a and b immutable with enough template magic.
```
nikomatsakis: This is interesting. I would remove this "TODO" and replace with a comment like this:

"Note that `a` and `b` use `&mut [T]` even though they are not mutated. This is because we want to leverage the `mut` qualifier to show that we do not share elements of these lists between threads. This allows us to avoid a `T: Sync` bound as would otherwise be necessary."

nikomatsakis:
Actually, I'll just make the comment change myself. =)

demo/mergesort/src/main.rs
edre: I'm pretty sure this is a templating issue and not a mutability issue. If I manually inline i32 everywhere instead of T, it lets me pass &[i32]. Is there some reason &mut [T:Send] is Send whereas &[T:Send] is not?

cuviper: A plain reference could be copied and sent to multiple threads, so the type behind it has to be `Sync`.  A mutable reference is exclusive, so the type behind only needs to be `Send`.  This is important for types like `Cell` which have interior mutability - sharing cell references would allow a data race, so it's not `Sync`.

PR #89, opened 2016-09-05
update compiletest_rs and sync matched errors

cuviper:
Everything that's supposed to error still does, but some of the exact
messages are a little different now.


cuviper:
I didn't mean to work on nbody so much too, but clean CI is good CI, so there it is -- even stable!

nikomatsakis:
Great! Thanks!

PR #91, opened 2016-09-07
Relax IntoParallelRefIterator and IntoParallelRefMutIterator

### LINKED ISSUE #90: IntoParallelRefIterator and IntoParallelRefMutIterator are overly constrained
I was trying some implementations for #88 std collections ([WIP](https://gist.github.com/cuviper/b8d7a42fe7a901c298e2fb12c4f3a73c)), but I hit a stumbling block on `IntoParallelRefIterator` and `IntoParallelRefMutIterator`.  These both require their `Item` to be the type *behind* the reference, so the `ParallelIterator`'s `Item` is this `&Item` and `&mut Item` respectively.  That's OK for most collections, but the map iterators use `(&K, &V)` and `(&K, &mut V)`.

It would be nicer if the `Item` represented the same as the parallel iterator, so both the typical `&T` and map's `(&K, &V)` can work.  Perhaps some future collection would even like something that only emulates a reference, like a cell `Ref` or a `MutexGuard`.

With this made more general, it can also have a blanket implementation, like:

```rust
pub trait IntoParallelRefIterator<'data> {
    type Iter: ParallelIterator<Item=Self::Item>;
    type Item: Send + 'data;

    fn par_iter(&'data self) -> Self::Iter;
}

impl<'data, T: 'data + ?Sized> IntoParallelRefIterator<'data> for T
    where &'data T: IntoParallelIterator
{
    type Iter = <&'data T as IntoParallelIterator>::Iter;
    type Item = <&'data T as IntoParallelIterator>::Item;

    fn par_iter(&'data self) -> Self::Iter {
        self.into_par_iter()
    }
}
```

Users of this trait would now lose the ability to assume that `par_iter()` yields simple references, but I'm not sure that's actually needed or valuable.

Thoughts?

cuviper:
The constraints on their `Iter` no longer require them to have simple
references.  Their `Item` now represents the complete type that is
produced, which could be something like `(&K, &mut V)` on a map, for
instance.

There are also now blanket implementations for types whose references
already implement `IntoParallelIterator`, which wasn't possible before.

[breaking-change] Implementers of these traits may have to adapt to the
changed constraints, and may be forced out by the blanket impl.  Users
of the traits may be affected by the changed `Item` semantics.

Closes #90.

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -55,17 +55,39 @@ pub trait IntoParallelIterator {
 }
 
 pub trait IntoParallelRefIterator<'data> {
-    type Iter: ParallelIterator<Item=&'data Self::Item>;
-    type Item: Sync + 'data;
+    type Iter: ParallelIterator<Item=Self::Item>;
+    type Item: Send + 'data;
 
     fn par_iter(&'data self) -> Self::Iter;
 }
 
+impl<'data, I: 'data + ?Sized> IntoParallelRefIterator<'data> for I
+    where &'data I: IntoParallelIterator
+{
+    type Iter = <&'data I as IntoParallelIterator>::Iter;
+    type Item = <&'data I as IntoParallelIterator>::Item;
+
+    fn par_iter(&'data self) -> Self::Iter {
+        self.into_par_iter()
+    }
+}
+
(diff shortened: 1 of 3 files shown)


nikomatsakis:
Yeah, I'm convinced this makes sense.

PR #92, opened 2016-09-07
IntoParallelIterator for Option and Result

### LINKED ISSUE #80: `par_iter` for `Option`
It would be nice to have one to allow uses like:

```rust
paths
    .map(File::open)
    .flat_map(Result::ok)
    // ...
```

cuviper:
Fixes #80.

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -43,6 +43,7 @@ pub mod weight;
 pub mod zip;
 pub mod range;
 pub mod vec;
+pub mod option;
 
 #[cfg(test)]
 mod test;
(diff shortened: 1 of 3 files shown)


cuviper:
I'm not sure this is terribly useful, but I said in #80 that it would be easy, so here it is. :)

nikomatsakis:
Looks good -- seems like this would work best if we had `chain` attached to it.

nikomatsakis:
By "work best" I mean "be most useful". I often want to do something like "iterate over this slice and maybe one more thing".

cuviper:
Huh, there's no `ParallelIterator::chain` yet, but that definitely sounds like a good idea, and I agree it would compose nicely here.

PR #95, opened 2016-09-10
Clean up deque access

cuviper:
The first commit has the most impact, I think, moving the deque `Worker` into the `WorkerThread` instead of the shared `Registry`. This lets us drop the `unsafe impl Send`/`Sync`.

The rest are more general cleanups.  Each `WorkerThread` can keep its own vector of the other `Stealer`s, and they don't actually need to access the `Registry` at all.

### DIFF
--- a/src/thread_pool.rs
+++ b/src/thread_pool.rs
@@ -211,6 +211,7 @@ impl ThreadInfo {
 
 pub struct WorkerThread {
     worker: Worker<JobRef>,
+    stealers: Vec<Stealer<JobRef>>,
     registry: Arc<Registry>,
     index: usize
 }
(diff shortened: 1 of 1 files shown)


nikomatsakis:
@cuviper this seems pretty nice. My one concern is that it would be harder to dynamically scale the number of threads this way (which I would probably like to eventually support). But (I think) not *a lot* harder, and I tend to think the best thing would be to factor current thread-pool out into another crate anyway and iterate there.

cuviper:
I can scale it back to just the `Worker` isolation, if you prefer.

nikomatsakis:
@cuviper nah let's leave it for now. I like having things be as simple as they can be in order to support the current needs, and this setup seems simpler.

PR #96, opened 2016-09-10
Add vectorized iterators for std::collections

### LINKED ISSUE #88: impl par_iter for other std collections.
Currently if you want parallelize iterating through anything that can't easily be turned into a `Vec<T>` or `&[T]` you have perform a `collect`, which can be quite slow on large collections.

cuviper:
Fixes #88.

### DIFF
--- a/src/par_iter/collections.rs
+++ b/src/par_iter/collections.rs
@@ -1,0 +1,67 @@
+use super::IntoParallelIterator;
+use super::vec::VecIter;
+use std::collections::{BinaryHeap, BTreeMap, BTreeSet, HashMap, HashSet, LinkedList, VecDeque};
+use std::iter::FromIterator;
+
+// This could almost be replaced with a blanket implementation for *all*
+// `IntoIterator` types, but we'd need specialization a few like `Vec`.
+macro_rules! vectorized {
+    ( impl<$($c:tt),*> IntoParallelIterator for $t:ty ) => {
+        impl<$($c),*> IntoParallelIterator for $t
+            where $t: IntoIterator,
+                  <$t as IntoIterator>::Item: Send
+        {
+            type Item = <Self as IntoIterator>::Item;
+            type Iter = VecIter<Self::Item>;
+
+            fn into_par_iter(self) -> Self::Iter {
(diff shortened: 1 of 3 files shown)


nikomatsakis:
@cuviper note travis failure

nikomatsakis:
Looks good except for travis failure.

I was wondering if we would want some more general "convert iterator to parallel" notion that avoids "overhead" of creating a vector, *but*:

- that would probably have to use some sort of task queue, which is basically a vector, but probably less efficient
- we can still have that if people write `foo.iter().parallelize()` or some such

Some data structures (e.g., btrees but prob also hashmaps) could support more efficient interaction if they gave access to more of their internals, or impl'd rayon directly. Said access might not necessarily be indexable and hence it could be a backwards incompat change to adopt it. But it seems like that would be a bridge to cross then -- and trying to ensure indexable would be a good goal for such integration anyway.

cuviper:
Hmm, that issue with the lifetime parameter looks annoying.  It worked with nightly, and works with beta, so I wonder what change that is that hasn't landed on stable yet?  Anyway, I think I can fix this by forking the macro into three (almost identical) variants... :/

The `cd demo/...` failures are even weirder -- hopefully just some spurious Travis thing.

nikomatsakis:
@cuviper weird; could well be a bug, I didn't look closely at it

cuviper:
OK, Travis is happy now.

src/par_iter/collections.rs
```diff
@@ -4,21 +4,40 @@ use std::collections::{BinaryHeap, BTreeMap, BTreeSet, HashMap, HashSet, LinkedL
 use std::iter::FromIterator;
 
 // This could almost be replaced with a blanket implementation for *all*
-// `IntoIterator` types, but we'd need specialization a few like `Vec`.
+// `IntoIterator` types, but we'd need to specialize a few like `Vec`.
 macro_rules! vectorized {
-    ( impl<$($c:tt),*> IntoParallelIterator for $t:ty ) => {
+    (@impl_body) => {
```
nikomatsakis: heh, cute.

PR #97, opened 2016-09-12
Refactor VecIter unsafety so it doesn't need Drop

cuviper:
The `consumed` flag really only served to let `with_producer` avoid
unsafe code itself.  But that made an implicit safe-unsafe contract with
the drop handler, which isn't good encapsulation.

Now all of the `VecIter` unsafety is contained in `with_producer`, and
it no longer needs a `Drop` implementation at all.

### DIFF
--- a/src/par_iter/vec.rs
+++ b/src/par_iter/vec.rs
@@ -4,7 +4,6 @@ use std;
 
 pub struct VecIter<T: Send> {
     vec: Vec<T>,
-    consumed: bool,
 }
 
 impl<T: Send> IntoParallelIterator for Vec<T> {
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Seems ok. I've been wanting for some time to do something smarter that doesn't leak as badly on panic. But that's a separate thing.

cuviper:
Can you explain what leaking scenario you're worried about?  The producer drops into a `SliceDrain`, and that explicitly drops every element that didn't get iterated out.  I thought this was RAII enough to work even for panics.

nikomatsakis:
> Can you explain what leaking scenario you're worried about? The producer drops into a SliceDrain, and that explicitly drops every element that didn't get iterated out. I thought this was RAII enough to work even for panics.

Oh, hmm, maybe that IS ok. I don't remember anymore. =)

PR #98, opened 2016-09-13
Implement a chain combinator

cuviper:
Two parallel iterators of the same type act as one big iterator.  All
common properties are maintained: bounded, exact, and indexed.

`VecDeque` gets to be the first user, replacing its temporary vectors
instead with a chain over its pair of slices.

### DIFF
--- a/src/par_iter/chain.rs
+++ b/src/par_iter/chain.rs
@@ -1,0 +1,173 @@
+use super::*;
+use super::internal::*;
+use std::cmp::min;
+use std::iter;
+
+pub struct ChainIter<A, B>
+    where A: ParallelIterator, B: ParallelIterator<Item=A::Item>
+{
+    a: A,
+    b: B,
+}
+
+impl<A, B> ChainIter<A, B>
+    where A: ParallelIterator, B: ParallelIterator<Item=A::Item>
+{
+    pub fn new(a: A, b: B) -> ChainIter<A, B> {
+        ChainIter { a: a, b: b }
+    }
+}
+
+impl<A, B> ParallelIterator for ChainIter<A, B>
+    where A: ParallelIterator, B: ParallelIterator<Item=A::Item>
+{
+    type Item = A::Item;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+        where C: UnindexedConsumer<Self::Item>
+    {
+        let a = self.a.drive_unindexed(consumer.split_off());
+        let b = self.b.drive_unindexed(consumer.split_off());
+        consumer.to_reducer().reduce(a, b)
+    }
+}
+
+impl<A, B> BoundedParallelIterator for ChainIter<A, B>
+    where A: BoundedParallelIterator, B: BoundedParallelIterator<Item=A::Item>
+{
+    fn upper_bound(&mut self) -> usize {
+        self.a.upper_bound() + self.b.upper_bound()
+    }
+
+    fn drive<C>(mut self, consumer: C) -> C::Result
+        where C: Consumer<Self::Item>
+    {
+        let (left, right, reducer) = consumer.split_at(self.a.upper_bound());
+        let a = self.a.drive(left);
+        let b = self.b.drive(right);
+        reducer.reduce(a, b)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


src/par_iter/chain.rs
```diff
@@ -0,0 +1,173 @@
...
+    where A: Producer, B: Producer<Item=A::Item>
+{
+    fn cost(&mut self, len: usize) -> f64 {
+        let a_len = min(self.a_len, len);
+        let b_len = len - a_len;
+        self.a.cost(a_len) + self.b.cost(b_len)
+    }
+
+    fn split_at(self, index: usize) -> (Self, Self) {
+        if index <= self.a_len {
```
nikomatsakis: Heh. It took me a while to see why you were always creating two chain-producers... but of course, we have to comply with the trait. Clever.

src/par_iter/collections.rs
```diff
@@ -80,7 +83,23 @@ vectorized!{ impl<'a, T> IntoParallelIterator for &'a mut LinkedList<T> }
...
-// TODO - we could easily write these last two directly, no temp vector,
-// with `as_slices`/`as_mut_slices` and a `chain` combinator.
+
+impl<'a, T: Sync> IntoParallelIterator for &'a VecDeque<T> {
+    type Item = &'a T;
+    type Iter = ChainIter<SliceIter<'a, T>, SliceIter<'a, T>>;
+
+    fn into_par_iter(self) -> Self::Iter {
+        let (a, b) = self.as_slices();
+        a.into_par_iter().chain(b)
```
nikomatsakis: nice!

src/par_iter/mod.rs
```diff
@@ -363,6 +365,16 @@ pub trait ParallelIterator: Sized {
         reduce(self, reduce_op)
     }
 
+    /// Takes two iterators and creates a new iterator over both.
+    ///
+    /// As with most parallel iterators, the order in which these are
+    /// combined is completely unspecified.
```
nikomatsakis: Hmm. I am not sure how to interpret this comment. I would expect (as a user) that if you do `a.chain(b)` then you get `a` first and then `b` -- e.g., if we use `enumerate()`. Is this not true? (I think it is, from the impl)

nikomatsakis:
Very cool!

nikomatsakis:
My only question is the comment, which seems a bit misleading, or maybe I'm confused.

src/par_iter/mod.rs
cuviper: Ah right.  For things like `enumerate`, `zip`, and `collect`, the items are ordered `a` then `b`.  I was just trying to parrot the normal parallel hazard that there's no *temporal* order to visitation.  Maybe I should s/combined/visited/ ?

src/par_iter/chain.rs
cuviper: TBH I'm not thrilled with that, since I have to keep splitting at the boundary to get producers that I know will be empty.  I also tried wrapping them in `Option`, but that made it a lot more difficult to deal with.  So this is a little weird, but it keeps things simple.

src/par_iter/mod.rs
nikomatsakis: I feel like I'd just drop it; It feels like it kind of goes w/o saying? but yeah something like "order in which these will be visited when folding/collecting/etc"

cuviper: OK, I just dropped it.

src/par_iter/chain.rs
```diff
@@ -0,0 +1,173 @@
...
+
+impl<A, B> ParallelIterator for ChainIter<A, B>
+    where A: ParallelIterator, B: ParallelIterator<Item=A::Item>
+{
+    type Item = A::Item;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+        where C: UnindexedConsumer<Self::Item>
+    {
+        let a = self.a.drive_unindexed(consumer.split_off());
```
lilianmoraru: I am always [scared of segfaults](https://users.rust-lang.org/t/rust-guarantees-no-segfaults-with-only-safe-code-but-it-segfaults-stack-overflow/4305) when I see recursion...

PR #103, opened 2016-10-07
Add the ability to push deferred jobs to the queue.

emilio:
This is the API that Servo needs to be able to run with Rayon.

There are a few things missing, mainly the ability to catch panics and tests. Also, not really happy with the names, but I'm submitting it here for feedback and so Servo people have context on how this looks like.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,12 +1,12 @@
-use latch::{LockLatch, SpinLatch};
+use latch::{Latch, LockLatch, SpinLatch};
 #[allow(unused_imports)]
 use log::Event::*;
-use job::JobImpl;
+use job::{Job, JobImpl, DeferredJobImpl};
 use std::sync::Arc;
 use std::error::Error;
 use std::fmt;
 use thread_pool::{self, Registry, WorkerThread};
-use std::{thread, mem};
+use std::mem;
 
 /// Custom error type for the rayon thread pool configuration.
 #[derive(Debug,PartialEq)]
@@ -157,9 +157,7 @@ pub fn join<A,B,RA,RB>(oper_a: A,
             fn drop(&mut self) {
                 unsafe {
                     if !(*WorkerThread::current()).pop() {
-                        while !self.0.probe() {
-                            thread::yield_now();
-                        }
+                        self.0.wait();
                     }
                 }
             }
@@ -203,8 +201,8 @@ unsafe fn join_inject<A,B,RA,RB>(oper_a: A,
 
     thread_pool::get_registry().inject(&[job_a.as_job_ref(), job_b.as_job_ref()]);
 
-    job_a.latch.wait();
-    job_b.latch.wait();
+    job_a.wait();
+    job_b.wait();
 
     (job_a.into_result(), job_b.into_result())
 }
@@ -232,9 +230,24 @@ impl ThreadPool {
         unsafe {
             let job_a = JobImpl::new(op, LockLatch::new());
             self.registry.inject(&[job_a.as_job_ref()]);
-            job_a.latch.wait();
+            job_a.wait();
             job_a.into_result()
         }
+    }
+
+    /// Sends `op` to the work queue to be eventually executed.
+    pub fn start<'a, OP>(&'a self, op: OP)
+        where OP: FnOnce() -> () + Send + 'static
+    {
+        let job = Box::new(DeferredJobImpl::new(op, LockLatch::new()));
+        unsafe {
+            self.registry.inject_deferred(job);
+        }
+    }
+
+    /// Waits until there are no more jobs in the work queue.
+    pub fn wait_all(&self) {
+        self.registry.wait_all()
     }
 }
 
--- a/src/thread_pool.rs
+++ b/src/thread_pool.rs
@@ -1,7 +1,7 @@
 use Configuration;
 use deque;
 use deque::{Worker, Stealer, Stolen};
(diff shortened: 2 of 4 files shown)


src/thread_pool.rs
```diff
@@ -116,6 +139,18 @@ impl Registry {
         self.work_available.notify_all();
     }
 
+    pub unsafe fn inject_deferred(&self, job: Box<Job>) {
+        log!(InjectJobs { count: 1 });
+        let mut state = self.state.lock().unwrap();
+        if state.terminate {
+            drop(state);
+            // See the comment in inject about this case.
+            panic!("rayon thread pool is contaminated by a previous panic; recovery is only available on nightly compilers");
```
lilianmoraru: `recovery is only available on nightly compilers` - is this talking about the `std` panic or Rayon's propagation of the panics?
If it is the std, it is available since Rust 1.9(Stable).

emilio: It was pretty much a copy-paste of the other `inject`, over all taking into account that I haven't implemented panic recovery for this code path. But yeah, it's only about Rayon's recovery.

notriddle:
"hability"?

emilio:
Sorry, I'm Spanish and apparently I'm really bad at English (in Spanish
that word has an h). Will fix :)

notriddle:
No problem.

nikomatsakis:
Interesting. This is quite distinct from https://github.com/nikomatsakis/rayon/pull/71, which is the new API I've been contemplating. But it seems like a harmless addition. Will try and read and give more detailed feedback today.

src/api.rs
```diff
@@ -232,10 +230,27 @@ impl ThreadPool {
...
         }
     }
+
+    /// Sends `op` to the work queue to be eventually executed.
+    pub fn push_and_forget<'a, OP>(&'a self, op: OP)
+        where OP: FnOnce() -> () + Send + 'static
+    {
+        unsafe {
+            // We transmute the closure to a 'static closure in order to be able
+            // to store it as a box, knowing that the job won't outlive us.
```
nikomatsakis: I don't follow this comment. I don't see a transmute here, for example, and i'm not sure who "us" is in this sentence (it seems like the job may well outlive the current frame, for example).

src/api.rs
```diff
@@ -203,8 +201,8 @@ unsafe fn join_inject<A,B,RA,RB>(oper_a: A,
 
     thread_pool::get_registry().inject(&[job_a.as_job_ref(), job_b.as_job_ref()]);
 
-    job_a.latch.wait();
-    job_b.latch.wait();
+    job_a.wait();
```
nikomatsakis: Why did the `wait()` method move onto the `Task`? Is this just general cleanup or is it needed for some other reason?

src/api.rs
```diff
@@ -232,10 +230,27 @@ impl ThreadPool {
...
             let job_a = JobImpl::new(op, LockLatch::new());
             self.registry.inject(&[job_a.as_job_ref()]);
-            job_a.latch.wait();
+            job_a.wait();
             job_a.into_result()
         }
     }
+
+    /// Sends `op` to the work queue to be eventually executed.
+    pub fn push_and_forget<'a, OP>(&'a self, op: OP)
```
nikomatsakis: Indeed I don't like the name very much. Not sure what name I prefer, maybe `spawn`, `fork`, or `start`? I don't think we generally use terms like `push` in our API to refer to starting threads, do we?

src/thread_pool.rs
nikomatsakis: Now that recover is on stable, we should fix that.

src/thread_pool.rs
```diff
@@ -164,16 +199,30 @@ impl Registry {
                 return Work::None;
             }
 
+            self.worker_idle.notify_all();
```
nikomatsakis: It's probably worth a comment in the `state.injected_jobs.pop_front()` section on line 192 above indicating why we choose not to notify `worker_idle` -- in particular, although popping from `injected_jobs` may cause the `worker_idle` condition to become true, incrementing `threads_at_work` will ensure it is false.

src/api.rs
emilio: It's mostly cleanup, but it's needed since the `wait_all` method needs to call `wait()` on the tasks.

src/api.rs
emilio: Yep, `start` sounds like a way better name :)

nikomatsakis:
@emilio so this overlaps somewhat with some of the changes I've been making in #71 (you inspired me to try and complete that PR, as well.) 

I'm going to try and factor out a few things from #71 into separate PRs so we can land them first:

- refactoring how we handle panic
- refactoring our benchmark infrastructure to make it easier to run
- refactoring the `Job` trait

The last one is interesting. Instead of having `JobRef` be an `enum`, what I did in #71 (which similarly has to have heap-allocated jobs) is to have the `Job` free *itself* when it is executed or aborted. This makes sense because every job will either be executed or aborted exactly once, I believe. The only problem here is that those methods take `&self`, so they wind up transmuting to a `Box<Self>` which is mega-sketchy. So I want to refactor `Job` to pass around a `*mut` instead. Let me see if I can get it done this morning (I tend to only have a spare hour or two to hack on Rayon etc)

emilio:
Closed in favor of #71, which introduces a much nicer API for this :)

PR #104, opened 2016-10-14
Consolidate benchmarks and demos

nikomatsakis:
This brings all the demos and benchmarks into one project, `rayon-demo`. This is nice because it allows for bigger demos (like `nbody`) that span multiple files. It also means we can run the full suite of benchmarks just by doing:

```
> cd rayon-demo
> cargo bench
```

We can then use [cargo-benchcmp](https://github.com/BurntSushi/cargo-benchcmp) to get a nice "before/after" snapshot.

The idea is that each benchmark has its own independent CLI, so you can do `cargo run --release -- nbody --help` and that sort of thing. I cheated on the fibonacci, pythagoras, and factorial benchmarks and just gave them `#[bench]` instances.

cc @cuviper -- those are your benchmarks I cheated on :)


nikomatsakis:
One problem: the naming scheme leaves something to be desired. For example, the modules already tell you what we are benchmarking, so something like `nbody::bench::nbody_par`. For another, we don't use the `_par` vs `_seq` suffix consistently enough, which is a drag (I'd like to be able to do `cargo bench _par`, for example, and get all tests that have anything to do with parallel execution). So I think it should be something like `nbody::bench::par`, `nbody::bench::par_reduce`, and `nbody::bench::seq`; similarly `quicksort::bench::par` vs `quicksort::bench::seq`, and so forth.

I have to check on this, but [cargo-benchcmp](https://github.com/BurntSushi/cargo-benchcmp) allows you to compare multiple tests **within** a benchmark run, if they share a common *prefix*. So hopefully we can use these to compare seq vs par for the same benchmark.

nikomatsakis:
Gonna be annoying to keep all these open PRs, decided to just merge this since it seems obviously better.

PR #106, opened 2016-10-14
RFC: adaptive "thief splitting" algorithm for par_iter

cuviper:
This introduces an adaptive algorithm for splitting par_iter jobs, used
only as a default case when no weights are specified.  Initially, it
will split into enough jobs to fill every thread.  Then whenever a job
is stolen, that job will again be split enough for every thread.

This is roughly based on @Amanieu's description in the users forum of
the algorithm used by Async++ and TBB.

### DIFF
--- a/src/par_iter/internal.rs
+++ b/src/par_iter/internal.rs
@@ -6,6 +6,7 @@
 use join;
 use super::IndexedParallelIterator;
 use super::len::*;
+use thread_pool::get_registry;
 
 pub trait ProducerCallback<ITEM> {
     type Output;
@@ -17,6 +18,9 @@ pub trait ProducerCallback<ITEM> {
 /// not queryable through the API; the consumer is expected to track
 /// it.
 pub trait Producer: IntoIterator + Send + Sized {
+    /// Reports whether the producer has explicit weights.
+    fn weighted(&self) -> bool { false }
+
     /// Cost to produce `len` items, where `len` must be `N`.
     fn cost(&mut self, len: usize) -> f64;
 
@@ -30,6 +34,9 @@ pub trait Consumer<Item>: Send + Sized {
     type Folder: Folder<Item, Result=Self::Result>;
     type Reducer: Reducer<Self::Result>;
     type Result: Send;
+
+    /// Reports whether the consumer has explicit weights.
+    fn weighted(&self) -> bool { false }
 
     /// If it costs `producer_cost` to produce the items we will
     /// consume, returns cost adjusted to account for consuming them.
@@ -69,6 +76,75 @@ pub trait UnindexedConsumer<ITEM>: Consumer<ITEM> {
     fn to_reducer(&self) -> Self::Reducer;
 }
 
+/// A splitter controls the policy for splitting into smaller work items.
+trait Splitter: Send + Sized {
+    fn needs_split(&self) -> bool;
+    fn split_off(self) -> (Self, Self);
+}
+
+/// Classic cost-splitting uses weights to split until below a threshold.
+#[derive(Clone, Copy)]
+struct CostSplitter(f64);
+
+impl Splitter for CostSplitter {
+    #[inline]
+    fn needs_split(&self) -> bool {
+        self.0 > THRESHOLD
+    }
+
+    #[inline]
+    fn split_off(mut self) -> (Self, Self) {
+        self.0 /= 2.0;
+        (self, self)
+    }
+}
+
+/// Thief-splitting is an adaptive policy that starts by splitting into enough
+/// jobs for every worker thread, and then resets itself whenever a job is
+/// actually stolen into a different thread.
+#[derive(Clone, Copy)]
+struct ThiefSplitter {
+    origin: usize,
+    splits: usize,
+}
+
+impl ThiefSplitter {
+    #[inline]
+    fn id() -> usize {
+        // The actual `ID` value is irrelevant.  We're just using its TLS
+        // address as a unique thread key, faster than a real thread-id call.
+        thread_local!{ static ID: bool = false; }
+        ID.with(|id| id as *const bool as usize )
+    }
+
+    #[inline]
+    fn new() -> ThiefSplitter {
+        ThiefSplitter {
+            origin: ThiefSplitter::id(),
+            splits: get_registry().num_threads(),
+        }
+    }
+}
+
+impl Splitter for ThiefSplitter {
+    #[inline]
+    fn needs_split(&self) -> bool {
+        self.splits > 0 || self.origin != ThiefSplitter::id()
+    }
+
+    #[inline]
+    fn split_off(mut self) -> (Self, Self) {
+        let id = ThiefSplitter::id();
+        if self.origin != id {
+            self.origin = id;
+            self.splits = get_registry().num_threads();
+        } else {
+            self.splits /= 2;
+        }
+        (self, self)
+    }
+}
+
 pub fn bridge<PAR_ITER,C>(mut par_iter: PAR_ITER,
                              consumer: C)
                              -> C::Result
@@ -90,28 +166,35 @@ pub fn bridge<PAR_ITER,C>(mut par_iter: PAR_ITER,
         fn callback<P>(mut self, mut producer: P) -> C::Result
             where P: Producer<Item=ITEM>
         {
-            let producer_cost = producer.cost(self.len);
-            let cost = self.consumer.cost(producer_cost);
-            bridge_producer_consumer(self.len, cost, producer, self.consumer)
+            if producer.weighted() || self.consumer.weighted() {
+                let producer_cost = producer.cost(self.len);
+                let cost = self.consumer.cost(producer_cost);
+                let splitter = CostSplitter(cost);
+                bridge_producer_consumer(self.len, splitter, producer, self.consumer)
+            } else {
+                let splitter = ThiefSplitter::new();
+                bridge_producer_consumer(self.len, splitter, producer, self.consumer)
+            }
         }
     }
 }
 
-fn bridge_producer_consumer<P,C>(len: usize,
-                                 cost: f64,
-                                 producer: P,
-                                 consumer: C)
-                                 -> C::Result
-    where P: Producer, C: Consumer<P::Item>
+fn bridge_producer_consumer<S,P,C>(len: usize,
+                                   splitter: S,
+                                   producer: P,
+                                   consumer: C)
+                                   -> C::Result
+    where S: Splitter, P: Producer, C: Consumer<P::Item>
 {
-    if len > 1 && cost > THRESHOLD {
+    if len > 1 && splitter.needs_split() {
         let mid = len / 2;
+        let (left_splitter, right_splitter) = splitter.split_off();
         let (left_producer, right_producer) = producer.split_at(mid);
         let (left_consumer, right_consumer, reducer) = consumer.split_at(mid);
         let (left_result, right_result) =
-            join(|| bridge_producer_consumer(mid, cost / 2.0,
+            join(|| bridge_producer_consumer(mid, left_splitter,
                                              left_producer, left_consumer),
-                 || bridge_producer_consumer(len - mid, cost / 2.0,
+                 || bridge_producer_consumer(len - mid, right_splitter,
                                              right_producer, right_consumer));
         reducer.reduce(left_result, right_result)
     } else {
--- a/src/par_iter/internal.rs
+++ b/src/par_iter/internal.rs
@@ -77,40 +77,20 @@ pub trait UnindexedConsumer<ITEM>: Consumer<ITEM> {
 }
 
 /// A splitter controls the policy for splitting into smaller work items.
-trait Splitter: Send + Sized {
-    fn needs_split(&self) -> bool;
-    fn split_off(self) -> (Self, Self);
-}
-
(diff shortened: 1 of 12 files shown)


nikomatsakis:
Neat, can't wait to take a look! I'm curious, can you rebase atop #104 so I can try doing some comparisons? =)

cuviper:
I've been sitting on this for a while now, intending to do more rigorous testing.  But I haven't found the time for that, and I fear I may sit on it forever, perfect being the enemy of the good, so here it is.

The changes to the euclid case show that now a default (totally unweighted) case performs just as well as the best of the manually weighted cases.  The nbody `flat_map` got a little slower though, not sure why.

cuviper:
Sure, I'll try rebasing.  Just the benchmarks tweaks should need updating.

cuviper:
OK, this is now over #104.

nikomatsakis:
I had a similar branch but I wasn't happy with the benchmark results. I'm curious to compare your approach versus mine.

nikomatsakis:
@cuviper just merged #104, maybe rebase over master one last time? :)

cuviper:
Voila!

nikomatsakis:
Nice, thanks. The basic algorithm you've got here looks pretty similar to [my "adaptive" branch](https://github.com/nikomatsakis/rayon/tree/adaptive), which makes sense since I was also trying the same trick. =) I've been meaning to read more into what TBB does. 

Anyway, I did some comparison between the master and your branch. This is the results. Definitely nbody par-reduce is hurting! It'd be interesting to see why -- it's the only benchmark we have that makes use of more than a trivial parallel iterator, I think (we need more of those!).

```
lunch-box. cargo-benchcmp --threshold 2 killme.before1 killme.cuviper2
 name                               killme.before1 ns/iter  killme.cuviper2 ns/iter  diff ns/iter   diff %
 factorial::factorial_join          14,201,653              14,494,568                    292,915    2.06%
 factorial::factorial_recursion     12,783,874              16,496,116                  3,712,242   29.04%
 fibonacci::fibonacci_iterative     7                       6                                  -1  -14.29%
 fibonacci::fibonacci_join_1_2      35,207,349              34,402,348                   -805,001   -2.29%
 fibonacci::fibonacci_join_2_1      31,778,174              30,816,077                   -962,097   -3.03%
 nbody::bench::nbody_parreduce      11,951,304              95,952,256                 84,000,952  702.86%
 pythagoras::euclid_faux_serial     8,658,208               36,898,243                 28,240,035  326.16%
 pythagoras::euclid_parallel_full   7,165,959               27,046,245                 19,880,286  277.43%
 pythagoras::euclid_parallel_outer  1,455,743               4,932,806                   3,477,063  238.85%
 pythagoras::euclid_serial          10,498,747              34,572,888                 24,074,141  229.30%
 sieve::bench::sieve_parallel       3,519,063               3,062,748                    -456,315  -12.97%```

File: src/par_iter/internal.rs [hunk 1]
```diff
@@ -69,6 +76,75 @@ pub trait UnindexedConsumer<ITEM>: Consumer<ITEM> {
...
+    splits: usize,
+}
+
+impl ThiefSplitter {
+    #[inline]
+    fn id() -> usize {
+        // The actual `ID` value is irrelevant.  We're just using its TLS
+        // address as a unique thread key, faster than a real thread-id call.
+        thread_local!{ static ID: bool = false; }
+        ID.with(|id| id as *const bool as usize )
```
nikomatsakis: Ah, this is cute. I was grabbing the worker's thread index, but that was a bit annoying since sometimes we were not on a worker, so I had to force us to get into a worker before-hand. I was wondering how you avoided that.

(An interesting thing to think about also is whether we can incorporate some simple version of this into  `rayon::join` -- or probably a wrapper around it -- so that algorithms like quicksort could benefit.)

src/par_iter/internal.rs
```diff
@@ -90,28 +166,35 @@ pub fn bridge<PAR_ITER,C>(mut par_iter: PAR_ITER,
         fn callback<P>(mut self, mut producer: P) -> C::Result
             where P: Producer<Item=ITEM>
         {
-            let producer_cost = producer.cost(self.len);
-            let cost = self.consumer.cost(producer_cost);
-            bridge_producer_consumer(self.len, cost, producer, self.consumer)
+            if producer.weighted() || self.consumer.weighted() {
```
nikomatsakis: Just from a pure API perspective, I am wondering about the sharp transition between "weight" and "not weighted". It is perhaps the right thing to do, but it seems to (very mildly) erode the benefits of "weight vs threshold", since if you add a weight to one subpart, you affect all the parts. But usually people build up their parallel iterators as a unit so that doesn't feel like a big deal.

cuviper:
I'm surprised at your losses in pythagoras, but it's *really* suspicious that `euclid_serial` changed -- that doesn't use rayon at all!

src/par_iter/internal.rs
cuviper: I actually started out that way, making a join with a closure parameter that indicated stealing.  I have that in a branch somewhere if you'd like to see it.  However, the way I had it was penalizing all joins, which I didn't like.

src/par_iter/internal.rs
cuviper: Hmm, I'm not sure how you would mix these two policies though.

cuviper:
Oh!  I do know why pythagoras got worse -- I doubled the iteration size to give me a bigger workload!  But that does make comparisons to master harder, sorry...

src/par_iter/internal.rs
nikomatsakis: I was imagining something like this. You call a function in another crate like:

```rust
fn pythagoras_computation() -> impl ParallelIterator<Item=f64> {...}
```

and you don't know that they have put some weights on it.

nikomatsakis:
@cuviper heh :) I was looking a bit more closely, I see that nbody is *weighted*, also. I am doing some more measurements and comparisons, but I've got to run just now so won't get them done this morning.

nikomatsakis:
also, re: euclid-serial, there is definitely drift from run to run. I did 3 runs and compared each with each, looking for things that remain consistent, but it's tough to be sure.

cuviper:
Well, nbody parreduce is weighted on the outer loop only, not the inner fold.  But I tried adding a `weight(1.0)` inside to avoid the adaptive algorithm, and it still suffers.  It seems my code is not as transparent for the weighted-cost case as I had hoped, so I need to figure out why.

src/par_iter/internal.rs
cuviper: Well, I get how you might acquire hidden weights - I meant that I don't know how we could actually implement a mixed policy.  We could probably add something to explicitly request the adaptive algorithm, to avoid surprise weights, but I need to think how to make that work...

cuviper:
It appears to be an optimization/inlining problem.  The nbody parreduce profile has two hotspots:

```
  61.35%  rayon-demo  rayon-demo  [.] _ZN137_$LT$rayon..par_iter..fold..FoldFolder$LT$$u27$r$C$$u20$I$C$$u20$FOLD_OP$GT$$u20$as$u20$rayon..par_iter..internal..Folder$LT$ITEM$GT$$GT$7consume17h4962b75898dfb3ebE
  31.41%  rayon-demo  rayon-demo  [.] _ZN5rayon8par_iter8internal24bridge_producer_consumer17hff07c98fb07ee489E
```

But on master, that `consume` doesn't exist, and the profile is all in `bridge_producer_consumer`.

As I mentioned, I tried `weight(1.0)` in `next_velocity_par`, so it should totally use `CostSplitter`, and performance still suffers.  But if I add a `panic!` right before `let splitter = ThiefSplitter::new()`, cutting out that whole branch, suddenly everything is inlined again and performance is back on par.

It's sad, because `producer.weighted() || self.consumer.weighted()` should be statically known, but that's probably only after inlining decisions have already been made to allow that const-prop.  Perhaps it could do better if `weighted()` were instead an associated constant, not sure.

I have some ideas how to make the splitter an associated type, so this decision gets made purely in type resolution.  I'll keep playing...

nikomatsakis:
Really interesting. Another question: can you make a variant that uses no weights and compare the performance of that with the weighted version as well? (I haven't had a chance to do anything yet :(

nikomatsakis:
I think using an associated type is totally doable though; good idea.

cuviper:
Do you have any type wizardry for joining/resolving two associated types? I'm having trouble making chain and zip work, as well as combining the producer to the consumer. If either of the two has CostSplitter, I want that to win, otherwise the combination should use ThiefSplitter.

cuviper:
The other thing I tried was making Splitter an enum, losing the trait monomorphization.  That does solve the inlining problem, but shows a little bit of measurable overhead.

nikomatsakis:
@cuviper can you push a WIP commit, or send me a branch?

cuviper:
I pushed two WIP branches, [thief-splitter-assoc](https://github.com/cuviper/rayon/commits/thief-splitter-assoc) (with FIXMEs) and [thief-splitter-enum](https://github.com/cuviper/rayon/commits/thief-splitter-enum).

Actually, if I set `next_velocity_par` with explicit `weight(1.0)`, then even the enum version is basically unnoticeable to performance.  I think before I was still comparing master's full-cost implementation to the new scenario with an outer weight using cost but inside using defaults (now thief), and that's not apples-to-apples for evaluating the new CostSplitter overhead.

So I think the enum version may be Good Enough.  The associated-type approach feels nicer if we can make it work, to let the splitter monomorphize, but it appears not to matter much.

nikomatsakis:
@cuviper 

> Do you have any type wizardry for joining/resolving two associated types? I'm having trouble making chain and zip work, as well as combining the producer to the consumer. If either of the two has CostSplitter, I want that to win, otherwise the combination should use ThiefSplitter.

Hmm. So, it would be easy with specialization -- but without that, we can resort to a trick that relies on LLVM. In particular, LLVM is good at constant propagation. So I did this trick on [my thief-splitter-assoc branch](https://github.com/nikomatsakis/rayon/tree/thief-splitter-assoc) where you add a `fn use_cost() -> bool` to the type. See [this commit](https://github.com/nikomatsakis/rayon/commit/79cdbd858642dd19725deb1352884dfc195a07c9) in particular.

> So I think the enum version may be Good Enough. The associated-type approach feels nicer if we can make it work, to let the splitter monomorphize, but it appears not to matter much.

Interestingly, I did a bunch of measurements, and I found that the enum branch was actually **the fastest**, at least locally. I can't really explain this and it warrants further study. =)

https://gist.github.com/nikomatsakis/396de1975f5210aadeea414d2eb6deff

I still don't know why the `par_reduce` case (which corresponds to a weight on the outer loop, adaptive on the inner) performs so poorly. I tried experimenting with 2x the number of splits on the thief (that's a clever thing btw; I didn't have a max on the number of splits in adaptive mode, which is probably why your branch is performing better than mine) but that made things worse. =) Using 1/2 the number of splits basically gets the same results.

It's not that big a deal, esp since the unweighted version gets pretty close (18mil-20mil ns) to the fully weighted version (15mil ns), at least with the enum version.

nikomatsakis:
Regarding the `par_reduce` case: if I reduce the weight to 100, things get worse; if I up it to `weight_max`, things get better (i.e., it performs the same as unweighted), so clearly that suggests that the outer loop is splitting too few times. It occurs to me that this might be a factor of how many CPUs you have. IOW maybe the problem is that weight 200 is just poorly tuned for the 16 active threads I have on my linux machine (lunch-box).

nikomatsakis:
Anyway, my conclusion is that I am happy to merge the enum version: it performs the best and the code is simplest. But I wish I understood why the assoc type variant is slower!

nikomatsakis:
@cuviper ok so I thought a bit more about this. I think we should adopt the enum-based splitter. In fact, I might be inclined to just remove the cost variant and only use the adaptive mode, since I think that the weight API worries me (it's kind of arbitrary). 

**But** there is still some work to be done. It seems to me that the current adaptive policy will (potentially) perform poorly if you have a cluster of really heavy tasks at the start of an array, and then a bunch of cheap ones. It will split N times offer up the tails of the array to be stolen, but those tasks are all cheap. Meanwhile it ends up with a bunch of heavyweight tasks to run through sequentially.

We might mitigate this by oversubscribing the number of splits (i.e., num threads * 2), but that costs in some benchmarks (e.g., `body_par_reduce` suffers). We can mitigate *that* probably by optimizing and reducing overhead. 

But it may also be useful to have a very simple knob with a few choice.

cuviper:
OK, I pushed the enum-based splitter.

With your par_reduce weight experiments, also try leaving the outside loop at 200, and add a `weight(1.0)` on the inner loop as I mentioned.  That seems to recover performance for me.  Or if we just ditch weights, this won't matter...

For tunable knobs, I think we could expose an oversubscribe value for the user to affect the initial number of splits.  Perhaps that could also affect the stolen re-splits, but that makes me nervous.  This could be a new adaptor, or we could try to reuse `Weight` -- but it's still fuzzy how to apply that.  e.g. Does `weight(2.0)` mean "split into twice the number of cpus"?  Or should we use `THRESHOLD/weight` or `THRESHOLD/weight * ncpus` to more closely approximate the weight tuning we already had, for compatibility?

nikomatsakis:
Merged. Nice work @cuviper! ⛷

cuviper:
Thanks!

PR #116, opened 2016-10-20
mergesort is correct on duplicate values

Eh2406:
This is @edre's fix in https://github.com/nikomatsakis/rayon/issues/105


nikomatsakis:
@edre nice! do you think you could also change the mergesort to use a randomly generated initial vector (with a deterministic key), more like quicksort?

Eh2406:
like this?

nikomatsakis:
looks good =)

PR #118, opened 2016-10-22
Use a weaker rng and reduce lock contention when possible.

emilio:
I had originally more improvements to `steal_work` based on [Servo's workqueue](https://github.com/servo/servo/blob/master/components/style/workqueue.rs), but they were pretty specialized and after the `scope` change those paths no longer appear in the profile as much as far as I know. I still want to profile it more, but wanted to get this reviewed.

The way to get a `Registry` from a `ThreadPool` is completely nasty, but otherwise I needed to expose it, which is undesirable. Any better solution is appreciated.

### DIFF
--- a/src/thread_pool.rs
+++ b/src/thread_pool.rs
@@ -5,7 +5,7 @@ use job::{JobRef, JobMode};
 use latch::{Latch, LockLatch, SpinLatch};
 #[allow(unused_imports)]
 use log::Event::*;
-use rand;
+use rand::{self, Rng};
 use std::cell::Cell;
 use std::sync::{Arc, Condvar, Mutex, Once, ONCE_INIT};
 use std::thread;
@@ -217,6 +217,9 @@ pub struct WorkerThread {
     /// on the current thread; this is used by the scope code to
     /// ensure that the depth of the local deque is maintained
     spawn_count: Cell<usize>,
+
+    /// A weak random number generator.
+    rng: rand::XorShiftRng,
 }
 
 // This is a bit sketchy, but basically: the WorkerThread is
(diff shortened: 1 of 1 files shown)


src/thread_pool.rs
```diff
@@ -112,23 +112,27 @@ impl Registry {
 
     fn start_working(&self, index: usize) {
         log!(StartWorking { index: index });
-        let mut state = self.state.lock().unwrap();
-        state.threads_at_work += 1;
+        {
+            let mut state = self.state.lock().unwrap();
+            state.threads_at_work += 1;
+        }
         self.work_available.notify_all();
```
nikomatsakis: Huh, I wonder why I had it in my head that one should hold the lock when notifying. It doesn't make any sense, of course. Seems good.

src/scope/mod.rs
```diff
@@ -231,13 +263,16 @@ impl<'scope> Scope<'scope> {
...
             assert!(old_value > 0); // scope can't have completed yet
+
             let job_ref = Box::new(HeapJob::new(self, body)).as_job_ref();
             let worker_thread = WorkerThread::current();
             if !worker_thread.is_null() {
                 let worker_thread = &*worker_thread;
                 let spawn_count = worker_thread.spawn_count();
                 spawn_count.set(spawn_count.get() + 1);
                 worker_thread.push(job_ref);
+            } else if let Some(thread_pool) = self.thread_pool {
```
nikomatsakis: So I'm a bit confused here. I think you could do `thread_pool.install(|| scope(|s| ...))` and you would be all set, no? 

The idea is that, when you call `install`, you are in the worker thread for this thread-pool, so then when we come into `scope`, we are already there.

emilio: I think we can't do that because `install` waits until the closure is completely executed, right? So the semantics of this would change significantly.

emilio:
I dropped that commit :)

PR #120, opened 2016-10-25
[RFC] [breaking-change] Simplify `reduce` and `fold`

### LINKED ISSUE #94: settle on one reduce, one fold
Currently we have a lot of method for doing "fold-like" things:

``` rust
    fn reduce_with<OP>(self, op: OP) -> Option<Self::Item>
        where OP: Fn(Self::Item, Self::Item) -> Self::Item + Sync;
    fn reduce_with_identity<OP>(self, identity: Self::Item, op: OP) -> Self::Item
        where OP: Fn(Self::Item, Self::Item) -> Self::Item + Sync,
              Self::Item: Clone + Sync;
    fn reduce<REDUCE_OP>(self, reduce_op: &REDUCE_OP) -> Self::Item
        where REDUCE_OP: ReduceOp<Self::Item>;
```

And of course there is also the (experimental, but faster in some cases) `fold` operation:

``` rust
    fn fold<I,FOLD_OP,REDUCE_OP>(self,
                                 identity: I,
                                 fold_op: FOLD_OP,
                                 reduce_op: REDUCE_OP)
                                 -> I
        where FOLD_OP: Fn(I, Self::Item) -> I + Sync,
              REDUCE_OP: Fn(I, I) -> I + Sync,
              I: Clone + Sync + Send;
```

Part of the challenge here is that parallel reduce is different from fold: you can't just start from the left and combine elements until the end. Instead, you wind up dividing the list into a kind of tree, folding each of those segments from left-to-right, and then reducing the results as you walk up the tree. 

Each of the variations we currently have makes sense in its own way:
- `reduce_with` is simplest, but it's the most inefficient. Because it requires two values from the iterator before we can invoke the closure, it has to introduce options under the hood to track things. Maybe there is a nicer way to implement it?
- `reduce_with_identity` avoids the problem of introducing `Option` by requiring an initial element to use as the "accumulator" value. However, unlike in sequential fold, this value must be _cloneable_, which is unfortunate sometimes. (It has to be cloneable because we will need to use it for each of the parallel segments we wind up with.)
- The original `reduce` was intended for letting people write custom operations that configure each detail. Its pretty silly for this to have the best name, but it might make sense to keep it around.
- The `fold` operation exposes this "tree-and-then-walk-back-up" structure directly. I added it to try and make a benchmark faster. It worked, but it's again somewhat complicated. I think I've decided that it _does_ make sense to keep it around, but I'm not 100% sure. 

My feeling is that we should probably have a variation on `reduce_with_identity` as the **only reduce** operation. It would instead have a closure that produces the identity, so that `sum` looks like this:

``` rust
iterator.reduce(|| 0, |a, b| a+b)
```

At that point, there is no need for the current `reduce`, since the `Reduce` trait just provides two methods that are basically equivalent to these two closures. Moreover, we can lift the `Clone` bounds.

If you need to artifically inject an identity value, as the current `reduce_with` does, it can be expressed like so:

``` rust
iterator.reduce_with(|| None, |a, b| match (a, b) {
    (Some(a), Some(b)) => a+b,
    (s @ Some(_), None) | (None, s @ Some(_)) => s,
    (None, None) => None,
})
```

But I think nobody ever will.

So that just leaves the question of whether to keep `fold` around. My feeling now is that we should, but modify it to take a closure. I can easily imagine cases where it's useful to be able to retain state across the sequential folds, which `fold` makes possible. `fold` also basically corresponds to the `map-reduce` problem -- you can express this as `.map().reduce()` but sometimes it more efficient to convert N items into the new type than to convert each item individually, which is basically the point.

As an example, the nbody demo uses fold like this:

``` rust
iterator.fold(
    || (0, 0), // represents the total "difference" in current position of this item
    |(diff1, diff2), item| ..., // adjust diff1, diff2 based on item
    |(diff1, diff2), (diff1b, diff2b)| (diff1 + diff1b, diff2 + diff2b));
```

Sound good?

cc @jorendorff, @cuviper  -- with whom I talked about this

nikomatsakis:
We currently have too many variants of `reduce`. This brings us down to two:
- `reduce(identity, combine)`, which will combine elements in your iterator by calling `combine(l, r)`, and using `identity()` to produce an identity value when it needs to
  - by taking a closure, we can remove the `Clone` bound on `reduce`
  - by taking an identity, we can make the return value `T` and be maximally efficient
- `reduce_with(combine)`, which will combine elements in your iterator by calling `combine(l, r)`
  - it does not take an identity, so it returns `None` if your iterator is empty

I've also stabilized and modified `fold` -- it might make sense to give it another name, though, like `partial_fold`. The new `fold` is kind of like sequential fold except that it operates only on subsequences (determined dynamically), so it produces a series of values instead of folding the entire sequence into one value. It is cool because the value you are producing does not have to be the same type as your inputs (unlike `reduce`), so you can do things like this:

``` rust
use rayon::prelude::*;
let s =
    ['a', 'b', 'c', 'd', 'e']
    .par_iter()
    .fold(|| String::new(), |mut s, &c| { s.push(*c); s })
    .reduce_with(|mut a: String, b: String| { a.push_str(&b); a });
```

Here, the `fold` is pushing characters onto strings, and the reduce then combines the resulting strings. This should produce some number of strings roughly proportional to the number of CPUs you have, instead of being proportional to the size of your input (as it would be with just `reduce`).

Fixes #94 

cc @jorendorff
cc @cuviper

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,12 @@
+# Release 0.5 (pending)
+
+- **Breaking change:** The `reduce` method has been vastly
+  simplified, and `reduce_with_identity` has been deprecated.
+- **Breaking change:** The `fold` method has been changed. It used to
+  always reduce the values, but now instead it is a combinator that
+  returns a parallel iterator which can itself be reduced. See the
+  docs for more information.
+
 # Release 0.4.3
 
 - Parallel iterators now offer an adaptive weight scheme,
--- a/rayon-demo/src/nbody/nbody.rs
+++ b/rayon-demo/src/nbody/nbody.rs
@@ -342,12 +342,10 @@ fn next_velocity_par(time: usize, prev: &Body, bodies: &[Body])
         acc += diff;
     }
 
-    let zero: Vector3<f64> = Vector3::zero();
     let (diff, diff2) = bodies
         .par_iter()
-        .fold(
-            (zero, zero),
-            |(mut diff, mut diff2), body| {
+        .fold(|| (Vector3::zero(), Vector3::zero()),
+              |(mut diff, mut diff2), body| {
                 let r = body.position - prev.position;
 
                 // make sure we are not testing the particle against its own position
@@ -395,8 +393,9 @@ fn next_velocity_par(time: usize, prev: &Body, bodies: &[Body])
                 }
 
                 (diff, diff2)
-            },
-            |(diffa, diff2a), (diffb, diff2b)| (diffa + diffb, diff2a + diff2b));
+              })
+        .reduce(|| (Vector3::zero(), Vector3::zero()),
+                |(diffa, diff2a), (diffb, diff2b)| (diffa + diffb, diff2a + diff2b));
 
     acc += diff;
     acc2 += diff2;
--- a/src/par_iter/fold.rs
+++ b/src/par_iter/fold.rs
@@ -1,94 +1,144 @@
-use super::ParallelIterator;
+use super::*;
 use super::len::*;
 use super::internal::*;
 
-pub fn fold<PAR_ITER,I,FOLD_OP,REDUCE_OP>(pi: PAR_ITER,
-                                          identity: &I,
-                                          fold_op: &FOLD_OP)
-                                          -> I
-    where PAR_ITER: ParallelIterator,
-          FOLD_OP: Fn(I, PAR_ITER::Item) -> I + Sync,
-          I: Clone + Send + Sync,
+pub fn fold<U, BASE, IDENTITY, FOLD_OP>(base: BASE,
+                                        identity: IDENTITY,
+                                        fold_op: FOLD_OP)
+                                        -> Fold<BASE, IDENTITY, FOLD_OP>
+    where BASE: ParallelIterator,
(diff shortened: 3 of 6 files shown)


nikomatsakis:
I thought of one twist on fold that is kind of interesting -- though it probably deserves another method. Given that we are iterating over values of take T, it could produce an iterator over values of type V, and take three closures:

- `identity() -> U` produces fresh intermediate value of type `U`
- `fold(U, T) -> U` adds a T into a U
- `produce(U) -> V` converts the final U into a V

On the one hand, this final `produce` step is just a `map`. On the other, the intermediate state (`U`) need not be `Send` in this setup, I think.

edre:
As a use case in support of the intermediate map between fold and reduce, I could implement merge sort with something like:

x.par_iter().fold(Vec::push).map(Vec::sort).reduce(merge)

nikomatsakis:
So it's been a while since I opened this PR. I'm still feeling really good about it. 

Any *objections* to me merging it? :) Speak in a day or two or forever hold your peace.

nikomatsakis:
OK, it's been a day or two and I want `fold` so I can do some unrelated experiments. Landing it. =)

PR #122, opened 2016-10-25
Steal while scope blocks

### LINKED ISSUE #108: scope should shift to helper thread, and then "help" when blocking
Currently, when a `scope` is created, it does not shift into a worker thread. This complicates the logic mildly. In particular, then when the scope blocks waiting for its children, we don't know if that code is on a worker thread or not. Right now it just blocks -- but if it is on a worker thread, it should *help*.

My opinion is that we should (a) shift to a worker thread when a scope is created to execute the body of the scope and (b) help out when blocking.

nikomatsakis:
This branch causes us to shift to a worker thread as soon as you call `scope()`; individual calls to `scope.spawn()` are thus forced to execute in a worker thread, I believe. Since the scope is executing in a worker thread, we can then change to use a `SpinLatch` for when the workers are done, and then use `steal_until()` to steal until that completes.

I'm quite happy with these changes, but unhappy with not having tests. We need some bigger scope benchmarks. Something modeled on servo would be ideal (i.e., descending a tree, doing work (and maybe creating scopes too?) at each step).

Fixes #108

cc @emilio

### DIFF
--- a/neg-tests-compile/scope_join_bad.rs
+++ b/neg-tests-compile/scope_join_bad.rs
@@ -1,0 +1,22 @@
+extern crate rayon;
+
+fn bad_scope<F>(f: F)
+    where F: FnOnce(&i32) + Send,
+{
+    rayon::scope(|s| {
+        let x = 22;
+        s.spawn(|_| f(&x)); //~ ERROR `x` does not live long enough
+    });
+}
+
+fn good_scope<F>(f: F)
+    where F: FnOnce(&i32) + Send,
+{
+    let x = 22;
+    rayon::scope(|s| {
+        s.spawn(|_| f(&x));
+    });
+}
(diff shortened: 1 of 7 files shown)


nikomatsakis:
I think I thought of a bug -- but you will only see it if you intermingle join/scope in just the right way. The `scope()` call should ensure that the thread-local deque remains balanced (like `spawn` does), but it's not doing that right now. We can probably factor this out into a `ScopeJob` (and then move `HeapJob`  into `job`, so that it just manages the heap part of the equation). This would make me happy, since `HeapJob` seems more generic than `scope`.

nikomatsakis:
I'm gonna' merge this. It's mostly touching the unstable scope API and seems like a clear improvement.

Issue #123, opened 2016-10-25
we need a convenient `collect()`, `collect_into()` should support other collection types


nikomatsakis:
In general, it'd be nice if `collect_into()` were more flexible and convenient. We may need multiple APIs in the end. Some things I would like:
- a convenient `collect()` that allocates a new collection, as with sequential iterators;
- maybe rename `collect_into()` into `extend()` and have it work like `extend()`?
  - this is possible for vectors, at least
- support for some more collection types:
  - for exact length iterators, we could support slices (not vectors)
    - but we'd have to assert that their length is correct
  - concurrent collections could be easily supported
  - also persistent collections like https://github.com/michaelwoerister/hamt-rs
  - maybe worth supporting something like `RwLock<&mut HashMap>`
    - could be practical in some scenarios...
    - or maybe collecting up pairs in a vec and then (transparently) storing them into the hashmap after, as a seq step?
- support for more parallel iterator types:
  - vecs can support bounded iterators, but they need a "compact" step
    - have to be careful about panics, though we can leak if needed

Well, ok, that's not very organized. But I feel like we can put a bit more effort into making this ergonomic and "bridging" between various things. I'm torn. =)

I suspect we'll wind up with more than just `collect_into`, in any case, though specialization might help here too.

nikomatsakis:
Marking as 1.0 -- **some** parts of this, e.g. `collect` and at least a trait-ified `collect_into`, we should have. The rest maybe not.

jwilm:
I'm interested in a `collect_into()` that works on a bounded iterator. Hoping to be able to achieve something like:

```rust
items.par_iter()
     .filter_map(func)
     .collect_into(&mut existing_vec);
```

What is the compaction step you're talking about? It seems like it should be possible to tightly pack the vec being collected into using an atomic counter for the next index to write into.

nikomatsakis:
@jwilm I'm going to open a separate issue for this. I'll spell out how I think it should work. An atomic counter is not needed.

nikomatsakis:
(Closing because we now have a `collect()` API)

PR #124, opened 2016-10-25
add a `collect()` method to parallel iterators

nikomatsakis:
Implement it for `Vec<T>` and `HashMap<K, V>`. The latter one is mildly
dubious, though I think it's functionality we will want to support (and
probably optimize over time). I included it in part to force type
inference hints.

cc #123

### DIFF
--- a/rayon-demo/src/main.rs
+++ b/rayon-demo/src/main.rs
@@ -1,4 +1,5 @@
 #![cfg_attr(test, feature(test))]
+#![cfg_attr(test, feature(conservative_impl_trait))]
 
 use std::env;
 use std::io;
@@ -12,6 +13,7 @@ mod sieve;
 
 // these are not "full-fledged" benchmarks yet,
 // they only run with cargo bench
+#[cfg(test)] mod map_collect;
 #[cfg(test)] mod factorial;
 #[cfg(test)] mod pythagoras;
 #[cfg(test)] mod fibonacci;
--- a/rayon-demo/src/map_collect.rs
+++ b/rayon-demo/src/map_collect.rs
@@ -65,9 +65,14 @@ mod util {
                     map
                 })
           .reduce(|| HashMap::new(),
-                  |mut map1, map2| {
-                      map1.extend(map2);
-                      map1
+                  |mut map1, mut map2| {
+                      if map1.len() > map2.len() {
+                          map1.extend(map2);
+                          map1
+                      } else {
+                          map2.extend(map1);
+                          map2
+                      }
                   })
     }
 }
--- a/rayon-demo/src/map_collect.rs
+++ b/rayon-demo/src/map_collect.rs
@@ -1,0 +1,168 @@
+//! Some benchmarks stress-testing various ways to build the standard
+//! `HashMap` data structures from the standard library.
+
+mod util {
+    use rayon::prelude::*;
+    use std::collections::{LinkedList, HashMap};
+    use std::hash::Hash;
+    use std::sync::Mutex;
+
+    /// Do whatever `collect` does by default.
+    pub fn collect<K, V, PI>(pi: PI) -> HashMap<K, V>
+        where K: Send + Hash + Eq,
+              V: Send,
+              PI: ParallelIterator<Item = (K, V)> + Send
+    {
+        pi.collect()
+    }
+
+    /// Use a system mutex.
+    pub fn mutex<K, V, PI>(pi: PI) -> HashMap<K, V>
+        where K: Send + Hash + Eq,
+              V: Send,
+              PI: ParallelIterator<Item = (K, V)> + Send
+    {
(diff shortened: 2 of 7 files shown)


edre:
I think LinkedList is a logical default implementation because it has constant time concatenation.

cuviper:
Looks fine to me.  I think we should go ahead an add all of the std collections, so it mirrors `IntoParallelIterator`.  Although now that I write that, you probably should spell out `FromParallelIterator` for consistency.

edre:
To clarify my suggestion, we could implement an efficient collect() for non Exact iterators by collecting into linked lists.

cuviper:
@edre that's interesting, yes we could do that.  Although collecting to `Vec` should be much faster when possible, which means we'd really want specialization here. (`LinkedList` in general, `Vec` for exact)

nikomatsakis:
So the actual thing I was planning on doing -- and maybe I should do it in this PR? -- was to develop a `for_each_mut` (not a good name, just don't know of a better one just yet -- maybe just call it `scatter`). This would basically run a given action for each item in the iterator in *some* order. Unlike `for_each`, though, it would allow mutable access to the environment, because it would ensure that each item is processed sequentially. This would be good for collecting into a `HashMap` or `HashSet`, in particular. You could also use it for gathering into a vector, but only if you don't care about ordering. =)

The implementation technique I had intended to use for `for_each_mut` was to do some variant on [flat combining](https://www.cs.bgu.ac.il/~hendlerd/papers/flat-combining.pdf), though I am not sure if there's some cool way to take advantage of the fork-join structure here to reduce overheads?

nikomatsakis:
> I think LinkedList is a logical default implementation because it has constant time concatenation.

One thought I had for a good general-purpose intermediate data structure was to use a `fold` (from #120) to group things into `Vec`, and then a `reduce` to arrange those vectors into a little tree. This is basically building up a rope.

nikomatsakis:
Just for fun, I impl'd the vec-tree approach (basically untested) in https://github.com/nikomatsakis/rayon/pull/130, so you can see what I'm talking about.

nikomatsakis:
OK, I updated this PR in two ways:

- It supports `collect()` into a linked list, as @edre suggested; this seems definitely useful.
- It uses a `LinkedList` as the intermediate structure for building a map, so that this can support very unrestricted bounds.

I'm not sure whether to remove support for building a map. It's convenient but the implementation isn't that great. The main reason I included it was to have two implementations to start with, and `LinkedList` serves *that* role better.

cuviper:
Any thought on renaming the trait to `FromParallelIterator`?  `FromParIter` is rather incongruous.

edre:
I've been thinking about efficient ways to use parallel iterators with HashMap.
IntoParIter: Use a BoundedParIter on the backing table, returning elements for filled cells.
FromParIter: Pre-hash keys, partition them into N buckets, reduce the buckets into N LinkedLists, insert elements from a bucket into their section of the table in parallel.

These involve mucking around with the internal implementation. Is there any way to access these without just forking HashMap?

nikomatsakis:
@cuviper done

cuviper:
Travis failed because you didn't update the prelude.

nikomatsakis:
I updated the prelude.

Also, I prototyped this "flat combine" stuff here this morning:

https://github.com/nikomatsakis/rayon/tree/flat-combine

It seems to work, though not thoroughly tested by any means. Also a lot of unsafe code that can probably be better factored.

Still: really cool!

nikomatsakis:
@edre 

> These involve mucking around with the internal implementation. Is there any way to access these without just forking HashMap?

I don't know of a way to do that, no. I think that another major gap I would like to close is to have an efficient way to go from a seq iter into par execution. For example, using `scope`, one could certainly just create tasks for each item, something like this:

```rust
scope(|s| {
    for item in seq_iter { s.spawn(move || process(item)); }
});
```

This is not the most efficient thing to do, however; I think doing better would require some offering some alternative iterator interfaces in the stdlib, closer to the Rayon `Producer` traits. Or else reimplementing those data structures of course.

I think that balanced-tree-based rope-like data structures are also really appealing here, as I've noted elsewhere. For example, it'd probably be easy to make [hamt-rs](https://crates.io/crates/hamt-rs) act as a Rayon producer (and it's also easy to construct one, since you can just use straight-up `reduce`, though I do wonder if that would *actually* be faster than the `for_each_atomic` approach; unlikely, I think).

nikomatsakis:
OK, updated to support hashmap/hashset/btreemap using a linked list intermediary (at least for now). Also tweaked documentation to promise deterministic results. 

I think I'd like to merge this. Thoughts? Objections?

nikomatsakis:
> I think I'd like to merge this. Thoughts? Objections?

Hearing none, merging.

PR #126, opened 2016-10-27
fix(thread_local): Removes unnecessary semicolon

durango:
Not sure if this is explicitly a rust `1.9.0` thing or not

### DIFF
--- a/src/par_iter/internal.rs
+++ b/src/par_iter/internal.rs
@@ -93,7 +93,7 @@ impl Splitter {
     fn thief_id() -> usize {
         // The actual `ID` value is irrelevant.  We're just using its TLS
         // address as a unique thread key, faster than a real thread-id call.
-        thread_local!{ static ID: bool = false; }
+        thread_local!{ static ID: bool = false }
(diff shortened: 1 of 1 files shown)


cuviper:
Oh, interesting, the semicolon was only added to the `thread_local!` pattern in 1.11.0.  It's completely optional though, so your fix is fine.

Maybe add `1.9.0` to `.travis.yml`?  That way we always know the required minimum, and make sure that it works.  This is the true minimum right now too, for `std::panic` support.  I don't know if @nikomatsakis has a policy on rustc versions being a breaking change -- opinions vary and I'm in the camp that it is breaking.  Either way, it's good to have that in CI so we know what works right now.

nikomatsakis:
@cuviper I'm not sure if I've formed a strong opinion, but I do expect users to basically keep up with stable --- that is to say, if there is a new stable feature, I wouldn't hesitate to add a dependency on it. But it seems reasonable to consider that a breaking change, at the same time.

In any case, I always forget that `;` isn't allowed in `thread_local!`.

PR #128, opened 2016-10-28
Implement ability to find something that matches

nikomatsakis:
This will find any matching thing -- but it's not deterministic which one it finds. I...think it's correct. =) I'd like to think about better ways to test.


cuviper:
Doh, I was just working on something similar last night:
https://gist.github.com/cuviper/401b55afd08993414f554e071d972cec

Mine has no unsafe code, uses relaxed atomic ops, and is a little more aggressive about terminating the folder loop too.

Should I open a competing PR for comparison? Or should we just work on yours?

nikomatsakis:
Let's compare them! In retrospect it would make sense to move the "should continue" check into the folder loop too, probably. 
But also yeah I was debating about changing the ops to acquire/release instead of seq CST. I am just lazy. :)
To replace the unsafe, did you use a mutex?
On Oct 28, 2016 3:56 PM, Josh Stone <notifications@github.com> wrote:Doh, I was just working on something similar last night:
https://gist.github.com/cuviper/401b55afd08993414f554e071d972cec

Mine has no unsafe code, uses relaxed atomic ops, and is a little more aggressive about terminating the folder loop too.

Should I open a competing PR for comparison? Or should we just work on yours?

cuviper:
Mine goes all the way to `Ordering::Relaxed`, because I don't use it for real synchronization.  It only enables the early-exit optimization, and in theory a relaxed flag should require much less bus-locking/snooping.  The actual values are just reduced as normal, arbitrarily picking one if it happened to find several before exiting.

I'll polish mine up into a PR tonight for us to compare.  I've got some simple benchmarks too, which we can try with yours to see how it actually fares.

src/par_iter/mod.rs
```diff
@@ -26,11 +26,13 @@ use self::internal::*;
 use self::weight::Weight;
 use self::zip::ZipIter;
 
+mod abort;
```
cuviper: Is this unrelated work?  The module is missing.

nikomatsakis:
Closing in favor of #129

PR #130, opened 2016-11-02
[WIP] Collect based on vec tree

nikomatsakis:
Just for fun, this is a variant on #124 that uses a `VecTree` (a lame-o data structure I just whipped up) as its intermediate storage. It also builds on #120. =) This allows us to build maps from non-exact parallel iterators.

Definitely need to make some benchmarks to test the performance of this as compared to a vector. Specialization would probably let us have the best of both worlds here once that stabilizes.

And then I'd like to implement flat-combining and compare the performance of both to that. =)

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,12 @@
+# Release 0.5 (pending)
+
+- **Breaking change:** The `reduce` method has been vastly
+  simplified, and `reduce_with_identity` has been deprecated.
+- **Breaking change:** The `fold` method has been changed. It used to
+  always reduce the values, but now instead it is a combinator that
+  returns a parallel iterator which can itself be reduced. See the
+  docs for more information.
+
 # Release 0.4.3
 
 - Parallel iterators now offer an adaptive weight scheme,
--- a/rayon-demo/src/nbody/nbody.rs
+++ b/rayon-demo/src/nbody/nbody.rs
@@ -342,12 +342,10 @@ fn next_velocity_par(time: usize, prev: &Body, bodies: &[Body])
         acc += diff;
     }
 
-    let zero: Vector3<f64> = Vector3::zero();
     let (diff, diff2) = bodies
         .par_iter()
-        .fold(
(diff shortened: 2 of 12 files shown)


edre:
While I think a LinkedList<Vec<T>> is sufficient for this use case, this VecTree could be used as both an efficient producer and an efficient consumer. I don't think there's anything currently in the API that could benefit from this though.

nikomatsakis:
Here are the advantages I see:

- Using a vector at the leaves may be more efficient, if the leaves are sufficiently large
    - this will depend on the efficacy of `fold` in creating good groupings though
- As @edre said, given a vec-tree, it's easy to produce again
    - one possible use would be to allow creating a `Vec` from a non-exact par iter
    - basically build into a vec-tree and then store back out

The last point is one of the reasons I've been wanting to experiment with `VecTree`. I was somewhat inspired by papers like [Lazy Tree Splitting](http://manticore.cs.uchicago.edu/papers/jfp-lts-submitted.pdf) and [Data Only Flattening](http://manticore.cs.uchicago.edu/papers/ppopp13-flat.pdf) that discuss the use of trees of this kind. 

That said, it'd be worth trying and comparing against a linked list. It's true that they are a good fit too.

cuviper:
I'm suggesting to use vectors in the linked list nodes, so you'd still get your grouping.

I see how using a tree makes it possible to fork again, e.g. as a new `into_par_iter`, but I'm not clear if that's actually what you're saying.

nikomatsakis:
Ah, I hadn't considered a `LinkedList<Vec>`. But yes I was saying that using a tree lets us do a new `into_par_iter`. In any case, upon further consideration I decided to update the original PR to support a `LinkedList` -- I'd like to experiment with stuff like `VecTree`, but I'd prefer to land `collect()` ASAP, and supporting a data structure like `LinkedList` makes sense anyhow.

nikomatsakis:
Closing this for now. Interesting thought experiment.

PR #131, opened 2016-11-03
[WIP] Use flat combining to implement `for_each_atomic`

nikomatsakis:
This is a branch based off of #124. It adds a new method called `for_each_atomic()` that executes a closure for each item in the parallel iterator; unlike normal `for_each`, those executes are guaranteed for execute atomically with respect to one another. Under the hood, it uses "flat combining" to manage the synchronization. 

This PR is marked as WIP because it is NOT thoroughly tested by any means. In fact, I just hacked it up this morning. I wanted to throw it out there so people can start taking a look. Some known problems:

- [ ] need stress tests
- [ ] need tests of panic behavior
- [ ] need benchmarks
- [ ] ordering requirements are surely stronger than necessary
- [ ] consider if we can refactor the interface out somewhat
    - do we want to expose the `FlatCombiner` type more generally?
- [ ] consider a "pipelined" flow
    - each time we want to execute the closure, the current code will block until it has executed
    - another option in our case would be to enqueue that we want to execute, and only block if, by the time we have a new item to execute, it still hasn't been executed
    - this would enable more pipeline parallelism but would probably also be less fair to the owner
    - also mildly harder to manage the memory perhaps

cc @aturon


edre:
I skimmed through the paper, and I don't this will necessarily be faster for `collect()` than the linked-list technique.

The main benefits of FC seem to be batched operations in a single thread, and ability to use better combining algorithms within a batch. Rayon can already use better combining algorithms with fold and reduce, and this isn't available with `for_each_atomic` anyway. The linked list approach uses much less synchronization to batch 100% of the operations into a single thread's execution. I haven't been able to think of any usecase of `for_each_atomic` where the same logic doesn't apply.

Of course I could be missing something, and benchmarks would be a powerful argument ;)

nikomatsakis:
@edre you may be right! definitely want benchmarks. I was going to say that it depends how much work the parallel-izer is doing, but I guess that the more work it does, then the less it matters to allocate a `LinkedList`. (Regardless, it's a useful combinator.)

nikomatsakis:
I wonder if it'd be better to expose the core primitive; in particular, I am wondering if it could be a useful way to add things in the middle of a pipeline.

e.g., in sequential iterators, I often de-dup with something like:

```rust
.filter(|x| hashset.insert(x))
```

I could imagine wanting to do that in the midst of a parallel iterator, and this might be a nice way to enable it.

nikomatsakis:
I'm envisioning something like

```rust
let mut set: HashSet();
let mut combiner: FlatCombiner::new();
...
foo.par_iter()
    ...
    .filter(|x| combiner.locked(|| set.insert(x))
...
```

Obviously we'd have to support the more full-featured version in the paper, that permits threads to dynamically "subscribe" and "unsubscribe" and which isn't really tied to Rayon's threadpool.

(It'd probably be a better fix for crossbeam)

edre:
The core primative seems to just have the API of `Mutex`, with an optional batch processing mode. You could also do your deduping example with a regular mutex.

Or we could provide `for_each_atomic`, and implement it with a linked list ;)

nikomatsakis:
Actually, it is not the same as mutex, and my example code was wrong. It would look more like this:

```rust
let mut set: HashSet();
let mut combiner: FlatCombiner::new(|x| set.insert(x));
...
foo.par_iter()
    ...
    .filter(|x| combiner.execute(x))
...
```

In particular, with a normal mutex, if thread X locks it, then thread X does stuff. With this API, thread X submits arguments, but the code which executes the function may be some other thread Y (the "owner" thread in my code). That is, conceptually, instead of "locking things", you "submit requests" to be done (and it may be you that serves them).

It's more like an RPC model where there is some other thread taking and responding to requests, except that the work of executing this thread is spread amonst the users.

edre:
At least in the paper, the thread blocks until some thread (maybe itself) finishes its task, so the semantics are still those of a mutex. This may be worked around for collect, when the collecting could be done asynchronously, but not for dedup.

nikomatsakis:
@edre blocking is not the issue; consider the current mutex api:

```rust
let data = mutex.lock();
data.insert(...); // <-- this will always be executed by the current thread
```

Under the combining model, the `data.insert()` is maybe executed by another thread as part of a batch processing. 

Certainly the measurements in the paper (which obviously I've not replicated) suggest that it scales quite well, unlike a standard mutex.

src/par_iter/for_each_atomic/flat_combine.rs
```diff
@@ -0,0 +1,230 @@
...
+                debug_assert!(sender.is_empty());
+                mem::forget(data_guard);
+                return;
+            }
+
+            if sender.is_empty() {
+                // Our request was served, so we can just return, and
+                // `data_guard` is not needed.
+                mem::forget(data_guard);
+                return;
```
cuviper: I think you could get around this with separate reads and writes, instead of `try_take`'s atomic swap.  Read the pointer, and if it's not null work on it.  No one else will race here because there's only one global owner.  Null the pointer only after you're done to indicate "request was served".

nikomatsakis: ah, yes, thanks =) I remember thinking that when I was writing the code but forgetting to come back and I guess I forgot to come back and decide how to handle it.

edre:
Overall, this abstraction doesn't seem particularly well matched to ParallelIterator, and I agree that some crate like crossbeam may be a better home for it.

That would leave 3 very similar ways to implement dedup in rayon:
.filter(|x| mutex_map.lock().insert(x))
.filter(|x| some_crate::FlatCombiner::Set.insert(x))
.filter(|x| concurrent_map.insert(x))

Which one to use depends on the data, the cpu, and the available concurrency, like any other parallelism tradeoff.

nikomatsakis:
OK, I did two simple benchmarks building up hashmaps. I am working on some more. Here are the results from my 8x2-core linux machine and my 4-core Mac. The TL;DR is that 

- as @edre predicted, linked lists are the fastest overall
    - despite my enthusiasm for shiny new toys, this makes some measure of sense when I think about it objectively =) memory allocation is pretty cheap
- as the paper suggests, flat-combining is much faster than a plain mutex.
    - particularly on mac

This benchmark maps `i -> i`, so each iteration generates a fresh entry in the map:

```
# linux
test collect::i_to_i::with_flat_combine        ... bench:  91,253,493 ns/iter (+/- 1,308,250)
test collect::i_to_i::with_linked_list         ... bench:  33,338,705 ns/iter (+/- 18,442,845)
test collect::i_to_i::with_mutex               ... bench: 134,447,913 ns/iter (+/- 1,708,647)
# mac
test collect::i_to_i::with_flat_combine        ... bench:  86,198,339 ns/iter (+/- 2,684,043)
test collect::i_to_i::with_linked_list         ... bench:  29,584,219 ns/iter (+/- 8,942,621)
test collect::i_to_i::with_mutex               ... bench: 878,947,979 ns/iter (+/- 8,055,243)
```

This benchmark maps `i % 10 -> i`, so there are lots of conflicts:

```
# linux
test collect::i_mod_10_to_i::with_flat_combine ... bench:  53,504,857 ns/iter (+/- 465,608)
test collect::i_mod_10_to_i::with_linked_list  ... bench:  27,524,499 ns/iter (+/- 15,400,415)
test collect::i_mod_10_to_i::with_mutex        ... bench:  75,214,173 ns/iter (+/- 1,050,789)
# mac
test collect::i_mod_10_to_i::with_flat_combine ... bench:  36,325,405 ns/iter (+/- 1,518,105)
test collect::i_mod_10_to_i::with_linked_list  ... bench:  12,779,631 ns/iter (+/- 2,130,939)
test collect::i_mod_10_to_i::with_mutex        ... bench: 823,202,360 ns/iter (+/- 9,224,250)
```

I am toying with a histogram benchmark now that uses various ways to count words and so forth. We'll see how far I get this morning, I'm running out of time. =)

I think my current feeling is that:

- I still feel that `for_each_atomic` is a useful primitive
    - I also think Rayon is a reasonable home for it, though I agree that crossbeam might also be a good place
    - in particular, because it's connected to Rayon, we are able to simplify the implementation
    - if it does live in Rayon, I'd prefer to refactor it into part of the "primitive interface" that the threadpool provides, like `join` and `scope`
- **But** linked list is a better choice for using as a basis for `collect()`
    - I was thinking I should benchmark `VecTree` too just to see what comes out of it
    - in particular if you wanted to build a `Vec` from a parallel iterator, you could do a `VecTree` and then recollect, that sort of thing

The reason I prefer linked list (or vec tree, or some such thing) as a basis for collect is not just performance, but (more important) the deterministic results.

src/par_iter/for_each_atomic/flat_combine.rs
nikomatsakis: (see the bug fix below; I used a sentinel, which I think is needed to ensure that everything gets freed properly during panics, but maybe it can be simplified)

nikomatsakis:
To clarify, when I said:

> I still feel that for_each_atomic is a useful primitive 

I meant "flat-combining". I'd prefer to factor it out of the `ParallelIterator` trait into something more re-usable. I'll experiment with that.

Going to close this PR for now, I think, since I wouldn't want to land it in quite this form.

PR #135, opened 2016-11-06
Add more map_collect benchmark alternatives

cuviper:

PR #137, opened 2016-11-08
Additions for parallel collect

cuviper:
- Re-add the extra benchmarks from #135 
- Change the general collect to use `LinkedList<Vec<_>>`
- Add `BinaryHeap` and `VecDeque` for symmetry with `IntoParallelIterator` collections.

### DIFF
--- a/src/par_iter/from_par_iter.rs
+++ b/src/par_iter/from_par_iter.rs
@@ -3,6 +3,7 @@ use super::{ParallelIterator, ExactParallelIterator};
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::hash::{BuildHasher, Hash};
 use std::collections::LinkedList;
+use std::collections::{BinaryHeap, VecDeque};
 
 pub trait FromParallelIterator<PAR_ITER> {
     fn from_par_iter(par_iter: PAR_ITER) -> Self;
@@ -29,6 +30,27 @@ impl<PAR_ITER, T> FromParallelIterator<PAR_ITER> for Vec<T>
         let mut vec = vec![];
         par_iter.collect_into(&mut vec);
         vec
+    }
+}
+
+/// Collect items from a parallel iterator into a vecdeque.
+impl<PAR_ITER, T> FromParallelIterator<PAR_ITER> for VecDeque<T>
+    where Vec<T>: FromParallelIterator<PAR_ITER>,
+          T: Send,
+{
+    fn from_par_iter(par_iter: PAR_ITER) -> Self {
+        Vec::from_par_iter(par_iter).into()
+    }
+}
+
+/// Collect items from a parallel iterator into a binaryheap.
+/// The heap-ordering is calculated serially after all items are collected.
+impl<PAR_ITER, T> FromParallelIterator<PAR_ITER> for BinaryHeap<T>
+    where Vec<T>: FromParallelIterator<PAR_ITER>,
+          T: Ord + Send,
+{
+    fn from_par_iter(par_iter: PAR_ITER) -> Self {
+        Vec::from_par_iter(par_iter).into()
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/par_iter/from_par_iter.rs
```diff
@@ -22,6 +33,27 @@ impl<PAR_ITER, T> FromParallelIterator<PAR_ITER> for Vec<T>
...
+    where Vec<T>: FromParallelIterator<PAR_ITER>,
+          T: Send,
+{
+    fn from_par_iter(par_iter: PAR_ITER) -> Self {
+        Vec::from_par_iter(par_iter).into()
+    }
+}
+
+/// Collect items from a parallel iterator into a binaryheap.
+/// The heap-ordering is calculated serially after all items are collected.
```
nikomatsakis: Nit: I'd prefer not to *specify* that it is computed *serially*, though I think we probably should guarantee some amount of left-to-right-ness.

src/par_iter/from_par_iter.rs
```diff
@@ -46,8 +78,11 @@ impl<PAR_ITER, K, V, S> FromParallelIterator<PAR_ITER> for HashMap<K, V, S>
           S: BuildHasher + Default + Send,
 {
     fn from_par_iter(par_iter: PAR_ITER) -> Self {
-        let vec: LinkedList<_> = par_iter.collect();
-        vec.into_iter().collect()
+        let list = collect_list_vec(par_iter);
```
nikomatsakis: Nit: it's probably worth adding a comment referencing the benchmarks in `rayon-demo` showing different strategies.

src/par_iter/from_par_iter.rs
```diff
@@ -61,8 +96,10 @@ impl<PAR_ITER, K, V> FromParallelIterator<PAR_ITER> for BTreeMap<K, V>
           V: Send,
 {
     fn from_par_iter(par_iter: PAR_ITER) -> Self {
-        let vec: LinkedList<(K, V)> = par_iter.collect();
-        vec.into_iter().collect()
+        let list = collect_list_vec(par_iter);
```
nikomatsakis: Nit: it feels like we can derive a helper here; something like

```rust
fn combine<...>(par_iter: PAR_ITER, make_start: FnOnce(&LinkedList<Vec<T>>)) {
    let list = collect_list_vec(par_iter);
    let start = make_start(&list);
    list.into_iter().fold(start, |mut coll, vec| { coll.extend(vec); coll })
}
```

given suitable generics and bounds :)

src/par_iter/test.rs
```diff
@@ -656,6 +657,22 @@ pub fn par_iter_collect() {
...
+    let c: VecDeque<i32> = a.iter().cloned().collect();
+    assert_eq!(b, c);
+}
+
+#[test]
+pub fn par_iter_collect_binaryheap() {
+    let a: Vec<i32> = (0..1024).collect();
+    let b: BinaryHeap<i32> = a.par_iter().cloned().collect();
+    assert_eq!(b.peek(), Some(&1023));
+    assert_eq!(b.len(), 1024);
```
nikomatsakis: This is probably fine, but of course a loop that drains would be mildly better, no?

e.g.

```rust
for n in (0..1024).rev() {
    assert_eq!(b.pop(), Some(n));
    assert_eq!(b.len(), n);
}
```

nikomatsakis: (I guess same holds for maps :)

src/par_iter/from_par_iter.rs
nikomatsakis: Maybe just say "Performance note:" at the front?

We could also add an explicit note that the collection order is considered undefined for the time being.

That said, I am a big fan of deterministic results, so I feel like whatever strategy we use, I would prefer to keep an overall "left beats right".

PR #138, opened 2016-11-12
avoid linear stack growth using scoped APIs

nikomatsakis:
The old scoped code was a bit too eager to pop work off the deque, resulting in linear stack growth in some scenarios -- precisely the thing that `scope` is designed to avoid!

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -151,6 +151,10 @@ pub fn join<A,B,RA,RB>(oper_a: A,
         let job_b = StackJob::new(oper_b, SpinLatch::new());
         (*worker_thread).push(job_b.as_job_ref());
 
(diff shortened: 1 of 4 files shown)


nikomatsakis:
So I did some measurements using rayon-demo. It's hard to tell if this has any impact on performance. =) It *seems* to be in the noise, though perhaps a *slight* drop (a few percent, but also only on the microbenchmarks -- stuff like `nbody`, `quicksort`, and `mergesort` are unaffected). There may be a faster way to formula `join`, I have to look closely. Regardless, I think this is basically an essential change -- we **have** to be able to run nested scopes without growing the stack. So I'm going to lang it.

PR #139, opened 2016-11-12
parallel TSP solver based on the scope construct

nikomatsakis:
This is not intended as a truly competitive implementation, at least not
yet =), but more as a stress test and general benchmark. It works by
exhaustively exploring the space of possible tours in parallel. At each
point, it retains a (not very clever) heuristic of a lower bound on the
cost to complete the tour. Behind the scenes, a priority queue is used
so that each parallel thread is exploring the most profitable
possibilities. After a certain point, we fall back to sequential
execution.

The current implementation uses the `scope` construct. There are various
enhancements I would like to make:

- perhaps other lower bound heuristics, such as leveraging the triangle inequality
- replace the mutex with a flat combining lock, once we offer those
- make a "pure sequential" variant to compare against, although simply
  setting a sequential threshold of N may be "good enough"
- replace the manual sequential threshold with a dynamic variant,
  not sure yet what this would look like
  - I imagine something like pushing a minimum amount of work
  - and then some kind of quick check on the length of the local queue
    to decide whether to push more


edre:
I was looking into something like this earlier and found this writeup of a SAT solver in cilk. The architecture is similar to what you have here, but they make use of cilk inlets that have no rayon equivalent. http://courses.csail.mit.edu/6.884/spring10/projects/lee_zhang_report.pdf

I think a priority queue may be a good example of where to use your flat combiner. A binomial heap can merge a batch of entries more efficiently than adding them sequentially.

nikomatsakis:
Indeed, I was planning on porting this to use flat combining once I have a chance to hack on that code a bit more to make it more reusable, as we discussed

PR #143, opened 2016-11-16
demo: add parallel matrix multiplication

edre:
Includes both row-major order and Z order, demonstrating that
cache-obliviousness is more important than parallelism.
Includes normal O(n³) parallel version and Strassen algorithm.


nikomatsakis:
@edre can you add some relevant `#[bench]` results? I'd primarily be interested in whatever configuration you would use to detect regressions in Rayon's core performance (i.e., the *best* one, probably).

nikomatsakis:
Nice :)

Issue #102, opened 2016-10-03
into_par_iter not defined for u64


russel:
I find that:

``` rust
(0..n).into_par_iter…
```

fails if `n` is `u64`, but if I do:

``` rust
(0..n as u32).into_par_iter…
```

everything works fine. Well mostly, but that is a different issue.

cuviper:
That's because a `Range<u64>` is not an `ExactSizeIterator`, since its total length might not fit in a `usize`.  That concept from serial iterators also applies to rayon, since its lengths are also `usize`.

russel:
This is sad since it leads to ugly `as u32` and `as u64` scattered around the place, and problems if you want say 10^30 items. However it sounds like this is a hard constraint and so not fixable. If so then this issue clearly has to be closed.

cuviper:
10^30 items?!?!  At 1ns per item, I figure that will take about 32 *trillion* years of cpu time.  You must have massive parallelization plans... :wink:

It's more realistically annoying to be limited by u32::MAX on 32-bit platforms, this is true, and also annoying that 32-bit concerns are still limiting cases that 64-bit machines could handle just fine.  Rayon is consistent with similar decisions in the standard library, for better or worse.

russel:
10^30 was an over-estimate. :-) Though if we want to do calculations for all the proton in the universe then that is 10^80 – at a rough guess.

For these sorts of computations, Rust has to compete with Chapel (and X10). The computers that excel at these things have many thousands of processors, and cost many millions of pounds.

It seems though the Rust standard library is dictating and they have made a very 32-bit decision in a clearly 64-bit world. :-( I will close this issue, but with a small tear.

cuviper:
If you only care about running on 64-bit, at least you can cast your u64 to/from usize without loss.

russel:
I think for production code I would use usize, and thus avoid the problem. But for some trivial language comparison codes, I have to guarantee 64-bit so that comparisons are not muddied by the 32-bit/64-bit integer and float numbers.

nikomatsakis:
I think it's not out of the question to address this shortcoming. Not that I have a great idea how to do it.

nikomatsakis:
But I'm going to re-open this in any case.

Binero:
Could you add a switch so it works on 64 bit systems?

cuviper:
See #144, which implements these as unindexed producers.

nikomatsakis:
Closed by #144.

Issue #133, opened 2016-11-04
`par_split_whitespace` for string


nikomatsakis:
I was thinking that if you have an `&str`, you can split it into words/lines in parallel in a pretty simple way. Let's just think about making a `par_split_whitespace` that acts like `split_whitespace`:

- First, pick a point in the middle of the string as measured in bytes; if this is in the middle of a utf-8 character, move to the start of the utf-8 character (you'd need to look at the raw `as_bytes` here, I don't think we have a helper for this -- but utf-8 is designed to make such "alignment" easy, you just have to check for bytes with the `0x80` bit set I think).
- Next, check if this is a whitespace character. If so, move backwards until you find a non-whitespace character.
- Split the string in half at this point.

This will give you two roughly strings of roughly even length that we can further subdivide.

The same principle can be applied to make a `par_lines` that acts like `lines`.

I don't think you can do this in the general case for `split()`, since it takes arbitrary functions, but you could probably do it for many cases (e.g., an arbitrary split-string).

This seems like a fun starter task for hacking on Rayon. If you're interested in taking it up, please reach out through the gitter channel.

UPDATE: `par_lines()` has been implemented, but not `par_split_whitespace()`; see [this comment](https://github.com/nikomatsakis/rayon/issues/133#issuecomment-272046596) for details.

cuviper:
> but utf-8 is designed to make such "alignment" easy, you just have to check for bytes with the 0x80 bit set I think

Yes, need to look at the upper two bits, so something like `c & 0xc0 == 0x80` is a continuation byte.

One thing that may be challenging is that you don't know how many substrings there will be, unless you pre-scan the whole thing.  Even the number of characters is not known, only the number of bytes.  The current par_iter internals only work with an `ExactParallelIterator` at the base.

edre:
&str has a handy is_char_boundary to do the bit twidding for you.

You can treat the string as a BoundedParallelIterator where every byte is potentially the start of a split. Shouldn't be any different semantics than a filter.

cuviper:
Yes, treating it as bounded should be fine.  My point is that `internal::bridge` is not set up to handle that as a starting point.  So you might need something like `(0..s.len())` as your base iterator, then filter that for offsets that are characters and match as a separator.  This seems much different than what @nikomatsakis outlined.

nikomatsakis:
@cuviper you don't have to use `bridge`, you can just drive the consumer directly

nikomatsakis:
[...]

PR #144, opened 2016-11-16
Implement parallel Range<i64> and Range<u64>

cuviper:
These couldn't be exact iterators, because they could exceed a `usize`
length on 32-bit targets.  So this adds a `bridge_unindexed` helper
function to drive consumers directly with imprecise splits.  64-bit
ranges can use this while happily splitting in normal halves.

### DIFF
--- a/src/par_iter/internal.rs
+++ b/src/par_iter/internal.rs
@@ -83,6 +83,13 @@ pub trait UnindexedConsumer<ITEM>: Consumer<ITEM> {
     fn to_reducer(&self) -> Self::Reducer;
 }
 
+/// An unindexed producer that doesn't know its exact length.
+/// (or can't represent its known length in a `usize`)
+pub trait UnindexedProducer: IntoIterator + Send + Sized {
+    fn can_split(&self) -> bool;
+    fn split(self) -> (Self, Self);
+}
+
 /// A splitter controls the policy for splitting into smaller work items.
 #[derive(Clone, Copy)]
 enum Splitter {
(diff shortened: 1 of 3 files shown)


cuviper:
I wrote `bridge_unindexed` while poking at #133, but this seems like an easy test bed for it.

src/par_iter/internal.rs
```diff
@@ -83,6 +83,13 @@ pub trait UnindexedConsumer<ITEM>: Consumer<ITEM> {
     fn to_reducer(&self) -> Self::Reducer;
 }
 
+/// An unindexed producer that doesn't know its exact length.
+/// (or can't represent its known length in a `usize`)
+pub trait UnindexedProducer: IntoIterator + Send + Sized {
```
nikomatsakis: Great! One nit: I've always imagined that we would have `Producer: UnindexedProducer`; but I guess it's just kind of a lot of boilerplate.

I would think something like this:

```rust
impl<T: Producer> UnindexedProducer for T {
    fn can_split(&self) -> bool { self.len() != 0 }
    fn split(&self) -> (Self, Self) { self.split_at(self.len() / 2) }
}
```

But I think we'd have to add the `len` method to `Producer` -- not that there is a good reason we don't have it.

Mostly I like the symmetry where you have:

- `Consumer` and `UnindexedConsumer: Consumed`
- `UnindexedProducer` and `Producer: UnindexedProducer`

nikomatsakis:
Going to merge this anyway. We can maybe refactor to give the producer traits more of a hierarchy later.

cuviper:
Thanks!

I agree that `Producer: UnindexedProducer` *looks* nice, but I'm not sure it's actually useful.  Is there any case where you'd have a `Producer` and might prefer to use it unindexed anyway?

BTW, this should close #102.  I had forgotten that issue existed when I wrote this, but it must have been floating in my subconscious already. :)

nikomatsakis:
@cuviper 

> I agree that Producer: UnindexedProducer looks nice, but I'm not sure it's actually useful. Is there any case where you'd have a Producer and might prefer to use it unindexed anyway?

Yeah, I can't think of any just now. =) I still like the idea of having the setup exposed (and it couldn't hurt), but I don't think it really matters, and if it does we can always do it then.

PR #153, opened 2016-11-17
Rename mul to product

jonathandturner:
Starting simple :)  Looking at https://github.com/nikomatsakis/rayon/issues/149

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -21,8 +21,8 @@ use self::filter_map::FilterMap;
 use self::flat_map::FlatMap;
 use self::from_par_iter::FromParallelIterator;
 use self::map::{Map, MapFn, MapCloned, MapInspect};
-use self::reduce::{reduce, ReduceOp, SumOp, MulOp, MinOp, MaxOp,
-                   ReduceWithIdentityOp, SUM, MUL, MIN, MAX};
+use self::reduce::{reduce, ReduceOp, SumOp, ProductOp, MinOp, MaxOp,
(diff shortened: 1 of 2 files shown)


nikomatsakis:
@jonathandturner can you keep the old method with a `#[deprecated]` annotation?

nikomatsakis:
Also, hmm, it seems that `rayon-demo` no longer builds. Not quite sure what's up with that?

jonathandturner:
does the rayon-demo not build when `cargo test` runs?

nikomatsakis:
@jonathandturner no, it's a separate project; not sure if there's a way to make them both build

jonathandturner:
@nikomatsakis will take a peek to see if there's a way

src/par_iter/mod.rs
```diff
@@ -459,6 +459,14 @@ pub trait ParallelIterator: Sized {
         reduce(self, MUL)
     }
 
+    /// DEPRECATED. Use .product() instead.
+    #[deprecated]
```
nikomatsakis: should be something like this:

```rust
#[deprecated(since = "v0.6.0", note = "name changed to `product()` to match sequential iterators)]
```

I put 0.6.0 because I think the next release includes a few breaking changes...but I can't remember for sure...

nikomatsakis: nah, make it 0.5.1 -- seems like we haven't had any breaking changes lately

jonathandturner:
Rebased.  Fixed the failing issues

nikomatsakis:
My hero!

PR #154, opened 2016-11-17
reorganize tests so that `cargo test` passes

nikomatsakis:
Before, tests that used unstable features were mixed in. That meant that `cargo test` by itself didn't pass, which would send a weird signal. It's still the case that `cargo test --features unstable` is the **right** way to run the tests though.


cuviper:
Probably should add a bare `cargo test` to travis too.

nikomatsakis:
@cuviper yeah I was thinking the same. Also I like how I broke the unstable tests. =)

PR #159, opened 2016-11-18
Match sequential iterator API for comparison-based reducers

### LINKED ISSUE #148: `max`, `min`, `max_by_key`, `min_by_key` should match sequential iterators
The parallel `max` and `min` have a different signature from the sequential variant. They return `Self::Item` instead of `Option<Self::Item>`. They also don't work on all `Ord` types. (These are related.) It'd be good to change them to match, I think. In so doing, we can also add support for `max_by_key` and `min_by_key`. This would also simplify the implementation to boot!

The idea is to replace the [existing, oddball methods](https://github.com/nikomatsakis/rayon/blob/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/mod.rs#L350-L354) with methods [that invoke `reduce_with()`](https://github.com/nikomatsakis/rayon/blob/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/mod.rs#L217-L221). They should work for any iterator where `Self::Item: Ord`.

tapeinosyne:
Closes #148.

(The failing tests should be unrelated; they fail on master as well.)

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -12,7 +12,7 @@
 //! good place to start.
 
 use std::f64;
-use std::cmp;
+use std::cmp::{self, Ordering};
 use std::ops::Fn;
 use self::chain::ChainIter;
 use self::collect::collect_into;
@@ -484,6 +484,25 @@ pub trait ParallelIterator: Sized {
         self.reduce_with(cmp::min)
     }
 
+    /// Computes the item that yields the minimum value for the given
+    /// function. If the iterator is empty, `None` is returned;
+    /// otherwise, `Some(item)` is returned.
+    ///
+    /// Note that the order in which the items will be reduced is not
+    /// specified, so if the `Ord` impl is not truly associative, then
+    /// the results are not deterministic.
+    fn min_by_key<K, F>(self, f: F) -> Option<Self::Item>
+        where K: Ord + Send,
+              F: Sync + Fn(&Self::Item) -> K,
+    {
+        self.map(|x| (f(&x), x))
+            .reduce_with(|a, b| match (a.0).cmp(&b.0) {
+                Ordering::Greater => b,
+                _ => a
+            })
+            .map(|(_, x)| x)
+    }
+
     /// Computes the maximum of all the items in the iterator. If the
     /// iterator is empty, `None` is returned; otherwise, `Some(max)`
     /// is returned.
@@ -497,6 +516,25 @@ pub trait ParallelIterator: Sized {
         where Self::Item: Ord
     {
         self.reduce_with(cmp::max)
+    }
+
+    /// Computes the item that yields the maximum value for the given
+    /// function. If the iterator is empty, `None` is returned;
+    /// otherwise, `Some(item)` is returned.
+    ///
+    /// Note that the order in which the items will be reduced is not
+    /// specified, so if the `Ord` impl is not truly associative, then
+    /// the results are not deterministic.
+    fn max_by_key<K, F>(self, f: F) -> Option<Self::Item>
+        where K: Ord + Send,
+              F: Sync + Fn(&Self::Item) -> K,
+    {
+        self.map(|x| (f(&x), x))
+            .reduce_with(|a, b| match (a.0).cmp(&b.0) {
+                Ordering::Greater => a,
+                _ => b
+            })
+            .map(|(_, x)| x)
     }
 
     /// Takes two iterators and creates a new iterator over both.
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -12,6 +12,7 @@
 //! good place to start.
 
 use std::f64;
+use std::cmp;
 use std::ops::Fn;
 use self::chain::ChainIter;
 use self::collect::collect_into;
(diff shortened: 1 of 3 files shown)


utkarshkukreti:
Here's a test I had written when I was working on this issue that _might_ be useful:

```diff
diff --git a/src/par_iter/test.rs b/src/par_iter/test.rs
index 5289cf1..ba9f615 100644
--- a/src/par_iter/test.rs
+++ b/src/par_iter/test.rs
@@ -1,3 +1,5 @@
+extern crate rand;
+
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 use super::*;
@@ -729,3 +731,14 @@ pub fn par_iter_collect_linked_list_flat_map_filter() {
     assert_eq!(b, c);
 }
 
+#[test]
+pub fn min_max() {
+    use rand::Rng;
+
+    let a: Vec<i32> = rand::thread_rng().gen_iter().take(1024).collect();
+    for i in 0..a.len() {
+        let slice = &a[..i];
+        assert_eq!(slice.par_iter().min(), slice.iter().min());
+        assert_eq!(slice.par_iter().max(), slice.iter().max());
+    }
+}
```

nikomatsakis:
@utkarshkukreti the test seems good, but maybe we can rewrite it to use a fixed seed, so that it is more deterministic? (random failures are no fun)

src/par_iter/mod.rs
```diff
@@ -468,35 +469,70 @@ pub trait ParallelIterator: Sized {
...
+    /// Note that the order in which the items will be reduced is not
+    /// specified, so if the `Ord` impl is not truly associative, then
+    /// the results are not deterministic.
+    fn min_by_key<K, F>(self, f: F) -> Option<Self::Item>
+        where K: Ord,
+              F: Sync + Fn(&Self::Item) -> K,
+    {
+        self.reduce_with(|a, b| match f(&a).cmp(&f(&b)) {
+            Ordering::Less => a,
+            Ordering::Equal=> b,
```
nikomatsakis: I *think* we want `a` in the case where they are equal, to preserve stability. At least this is what I think sequential iterators do. Can you edit `src/par_iter/test.rs` and add some tests for `min_by_key`, and specifically check for cases where two things compare equal?

Something like this, but probably with a bigger array :)

```rust
#[test]
fn stable_min_by_key() {
    let x = vec![(0, 'a'), (22, 'b'), (0, 'c')];
    assert_eq!(
        x.par_iter().weight_max().min_by_key(|pair| pair.0),
        x.iter().min_by_key(|pair| pair.0)
    );
}
```

src/par_iter/mod.rs
```diff
@@ -468,35 +469,70 @@ pub trait ParallelIterator: Sized {
...
+    /// specified, so if the `Ord` impl is not truly associative, then
+    /// the results are not deterministic.
+    fn max_by_key<K, F>(self, f: F) -> Option<Self::Item>
+        where K: Ord,
+              F: Sync + Fn(&Self::Item) -> K,
     {
-        reduce(self, MAX)
+        self.reduce_with(|a, b| match f(&a).cmp(&f(&b)) {
+            Ordering::Less => b,
+            Ordering::Equal=> b,
```
nikomatsakis: Same here, I think this is not right in the case of `Equal` -- but we should just test it to be sure.

src/par_iter/mod.rs
```diff
@@ -468,35 +469,70 @@ pub trait ParallelIterator: Sized {
...
+    ///
+    /// Note that the order in which the items will be reduced is not
+    /// specified, so if the `Ord` impl is not truly associative, then
+    /// the results are not deterministic.
+    fn max_by_key<K, F>(self, f: F) -> Option<Self::Item>
+        where K: Ord,
+              F: Sync + Fn(&Self::Item) -> K,
     {
-        reduce(self, MAX)
+        self.reduce_with(|a, b| match f(&a).cmp(&f(&b)) {
```
cuviper: I think it would be better if this only called `f` once per item.  Something like:
```rust
self.map(|x| (f(&x), x)).reduce_with(...).map(|(_, x)| x)
```
That will require `K: Send` though.

utkarshkukreti:
@nikomatsakis how about this?

```rust
#[test]
pub fn min_max() {
    use rand::{Rng, SeedableRng, XorShiftRng};

    let mut rng = XorShiftRng::from_seed([14159, 26535, 89793, 23846]);
    let a: Vec<i32> = rng.gen_iter().take(1024).collect();
    for i in 0..a.len() {
        let slice = &a[..i];
        assert_eq!(slice.par_iter().min(), slice.iter().min());
        assert_eq!(slice.par_iter().max(), slice.iter().max());
    }
}
```

Feel free to use it @tapeinosyne if @nikomatsakis likes it. :)

src/par_iter/mod.rs
tapeinosyne: Yes, applying the function once and doing the Schwartzian-style transform is perfectly reasonable if we agree that the `Send` bound is acceptable. (Note that `Iterator` also calls the function twice per element.)

src/par_iter/mod.rs
tapeinosyne: That was my original thinking, but the std::iter variant reportedly favors the rightmost element for both `max_by` and `min_by`.

src/par_iter/mod.rs
nikomatsakis: Yeah, good catch. This is in fact what sequential iterators do, so I agree we should match. I think the `Send` bound is fine.

tapeinosyne:
@utkarshkukreti, note that `for i in 0..a.len() { let slice = &a[..i] …` skips the last element, but otherwise it looks good.

nikomatsakis:
(test looks good to me )

tapeinosyne:
Addressed comments in rebased branch.

PR #160, opened 2016-11-18
add a description of `ProducerCallback`

nikomatsakis:


nikomatsakis:
OK, so, I talked to @cuviper and we agreed it'd be good to make this a bit more top-down, but I'm going to land it anyway for now since I don't have time to do that yet.

PR #161, opened 2016-11-18
add a rough-and-tumble `spawn_async` API

nikomatsakis:
For some use cases, its useful to be able to delegate work from your
main thread that can run asynchronously.

Some work items:

- [ ] more tests =)
- [ ] update docs to mention `scope()` as a preferred alternative
- [ ] refactor `HeapJob` and `StackJob` a bit (probably defer for later)

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,7 +1,7 @@
 use latch::{LockLatch, SpinLatch};
 #[allow(unused_imports)]
 use log::Event::*;
-use job::StackJob;
+use job::{HeapJob, StackJob, JobMode};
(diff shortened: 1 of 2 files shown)


nikomatsakis:
After some discussion with @cuviper, we were thinking that -- before landing -- it'd be good to have this return something with which you can poll the result. More-or-less like the example, but one that is integrated with the worker so that if you poll and you are on a rayon worker, you steal.

nikomatsakis:
Gonna close this while I play around with how to rework it. I'm planning on also doing some (optional) futures-rs integration.

Issue #147, opened 2016-11-17
support `skip` and `take` on exact parallel iterators


nikomatsakis:
We should support the standard `skip()` and `take()` methods from sequential iterators. I am only envisioning supporting these for *indexed* parallel iterators; in that case, it should be fairly straightforward, as one can simply call `split_at()` to divide out the relevant items.

The steps would be as follows (this just describes adding `skip()`; `take()` would be similar):

- We would be modeling this change roughly on the [existing `Enumerate` combinator](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L6-L8); so we'll add a type like `Skip<M>`
- Like `Enumerate`, `Skip<M>` will implement the various parallel iterators traits, but [only if `M: IndexedParallelIterator`](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L16-L17)
- By far the most complex part is implementing [the `with_producer()` method](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L51-L72). As [described in the parallel iterator README](https://github.com/nikomatsakis/rayon/blob/master/src/par_iter/README.md#what-on-earth-is-producercallback), this method is basically a callback scheme:
    - The heart of it is that you [ask your base producer (the one that we are going to be skipping over) to call you back with a producer](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0//src/par_iter/enumerate.rs#L54).
    - But it can't use Rust's closures, so you have to [make a custom struct](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L56-L58) which [implements the `ProducerCallback` trait](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L60-L62).
    - Ultimately, [the `callback()` method of this struct will be invoked](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L67-L69) with the base producer.
[...]

PR #162, opened 2016-11-19
Fixes issue #147

martinhath:
These commits implements `skip` and `take` on exact parallel iterators.

I am very unsure if I have put the bounds checking in the right place; right now it is in `len()`, using the fact that it's `&mut self`. However, maybe there is no guarantee for `len()` to have been called before `with_producer`, in which case we will crash.

I'm a first time contributor (and almost first time user) of Rayon, and most of the code is taken from `enumerate.rs`, so I'm sorry if there are obvious errors here :)

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -25,6 +25,7 @@ use self::map::{Map, MapFn, MapCloned, MapInspect};
 use self::reduce::{reduce, ReduceOp, SumOp, ProductOp,
                    ReduceWithIdentityOp, SUM, PRODUCT};
 use self::skip::Skip;
+use self::take::Take;
 use self::internal::*;
 use self::weight::Weight;
 use self::zip::ZipIter;
@@ -43,6 +44,7 @@ pub mod for_each;
 pub mod fold;
 pub mod reduce;
 pub mod skip;
+pub mod take;
 pub mod slice;
 pub mod slice_mut;
 pub mod map;
@@ -690,6 +692,12 @@ pub trait IndexedParallelIterator: ExactParallelIterator {
         use std::cmp::min;
         let n = min(n, self.len());
         Skip::new(self, n)
+    }
+
+    fn take(mut self, n: usize) -> Take<Self> {
+        use std::cmp::min;
+        let n = min(n, self.len());
+        Take::new(self, n)
     }
 
     /// Searches for **some** item in the parallel iterator that
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -688,15 +688,11 @@ pub trait IndexedParallelIterator: ExactParallelIterator {
         Enumerate::new(self)
     }
 
-    fn skip(mut self, n: usize) -> Skip<Self> {
-        use std::cmp::min;
-        let n = min(n, self.len());
+    fn skip(self, n: usize) -> Skip<Self> {
         Skip::new(self, n)
     }
 
-    fn take(mut self, n: usize) -> Take<Self> {
-        use std::cmp::min;
-        let n = min(n, self.len());
+    fn take(self, n: usize) -> Take<Self> {
         Take::new(self, n)
     }
 
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -54,6 +54,7 @@ pub mod range;
 pub mod vec;
 pub mod option;
 pub mod collections;
+pub mod noop;
 
 #[cfg(test)]
 mod test;
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -688,10 +688,12 @@ pub trait IndexedParallelIterator: ExactParallelIterator {
         Enumerate::new(self)
     }
 
+    /// Creates an iterator that skips the first `n` elements.
     fn skip(self, n: usize) -> Skip<Self> {
         Skip::new(self, n)
     }
 
+    /// Creates an iterator that yields the first `n` elements.
     fn take(self, n: usize) -> Take<Self> {
         Take::new(self, n)
     }
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -24,6 +24,7 @@ use self::from_par_iter::FromParallelIterator;
 use self::map::{Map, MapFn, MapCloned, MapInspect};
 use self::reduce::{reduce, ReduceOp, SumOp, ProductOp,
                    ReduceWithIdentityOp, SUM, PRODUCT};
+use self::skip::Skip;
 use self::internal::*;
 use self::weight::Weight;
 use self::zip::ZipIter;
@@ -41,6 +42,7 @@ pub mod len;
 pub mod for_each;
 pub mod fold;
 pub mod reduce;
+pub mod skip;
 pub mod slice;
 pub mod slice_mut;
 pub mod map;
@@ -682,6 +684,12 @@ pub trait IndexedParallelIterator: ExactParallelIterator {
     /// Yields an index along with each item.
     fn enumerate(self) -> Enumerate<Self> {
         Enumerate::new(self)
+    }
+
+    fn skip(mut self, n: usize) -> Skip<Self> {
+        use std::cmp::min;
+        let n = min(n, self.len());
+        Skip::new(self, n)
     }
 
     /// Searches for **some** item in the parallel iterator that
--- a/src/par_iter/noop.rs
+++ b/src/par_iter/noop.rs
@@ -47,3 +47,9 @@ impl<ITEM> UnindexedConsumer<ITEM> for NoopConsumer {
         NoopReducer
     }
 }
+
+pub struct NoopReducer;
+
+impl Reducer<()> for NoopReducer {
+    fn reduce(self, _left: (), _right: ()) { }
+}
--- a/src/par_iter/noop.rs
+++ b/src/par_iter/noop.rs
@@ -1,0 +1,49 @@
+use super::internal::*;
+
+pub struct NoopConsumer;
+
+impl NoopConsumer {
+    pub fn new() -> Self {
+        NoopConsumer
+    }
+}
+
+impl<ITEM> Consumer<ITEM> for NoopConsumer
+{
+    type Folder = NoopConsumer;
+    type Reducer = NoopReducer;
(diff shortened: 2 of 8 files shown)


nikomatsakis:
Nice!

src/par_iter/skip.rs
```diff
@@ -0,0 +1,78 @@
...
+        bridge(self, consumer)
+    }
+}
+
+impl<M> ExactParallelIterator for Skip<M>
+    where M: IndexedParallelIterator
+{
+    fn len(&mut self) -> usize {
+        let base_len = self.base.len();
+        if self.n > base_len {
```
nikomatsakis: Hmm, I think I would not mutate `self.n` here, but instead just return `base_len.saturating_sub(self.n)`. It does seem a bit surprising to have `len()` mutate state. I should probably change it to `&self`; I think I just wanted to keep things maximally flexible in case computing the length required (e.g.) updating a cache or something. Not that I have an example where that would be the case.

src/par_iter/skip.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+impl<M> IndexedParallelIterator for Skip<M>
+    where M: IndexedParallelIterator
+{
+    fn with_producer<CB>(self, callback: CB) -> CB::Output
+        where CB: ProducerCallback<Self::Item>
+    {
+        return self.base.with_producer(Callback { callback: callback,
```
nikomatsakis: then here I would call:

```rust
let base_len = self.base.len();
let n = cmp::min(self.n, base_len);
return self.base.with_producer(Callback { callback: callback, n: n });
```

src/par_iter/test.rs
```diff
@@ -179,6 +179,57 @@ pub fn check_increment() {
 }
 
 #[test]
+pub fn check_skip() {
+    let a: Vec<usize> = (0..10).collect();
```
nikomatsakis: Good to have a test; but why not make it longer...say 1024? can't hurt...

src/par_iter/test.rs
```diff
@@ -179,6 +179,57 @@ pub fn check_increment() {
...
+
+    let mut b = vec![];
+    a.par_iter()
```
nikomatsakis: Nit: I would do `let b: Vec<_> = a.par_iter()...collect();` and  then compare against the same iterator chain, but sequential (i.e., just check that par/seq get same result).

Also, we usually add a `weight_max()` which forces maximum splitting. It tends to expose bugs.

src/par_iter/test.rs
```diff
@@ -179,6 +179,57 @@ pub fn check_increment() {
...
+        .skip(5)
+        .cloned()
+        .collect_into(&mut b);
+    assert_eq!(b, vec![5, 6, 7, 8, 9]);
+
+    let mut b = vec![];
```
nikomatsakis: Good idea to test the extreme cases =)

cuviper:
> I am very unsure if I have put the bounds checking in the right place; right now it is in len(), using the fact that it's &mut self. However, maybe there is no guarantee for len() to have been called before with_producer, in which case we will crash.

It wouldn't hurt to just normalize that right away in `new`, like `min(base.len(), n)`.

src/par_iter/test.rs
nikomatsakis: @cuviper convinced me that we shouldn't call `weight_max()` -- but do make it longer :)

martinhath:
Thanks for the feedback! 
Should I do the `saturating_sub` approach, or make `new` bounded by `IndexedParallelIterator`, and do it there, as @cuviper suggested?

src/par_iter/mod.rs
```diff
@@ -646,6 +650,14 @@ pub trait IndexedParallelIterator: ExactParallelIterator {
         Enumerate::new(self)
     }
 
+    fn skip(self, n: usize) -> Skip<Self> {
+        Skip::new(self, n)
+    }
+
+    fn take(self, n: usize) -> Take<Self> {
```
cuviper: These should have at least a little documentation, even if there's not much to say. :)

nikomatsakis:
@martinhath either one should be fine; I guess that doing it in `new()` seems slightly nicer

src/par_iter/skip.rs
bwo: I was just writing up a PR for this very issue and had run into the same problem as you're discussing here, and tried this solution. But you can't call `self.base.len()` here:

```
error: cannot borrow immutable field `self.base` as mutable
  --> src/par_iter/take.rs:54:17
   |
54 |         let n = self.base.len();
   |                 ^^^^^^^^^

```

cuviper:
This is more philosophical, but if we're matching `Iterator::skip` semantics, it may actually be surprising to just throw away the skipped items.  For example, if those had a `map` with some side effect like an atomic increment, that would still execute in serial code.  If we want to emulate that, I think you could drive the discarded half with a noop reducer.

On the other hand, discarding the unused part of `take` is exactly like serial code, so that should be fine.

nikomatsakis:
Hmm, I suppose that's true. I don't think I would drive the discarded half with a reducer: rather something like an empty `for_each`. I wonder if LLVM could successfully just optimize the whole thing away if it is side-effect free. Maybe sometimes, but I wouldn't bet on it. 

If we really cared, we could probably propagate a flag to indicate whether there could possibly be side-effects.

src/par_iter/skip.rs
bwo: fwiw : https://github.com/nikomatsakis/rayon/compare/master...bwo:feature/skip-and-take

martinhath:
Should I try to fix matching `Iterator::skip` semantics?

src/par_iter/skip.rs
nikomatsakis: just fyi, instead of `let mut me = self` you can also write `mut self` in the fn signature

src/par_iter/skip.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        impl<ITEM, CB> ProducerCallback<ITEM> for Callback<CB>
+            where CB: ProducerCallback<ITEM>
+        {
+            type Output = CB::Output;
+            fn callback<P>(self, base: P) -> CB::Output
+                where P: Producer<Item=ITEM>
+            {
+                let (_, producer) = base.split_at(self.n);
```
nikomatsakis: So I guess if we were going to match the semantics of sequential `skip` we might change this to:

```rust
let (before_skip, after_skip) = base.split_at(self.n);
bridge_producer_consumer(before_skip, NoopConsumer::new());
self.callback.callback(after_skip);
```

(We'd have to write the no-op consumer, I don't think that exists.)

It feels a *bit* weird to call that *here*, in the `with_producer` method -- but it's probably the right place to do it. The other place would be `drive()` but, we'd still have to convert into a producer, and after all, all that `drive()` does is call `bridge()` which calls `with_producer()`.

We should also make some tests that observe side-effects from the skipped entries, e.g. using `map()` or `inspect()`.

martinhath: Alright, I'm on it!

nikomatsakis: (To make the no-op consumer, it's probably easiest to copy the `for_each::ForEachConsumer` and strip out most of its logic.)

martinhath:
I got something to work :) The test got kind of awkward (I summed to an `Arc<Mutex<usize>>`) but that was the first thing that came to mind.

src/par_iter/skip.rs
```diff
@@ -0,0 +1,80 @@
...
+        impl<ITEM, CB> ProducerCallback<ITEM> for Callback<CB>
+            where CB: ProducerCallback<ITEM>
+        {
+            type Output = CB::Output;
+            fn callback<P>(self, base: P) -> CB::Output
+                where P: Producer<Item=ITEM>
+            {
+                let (before_skip, after_skip) = base.split_at(self.n);
+                bridge_producer_consumer(self.n, Splitter::Cost(self.n as f64),
+                                         before_skip, noop::NoopConsumer::new());
```
cuviper: Hmm.  I'm not sure I like making the `Splitter` and `bridge_producer_consumer`, and in particular here the `Cost` is not a great choice as we're phasing that out.  *If* we use `Cost` at all, it should follow the same thief/cost logic as `bridge` does.

Maybe that inner part of `bridge` should be factored out, for when you have a length, producer, and consumer -- keeping the splitter logic private to that module.

src/par_iter/noop.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl<ITEM> UnindexedConsumer<ITEM> for NoopConsumer {
+    fn split_off(&self) -> Self {
+        NoopConsumer
+    }
+
+    fn to_reducer(&self) -> NoopReducer {
+        NoopReducer
+    }
+}
```
cuviper: Just for good code organization, we should move the `NoopReducer` into this module too.

src/par_iter/test.rs
```diff
@@ -180,6 +180,53 @@ pub fn check_increment() {
...
+
+    let mut v1 = Vec::new();
+    a.par_iter().skip(0).collect_into(&mut v1);
+    let v2 = a.iter().skip(0).collect::<Vec<_>>();
+    assert_eq!(v1, v2);
+
+    // Check that the skipped elements side effects are executed
+    use std::sync::{Arc, Mutex};
+    let num = Arc::new(Mutex::new(0usize));
+    a.par_iter().map(|&n| *num.lock().unwrap() += n).skip(512).count();
```
cuviper: An `AtomicUsize` should suffice, but it doesn't make much difference either way for testing purposes.

martinhath: I replaced it with `AtomicUsize` in the newest commit, as I found it to be slightly prettier. Having never used the module, I used `Ordering::Relaxed` on both `fetch_add` and `load`. Could that be a problem?

martinhath:
When we are ready to merge, do you want me to squash some commits, eg. the two latest?

src/par_iter/test.rs
cuviper: `Ordering::Relaxed` is fine, since this atomic not related to any other memory values.

cuviper:
We're not too worried about squashing, especially when the intermediate steps are all functional.  Thanks!

PR #163, opened 2016-11-19
Fixing Issue 152

### LINKED ISSUE #152: implement `cmp`, `partial_cmp`, and friends for parallel iterator
To achieve parity with sequential iterators, we need to support `cmp` and friends. These are methods which basically zip two iterators and compare corresponding elements; so long as the elements are equal, it proceeds to the next. 

We can parallelize this fairly easily by using the existing zip, map, and reduce combinators I think. Roughly speaking I would do something like `left.zip(right).map(Ord::cmp).reduce(|| Equal, |cmp_a, cmp_b| if cmp_a != Equal { cmp_b } else { cmp_a })`. This won't, however, be the most efficient thing, because it will do *all* the comparisons, even if it already knows the final answer.

To support the most efficient thing, we probably want `find_first` (#151), I should think, so that we can basically search for the first "non-equal" result.

Here are the full set of methods:

- `fn cmp<I>(self, other: I) -> Ordering where I: IntoIterator<Item=Self::Item>, Self::Item: Ord { ... }`
- `fn partial_cmp<I>(self, other: I) -> Option<Ordering> where I: IntoIterator, Self::Item: PartialOrd<I::Item> { ... }`
- `fn eq<I>(self, other: I) -> bool where I: IntoIterator, Self::Item: PartialEq<I::Item> { ... }`
- `fn ne<I>(self, other: I) -> bool where I: IntoIterator, Self::Item: PartialEq<I::Item> { ... }`
- `fn lt<I>(self, other: I) -> bool where I: IntoIterator, Self::Item: PartialOrd<I::Item> { ... }`
- `fn le<I>(self, other: I) -> bool where I: IntoIterator, Self::Item: PartialOrd<I::Item> { ... }`
- `fn gt<I>(self, other: I) -> bool where I: IntoIterator, Self::Item: PartialOrd<I::Item> { ... }`
- `fn ge<I>(self, other: I) -> bool where I: IntoIterator, Self::Item: PartialOrd<I::Item> { ... }`

ChristopherDavenport:
Working as group which is why the pull request is being initiated early so that we can each merge into this common branch.

First Commit Implements cmp
Uses the existing zip, map, and reduce combinators to implement cmp.

Closes #152

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -22,8 +22,7 @@ use self::filter_map::FilterMap;
 use self::flat_map::FlatMap;
 use self::from_par_iter::FromParallelIterator;
 use self::map::{Map, MapFn, MapCloned, MapInspect};
-use self::reduce::{reduce, ReduceOp, SumOp, ProductOp,
-                   ReduceWithIdentityOp, SUM, PRODUCT};
+use self::reduce::{reduce, ReduceOp, SumOp, ProductOp, ReduceWithIdentityOp, SUM, PRODUCT};
 use self::skip::Skip;
 use self::take::Take;
 use self::internal::*;
@@ -60,14 +59,14 @@ pub mod noop;
 mod test;
 
 pub trait IntoParallelIterator {
-    type Iter: ParallelIterator<Item=Self::Item>;
+    type Iter: ParallelIterator<Item = Self::Item>;
     type Item: Send;
 
     fn into_par_iter(self) -> Self::Iter;
 }
 
 pub trait IntoParallelRefIterator<'data> {
-    type Iter: ParallelIterator<Item=Self::Item>;
+    type Iter: ParallelIterator<Item = Self::Item>;
     type Item: Send + 'data;
 
     fn par_iter(&'data self) -> Self::Iter;
@@ -85,7 +84,7 @@ impl<'data, I: 'data + ?Sized> IntoParallelRefIterator<'data> for I
 }
 
 pub trait IntoParallelRefMutIterator<'data> {
-    type Iter: ParallelIterator<Item=Self::Item>;
+    type Iter: ParallelIterator<Item = Self::Item>;
     type Item: Send + 'data;
 
     fn par_iter_mut(&'data mut self) -> Self::Iter;
@@ -103,7 +102,7 @@ impl<'data, I: 'data + ?Sized> IntoParallelRefMutIterator<'data> for I
 }
 
 pub trait ToParallelChunks<'data> {
-    type Iter: ParallelIterator<Item=&'data [Self::Item]>;
+    type Iter: ParallelIterator<Item = &'data [Self::Item]>;
     type Item: Sync + 'data;
 
     /// Returns a parallel iterator over at most `size` elements of
@@ -118,7 +117,7 @@ pub trait ToParallelChunks<'data> {
 }
 
 pub trait ToParallelChunksMut<'data> {
-    type Iter: ParallelIterator<Item=&'data mut [Self::Item]>;
+    type Iter: ParallelIterator<Item = &'data mut [Self::Item]>;
     type Item: Send + 'data;
 
     /// Returns a parallel iterator over at most `size` elements of
@@ -172,7 +171,7 @@ pub trait ParallelIterator: Sized {
 
     /// Applies `map_op` to each item of this iterator, producing a new
     /// iterator with the results.
-    fn map<MAP_OP,R>(self, map_op: MAP_OP) -> Map<Self, MapFn<MAP_OP>>
+    fn map<MAP_OP, R>(self, map_op: MAP_OP) -> Map<Self, MapFn<MAP_OP>>
         where MAP_OP: Fn(Self::Item) -> R + Sync
     {
         Map::new(self, MapFn(map_op))
@@ -181,7 +180,8 @@ pub trait ParallelIterator: Sized {
     /// Creates an iterator which clones all of its elements.  This may be
     /// useful when you have an iterator over `&T`, but you need `T`.
     fn cloned<'a, T>(self) -> Map<Self, MapCloned>
-        where T: 'a + Clone, Self: ParallelIterator<Item=&'a T>
+        where T: 'a + Clone,
+              Self: ParallelIterator<Item = &'a T>
     {
         Map::new(self, MapCloned)
     }
@@ -205,7 +205,7 @@ pub trait ParallelIterator: Sized {
 
     /// Applies `filter_op` to each item of this iterator to get an `Option`,
     /// producing a new iterator with only the items from `Some` results.
-    fn filter_map<FILTER_OP,R>(self, filter_op: FILTER_OP) -> FilterMap<Self, FILTER_OP>
+    fn filter_map<FILTER_OP, R>(self, filter_op: FILTER_OP) -> FilterMap<Self, FILTER_OP>
         where FILTER_OP: Fn(Self::Item) -> Option<R> + Sync
     {
         FilterMap::new(self, filter_op)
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Looks good so far: one thing that's missing is tests =)

ChristopherDavenport:
This is a basic implementation where performance can be improved through early returns, however this should effectively satisfy the contracts.

src/par_iter/mod.rs
```diff
@@ -674,11 +671,114 @@ pub trait IndexedParallelIterator: ExactParallelIterator {
...
+    fn cmp<I>(self, other: I) -> Ordering
+        where I: IntoParallelIterator<Item = Self::Item>,
+              I::Iter: IndexedParallelIterator,
+              Self::Item: Ord
+    {
+        self.zip(other.into_par_iter())
+            .map(|(x, y)| Ord::cmp(&x, &y))
+            .reduce(|| Ordering::Equal,
+                    |cmp_a, cmp_b| if cmp_a != Ordering::Equal {
+                        cmp_b
```
nikomatsakis: I think this is backwards. You want `if cmp_a == Ordering::Equal { cmp_ b } else { cmp_a }`.

src/par_iter/mod.rs
```diff
@@ -674,11 +671,114 @@ pub trait IndexedParallelIterator: ExactParallelIterator {
...
+    /// another.
+    fn partial_cmp<I>(self, other: I) -> Option<Ordering>
+        where I: IntoParallelIterator,
+              I::Iter: IndexedParallelIterator,
+              Self::Item: PartialOrd<I::Item>
+    {
+        self.zip(other.into_par_iter())
+            .map(|(x, y)| PartialOrd::partial_cmp(&x, &y))
+            .reduce(|| Some(Ordering::Equal),
+                    |cmp_a, cmp_b| match (cmp_a, cmp_b) {
```
nikomatsakis: This is also not quite right. I think you want something like this:

```
match cmp_a {
    Some(Ordering::Equal) => cmp_b,
    Some(_) => cmp_a,
    None => None,
}
```

The idea is that once we find something which *is* ordered (and not equal), it doesn't matter what comes after. But if we find something unordered, same is true. 

Here is an example demonstrating this: http://play.integer32.com/?gist=87daf4489fb2ca744e1ea99abac0f443&version=stable

src/par_iter/test.rs
```diff
@@ -272,6 +272,144 @@ pub fn check_range_exact_and_bounded() {
...
+    a[0] = ::std::f32::NAN;
+
+    let result = a.partial_cmp(&b);
+
+    assert!(result == None);
+}
+
+#[test]
+pub fn check_eq() {
+    let a: Vec<usize> = (0..1024).collect();
```
nikomatsakis: These tests are actually not testing your code. You are building vectors and then comparing them sequentially. You need something like:

```rust
let a = (0..1024).into_par_iter(); // NB: not collect() :)
let b = (0..1024).into_par_iter();
let result = a.eq(b);
assert!(result);
```

However, my preference is to write the tests so that they compare par against seq iterators, rather than directly encoding the expected result:


```rust
assert_eq!((0..1024).into_par_iter().eq(0..1024), (0..1024).eq(0..1024));
```

src/par_iter/test.rs
```diff
@@ -272,6 +272,144 @@ pub fn check_range_exact_and_bounded() {
...
+    let a: Vec<usize> = (1..1024).collect();
+    let b: Vec<usize> = (0..1024).collect();
+    let c: Vec<usize> = (1..1024).collect();
+
+    let result_gt = a.ge(&b);
+    let result_eq = a.ge(&c);
+
+    assert!(result_gt);
+    assert!(result_eq);
+}
```
nikomatsakis: can you also add some tests involving floats (which are not fully ordered)? 

http://play.integer32.com/?gist=87daf4489fb2ca744e1ea99abac0f443&version=stable would be a good one, if we added the parallel version and compared them against one another.

It'd also be good to add a test with some randomly generated data, using a fixed seed. 

Something like this:

```rust
#[test]
pub fn cmp_and_partial_cmp() {
    use rand::{Rng, SeedableRng, XorShiftRng};

    let mut rng = XorShiftRng::from_seed([14159, 26535, 89793, 23846]);
    let a: Vec<i32> = rng.gen_iter().take(1024).collect();
    let b: Vec<i32> = rng.gen_iter().take(1024).collect();
    for i in 0..a.len() {
        assert_eq!(a[i..].par_iter().partial_cmp(b[i..]));
        assert_eq!(a[i..].iter().cmp(b[i..]);
    }
}
```

(a twist on [@utkarshkukreti's test](https://github.com/nikomatsakis/rayon/pull/159#issuecomment-261599932))

ChristopherDavenport:
I'll get on it right away. I know the rest of the crew that was working on this with me went onto vacation, so I'll see what I can get under control.

As an aside, thank you for the really useful feedback. I'm fairly new to rust and such constructive criticism is very rare and very appreciated as I learned a lot from your change suggestions.

src/par_iter/test.rs
ChristopherDavenport: I'll add tests both between pars and between par and seq iterators. That way you'll have tests for expected behavior internally as well as expected behavior as compared to the external standard of seq iterators.

ChristopherDavenport:
1.   Fixed cmp, it was indeed backwards.
2.   For partial_cmp the much simpler match semantics play out much easier and return correctly if an earlier value is greater.
3.   All tests have been switched to into_par_iter or par_iter where appropriate. Tests check for appropriate evaluation as set by this library, and with comparison to seq iterators, therefore if seq iterator implementation in some way changes this will be apparent, as local tests will pass but comparison tests will fail.
4.   Float vectors were analyzed like in the simple example suggested pulling both with an early NaN returning None, and after a difference not making a difference to the return. Randomly generated data using a fixed seed was implemented for both cmp and partial_cmp with a comparison to the sequential version showing that the comparisons are indeed equal. 

If there is anything more I need to do just let me know.

src/par_iter/test.rs
```diff
@@ -352,6 +367,21 @@ pub fn check_partial_cmp_to_seq() {
...
 
 #[test]
+pub fn check_partial_cmp_rng_to_seq() {
+    use rand::{Rng, SeedableRng, XorShiftRng};
+
+    let mut rng = XorShiftRng::from_seed([9346, 26355, 87943, 28346]);
+    let a: Vec<i32> = rng.gen_iter().take(1024).collect();
+    let b: Vec<i32> = rng.gen_iter().take(1024).collect();
+    for i in 0..a.len() {
+        let par_result = a[i..].par_iter().partial_cmp(b[i..].par_iter());
```
nikomatsakis: Just FYI, the second call to `par_iter` (and `iter`) is not necessary: `a[i..].par_iter().partial_cmp(&b[i..])` is equivalent. But this is also fine, no need to change it -- more explicit this way.

nikomatsakis:
I see you have some merge conflicts -- do you think you can rebase? I suspect part of the problem is that you seem to have run rustfmt, which caused a lot of miscellaneous unrelated diffs in the code. (Though I don't mind, since I've been wanting to run rustfmt regularly).

One thing you might try if you are going to rebase is to introduce a first commit where you run rustfmt -- **before** you make any of our changes. That will isolate the rustfmt diffs into one commit and also make merging/rebasing easier.

Let me know though: if you are not that familiar with git, or don't have time, I can also help out and merge the changes.

ChristopherDavenport:
I will run rustfmt on these files which has no issue and then issue a rebase of an intermediate commit. 

I can do one across the whole project but it seems to create an oddity in the implementation of Tree in scope/test.rs. So rather than include it as the baseline after I finish the changes I can issue a pr with the whole project run through the rustfmt blender, that way the commits have good separation of concerns.

nikomatsakis:
Thanks!

PR #164, opened 2016-11-20
Add ParallelString with par_chars

cuviper:
This is a first step toward useful parallel operations on strings.  It
works as an `UnindexedProducer` to split on UTF-8 character boundaries.

### DIFF
--- a/src/par_iter/string.rs
+++ b/src/par_iter/string.rs
@@ -1,0 +1,66 @@
+use super::*;
+use super::internal::*;
+use std::str::Chars;
+
+/// Test if a byte is the start of a UTF-8 character.
+/// (extracted from `str::is_char_boundary`)
+fn is_char_boundary(b: u8) -> bool {
+    // This is bit magic equivalent to: b < 128 || b >= 192
+    (b as i8) >= -0x40
+}
+
+
+impl<'a> ParallelString for &'a str {
+    type Chars = ParChars<'a>;
+
+    fn par_chars(self) -> Self::Chars {
+        ParChars {
+            chars: self
+        }
+    }
+}
+
+
+pub struct ParChars<'a> {
+    chars: &'a str
+}
+
+impl<'a> ParallelIterator for ParChars<'a> {
+    type Item = char;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+        where C: UnindexedConsumer<Self::Item>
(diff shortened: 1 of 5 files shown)


cuviper:
It might make more sense to move this into a new neighboring `par_string` module.

(I'm considering that for slices too, e.g. with the chunks methods on a `ParallelSlice` trait.)

src/par_iter/string.rs
```diff
@@ -0,0 +1,66 @@
...
+        // This is pessimistic, as we only *know* there are multiple characters
+        // when it's longer than Unicode's maximum UTF-8 length of 4.  There
+        // could be smaller characters, but it's ok not to split maximally.
+        self.chars.len() > 4
+    }
+
+    fn split(self) -> (Self, Self) {
+        let mid = self.chars.len() / 2;
+        let (left, right) = self.chars.as_bytes().split_at(mid);
+        let index = right.iter().cloned().position(is_char_boundary).map(|i| mid + i)
```
nikomatsakis: Nit: Seems like a good place for a comment.

nikomatsakis:
> It might make more sense to move this into a new neighboring par_string module.

Perhaps. In general, I'd like to revisit the existing modules. Or, at least, make them private and `pub use` where necessary. I'm not sure why I made them all (or most of them?) `pub`.

cuviper:
Added a comment as requested.

:+1: on nuking most of the `pub` from modules.

nikomatsakis:
very cool.

src/par_iter/string.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    fn split(self) -> (Self, Self) {
+        let mid = self.chars.len() / 2;
+
+        // We want to split near the midpoint, but we need to find an actual
+        // character boundary.  So we look at the raw bytes, first scanning
+        // forward from the midpoint for a boundary, then trying backward.
+        let (left, right) = self.chars.as_bytes().split_at(mid);
+        let index = right.iter().cloned().position(is_char_boundary).map(|i| mid + i)
```
nikomatsakis: PS, @cuviper, I was browsing the docs and I see that `str` does offer an `is_char_boundary(i)` function. If we wanted, we could use that -- though we'd have to restructure this to iterate over indices, I think. Not a big thing either way.

PR #166, opened 2016-11-20
Make FromParallelIterator look more like FromIterator

cuviper:
It's now parameterized on the item type, which allows collecting
different things, e.g. `String` (#164) might collect `char` or `&str`.

The `from_par_iter` method is parameterized on `IntoParallelIterator`,
which means `Vec` has to be generalized too, not just exact.  There's
always `collect_into`, and we can eventually do better when Rust gets
specialization.

### DIFF
--- a/src/par_iter/collect/mod.rs
+++ b/src/par_iter/collect/mod.rs
@@ -1,4 +1,6 @@
-use super::{ParallelIterator, ExactParallelIterator};
+use super::{ParallelIterator, ExactParallelIterator, IntoParallelIterator};
+use super::from_par_iter::FromParallelIterator;
+use std::collections::LinkedList;
 use std::isize;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
@@ -25,8 +27,7 @@ pub fn collect_into<PAR_ITER, T>(mut pi: PAR_ITER, v: &mut Vec<T>)
 /// *any* `ParallelIterator` here, and `CollectConsumer` has to also implement
 /// `UnindexedConsumer`.  That implementation panics `unreachable!` in case
 /// there's a bug where we actually do try to use this unindexed.
-#[doc(hidden)]
-pub fn special_collect_into<PAR_ITER, T>(pi: PAR_ITER, len: usize, v: &mut Vec<T>)
+fn special_collect_into<PAR_ITER, T>(pi: PAR_ITER, len: usize, v: &mut Vec<T>)
     where PAR_ITER: ParallelIterator<Item = T>,
           T: Send
 {
@@ -56,5 +57,43 @@ pub fn special_collect_into<PAR_ITER, T>(pi: PAR_ITER, len: usize, v: &mut Vec<T
                 len,
                 actual_writes);
         v.set_len(len);
+    }
+}
+
+/// Collect items from a parallel iterator into a freshly allocated vector.
+impl<T> FromParallelIterator<T> for Vec<T>
+    where T: Send
+{
+    fn from_par_iter<PAR_ITER>(par_iter: PAR_ITER) -> Self
+        where PAR_ITER: IntoParallelIterator<Item = T>
+    {
+        // See the vec_collect benchmarks in rayon-demo for different strategies.
+        let mut par_iter = par_iter.into_par_iter();
+        match par_iter.opt_len() {
+            Some(len) => {
+                // When Rust gets specialization, call `par_iter.collect_into()`
+                // for exact iterators.  Until then, `special_collect_into()` fakes
+                // the same thing on the promise that `opt_len()` is accurate.
+                let mut vec = vec![];
+                super::collect::special_collect_into(par_iter, len, &mut vec);
+                vec
+            }
+            None => {
+                // This works like `combine`, but `Vec::append` is more efficient than `extend`.
+                let list: LinkedList<_> = par_iter.fold(Vec::new, |mut vec, elem| {
+                        vec.push(elem);
+                        vec
+                    })
+                    .collect();
+
+                let len = list.iter().map(Vec::len).sum();
+                let start = Vec::with_capacity(len);
+                list.into_iter()
+                    .fold(start, |mut vec, mut sub| {
+                        vec.append(&mut sub);
+                        vec
+                    })
+            }
+        }
     }
 }
--- a/src/par_iter/collect/mod.rs
+++ b/src/par_iter/collect/mod.rs
@@ -1,16 +1,35 @@
-use super::ExactParallelIterator;
+use super::{ParallelIterator, ExactParallelIterator};
 use std::isize;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 mod consumer;
 use self::consumer::CollectConsumer;
 
+/// Collects the results of the exact iterator into the specified vector.
 pub fn collect_into<PAR_ITER, T>(mut pi: PAR_ITER, v: &mut Vec<T>)
     where PAR_ITER: ExactParallelIterator<Item = T>,
-          PAR_ITER: ExactParallelIterator,
           T: Send
 {
     let len = pi.len();
+    special_collect_into(pi, len, v);
+}
+
+/// Collects the results of the iterator into the specified vector.
+///
+/// Technically, this only works for `ExactParallelIterator`, but we're faking a
+/// bit of specialization here until Rust can do that natively.  Callers are
+/// using `opt_len` to find the length before calling this, and only exact
+/// iterators will return anything but `None` there.
+///
+/// Since the type system doesn't understand that contract, we have to allow
+/// *any* `ParallelIterator` here, and `CollectConsumer` has to also implement
+/// `UnindexedConsumer`.  That implementation panics `unreachable!` in case
+/// there's a bug where we actually do try to use this unindexed.
+#[doc(hidden)]
+pub fn special_collect_into<PAR_ITER, T>(pi: PAR_ITER, len: usize, v: &mut Vec<T>)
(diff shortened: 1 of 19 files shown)


nikomatsakis:
So, I like these changes but I am reluctant to lose the performance. We could wait till specialization but I'm nervous: most newcomers will try `collect()` (vs `collect_into()`) as it is the most familiar thing, and I'd prefer not to lose performance.

I've been thinking about some alternative schemes for a bit now. Unfortunately, most of them are pretty tedious. =) For example, I was considering adding a callback on parallel iterators that lets you effectively downcast. Something like:

```rust
par_iter.if_indexed(|par_iter| { // 
    ... // this code executes if indexed_par_iter is indexed
}, |par_iter| {
    ... // this code executes otherwise
});
```

Unfortunately, this requires using the same "closure-like" struct pattern as `with_producer()` and so it seems too tedious to do.

I also experimented with a variant on `drive` that took two *closures*, one of which returned an unindexed consumer, and one an indexed one, but that ran aground with `chain()`, which wants to be able to split closures and so forth.

My latest thought was that we could have a `fn opt_len() -> Option<usize>` sort of method (sort of like `size_hint`, and maybe we should just call it that for compatibility) to test dynamically. If a direct len is available, we could then use the exact vector consumer -- or a wrapper around it which panics if `split_off()` is called (since it shouldn't be, if `opt_len` returns `Some(_)`, though we'd have to make a few modifications to ensure that). Of course, if `opt_len` returns `None`, we'd use the system you have here.

Thoughts?

src/par_iter/from_par_iter.rs
```diff
@@ -1,21 +1,22 @@
...
+use super::{IntoParallelIterator, ParallelIterator};
 
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::hash::{BuildHasher, Hash};
 use std::collections::LinkedList;
 use std::collections::{BinaryHeap, VecDeque};
 
-pub trait FromParallelIterator<PAR_ITER> {
-    fn from_par_iter(par_iter: PAR_ITER) -> Self;
+pub trait FromParallelIterator<ITEM> where ITEM: Send {
```
nikomatsakis: I'm torn about moving `PAR_ITER` to the method. On the one hand, I like the idea that every `collect` can always work from any parallel iterator, and it's certainly true that this can be achieved with buffering. On the other hand, I like the idea that you can control -- for your data structure -- whether you want to expose a buffering mode. 

I guess on balance I like the parameter being on the method. I still suspect we will tinker with this trait a bit, particularly once we have specialization available. (I'd sort of like `.collect()` to ultimately just alias into `.collect_into()`, also.)

Which reminds me, another idea I had been toying with -- basically orthogonal to this PR -- was giving people some control over how `collect()` proceeded (at some point) by having newtype'd wrappers. e.g., for hashmap, you might want to do `let SpareHashMap(map) = ...collect()`, which would be a hint to use folded *hashmaps* (which are faster if there is a lot of overlap). Anyway.

cuviper:
> So, I like these changes but I am reluctant to lose the performance. We could wait till specialization but I'm nervous: most newcomers will try collect() (vs collect_into()) as it is the most familiar thing, and I'd prefer not to lose performance.

Yes, but -- Someone on gitter just yesterday was trying to use `collect` on an iterator with `flat_map`, and didn't know why that doesn't work.  This is what made me think we should really support the general case first, and worry about performance specialization later.  *Especially* since it's already possible to use `collect_into` if your iterator is exact and you find that this performance matters.

I don't love trying to kludge around this with fake specialization, but I think it would be OK if we can do it without too much boilerplate, and completely hidden from users, so we can drop it when true specialization is available.  The `opt_len` approach sounds ok -- maybe I'll play with that...

cuviper:
I rebased and expanded the `String` implementations, then added `opt_len` to optimize vector `collect` as suggested.  Some new benchmarks look very nice!

```
running 10 tests
test vec_collect::vec_i::with_collect                        ... bench:   4,965,465 ns/iter (+/- 5,228,925)
test vec_collect::vec_i::with_collect_into                   ... bench:   4,977,562 ns/iter (+/- 4,186,311)
test vec_collect::vec_i::with_collect_into_reused            ... bench:   3,489,303 ns/iter (+/- 4,658,608)
test vec_collect::vec_i::with_fold                           ... bench:  25,890,601 ns/iter (+/- 3,389,768)
test vec_collect::vec_i::with_linked_list_vec                ... bench:  12,575,224 ns/iter (+/- 5,478,715)
test vec_collect::vec_i::with_linked_list_vec_sized          ... bench:  12,378,784 ns/iter (+/- 6,763,186)
test vec_collect::vec_i_filtered::with_collect               ... bench:  12,281,242 ns/iter (+/- 6,643,686)
test vec_collect::vec_i_filtered::with_fold                  ... bench:  25,655,800 ns/iter (+/- 3,676,608)
test vec_collect::vec_i_filtered::with_linked_list_vec       ... bench:  11,916,667 ns/iter (+/- 5,835,823)
test vec_collect::vec_i_filtered::with_linked_list_vec_sized ... bench:  11,595,895 ns/iter (+/- 6,147,349)
```

I also tried this for other collections, gathering `Vec<_>` instead of `LinkedList<Vec<_>>` where possible, but it didn't help in any measurable way.  That's probably because those still have to copy into their final collection, and it doesn't matter much whether the intermediate is a giant block of memory or a relative handful of smaller chunks.

src/par_iter/mod.rs
```diff
@@ -609,18 +609,19 @@ pub trait ParallelIterator: Sized {
     fn drive_unindexed<C>(self, consumer: C) -> C::Result
         where C: UnindexedConsumer<Self::Item>;
 
+    /// Internal method used to fake specialization for indexed collect.
+    #[doc(hidden)]
+    fn opt_len(&mut self) -> Option<usize> { None }
```
nikomatsakis: Let's say a bit more here. It's not clear to me that this method needs to be hidden and internal, but it can't hurt to keep it that way to start:

"Returns the number of items produced by this iterator, if known statically. This can be used by consumers to trigger special fast paths. Therefore, if `Some(_)` is returned, this iterator must only use the (indexed) `Consumer` methods when driving a consumer, such as `split_at()`. Calling `UnindexedConsumer::split_off()` or other `UnindexedConsumer` methods -- or returning an inaccurate value -- may result in panics."

src/par_iter/mod.rs
```diff
@@ -609,18 +609,19 @@ pub trait ParallelIterator: Sized {
...
-    ///
-    /// You may also prefer to use `collect_into()`, which allows you
-    /// to reuse the vector's backing store rather than allocating a
-    /// fresh vector.
+    /// by this parallel iterator.
+    ///
+    /// You may prefer to use `collect_into()`, which allocates more
+    /// efficiently with precise knowledge of how many elements the
+    /// iterator contains, and even allows you to reuse an existing
+    /// vector's backing store rather than allocating a fresh vector.
```
nikomatsakis: Do you think we can make `collect_into()` work for any collection `C` where `C: FromParallelIterator`? I guess we'd have to rework the `FromParallelIterator` trait somewhat to do so. Let's leave it as is for now.

src/par_iter/collect/mod.rs
```diff
@@ -1,16 +1,35 @@
...
 use self::consumer::CollectConsumer;
 
+/// Collects the results of the exact iterator into the specified vector.
 pub fn collect_into<PAR_ITER,T>(mut pi: PAR_ITER, v: &mut Vec<T>)
     where PAR_ITER: ExactParallelIterator<Item=T>,
-          PAR_ITER: ExactParallelIterator,
           T: Send,
 {
     let len = pi.len();
+    special_collect_into(pi, len, v);
```
nikomatsakis: Instead of adding `special_collect_into` as a method, can we make it a free fn in the `collect` module and call *that*?

(In general, I'd prefer to have fewer `doc(hidden)` methods, not more.)

src/par_iter/mod.rs
cuviper: I made it hidden and internal so we could feel free to remove it later, when we're able to accomplish the same thing with specialization.  I'm not sure it's much use outside of that purpose.  Maybe we can write it as `size_hint` if you want to mirror `Iterator`, but I don't think we *need* that anywhere.

Anyway, your expanded comment is fine with me -- I'll take it verbatim if you don't mind.

src/par_iter/mod.rs
cuviper: Well, maybe.  We'd have to gain methods for `clear` and `reserve`, and some collections have no notion of buffers to reuse (`BTreeMap`/`Set`).  And most would still need an intermediate vector for the actual parallel gathering.  Hmm, maybe as a new `FromExactParallelIterator` trait?

cuviper:
Rebased to deal with rustfmt changes, and addressed the `opt_len` and `special_collect_into` requests.

mathstuf:
After merging this PR, I'm seeing errors in my Rayon-using code:

```
panicked at 'internal error: entered unreachable code: CollectConsumer must be indexed!', /home/boeckb/.cargo/git/checkouts/rayon-2af7db7e4550ecfd/e2d8593dd0d870d54b1c7cd92342560f830f035d/src/par_iter/collect/consumer.rs:94
```

Is there an API break or am I doing something not covered by the test suite?

mathstuf:
For the record, my rayon-using code is [here](https://gitlab.kitware.com/utils/rust-git-hooks/blob/master/src/run.rs#L77).

cuviper:
That's unexpected, but my hunch is to blame `chain` for driving unindexed.  Can you please open a bug?

PR #167, opened 2016-11-21
Run rustfmt Across Entire Project

ChristopherDavenport:
This causes a project with compilation as it converts the implementation of Tree iter function in src/scope/test.rs line 95 to a todo. However this should give us a baseline of how to proceed. 

I can easily rebase this moving forwards so no worries there. 

Now fixing this is as easy as copy pasting, however it evades the problem that people can't safely run rustfmt on the code without an unintended side-effect. So instead it seems like we need an implementation that can be converted successfully so that way people can keep their code formatted in the standardized way.

First commit should obviously fail tests until the problem solution has been formalized.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ Cargo.lock
 target
 *~
 TAGS
+*.bk
--- a/src/api.rs
+++ b/src/api.rs
@@ -23,11 +23,15 @@ pub enum InitError {
 impl fmt::Display for InitError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            InitError::NumberOfThreadsZero =>
-                write!(f, "The number of threads was set to zero but must be greater than zero."),
-            InitError::GlobalPoolAlreadyInitialized =>
-                write!(f, "The gobal thread pool has already been initialized with a different \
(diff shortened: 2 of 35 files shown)


nikomatsakis:
> This causes a project with compilation as it converts the implementation of Tree iter function in src/scope/test.rs line 95 to a todo. However this should give us a baseline of how to proceed.

Argh, oh yeah, I hate that. It's https://github.com/rust-lang-nursery/rustfmt/issues/1154, which is supposedly fixed, but I'm not sure if the fix has found its way into a public release yet.

ChristopherDavenport:
If thats the case I can implement as is and people would need to be aware of that with fmt until it makes it into the public release. But if it's on its way I see no reason to introduce an intermediary commit unless we were going to fix it directly. So I can just place the code as is so people have a closer baseline and know that they will need to git checkout THEIR_BRANCH /src/scope/test.rs after they run a fmt. Also, if you are going to encourage fmt it might also be a good idea to introduce *.bk into the .gitignore file.

Let me know what you'd like.

ChristopherDavenport:
Please don't merge this till after #152 if there is a painful difference between the older version of rustfmt and the new one I can redo this over the changes but to go the other direction would be much less pleasant.

nikomatsakis:
Thanks :)

PR #169, opened 2016-11-29
Use slices to simplify vector collect

cuviper:
There's a lot less fuss required about invariants in `CollectConsumer`
and `CollectFolder` if we just use `&mut [T]` for the target memory.
We can use standard `split_at_mut` and `into_iter`, instead of dealing
with our own pointer math and hoping we got mutable aliasing right.

### DIFF
--- a/src/par_iter/collect/consumer.rs
+++ b/src/par_iter/collect/consumer.rs
@@ -2,50 +2,39 @@ use super::super::len::*;
 use super::super::internal::*;
 use super::super::noop::*;
 use std::ptr;
+use std::slice;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
-pub struct CollectConsumer<'c, ITEM: Send> {
+pub struct CollectConsumer<'c, ITEM: Send + 'c> {
     /// Tracks how many items we successfully wrote. Used to guarantee
     /// safety in the face of panics or buggy parallel iterators.
     writes: &'c AtomicUsize,
 
-    target: *mut ITEM,
-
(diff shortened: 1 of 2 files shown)


src/par_iter/collect/consumer.rs
```diff
@@ -2,50 +2,39 @@ use super::super::len::*;
...
     /// Tracks how many items we successfully wrote. Used to guarantee
     /// safety in the face of panics or buggy parallel iterators.
     writes: &'c AtomicUsize,
 
-    target: *mut ITEM,
-
-    /// Invariant: we can safely write from `target` .. `target+len`
-    len: usize,
+    /// A slice covering the target memory, not yet initialized!
+    target: &'c mut [ITEM],
```
nikomatsakis: Hmm. I chose to use `*mut` to signal that the memory may not be initialized. However, given that there do not yet exist unsafe code guidelines that forbid this, and that I think any such guidelines would almost certainly have to accept that `&T` may point to uninitialized memory so long as you don't read from it (which would include storing to it without a `ptr::write`, it is probably ok). Then the question becomes whether indeed it is better to use `&mut[]` to track our invariants for us, but with the caveat that we must be aware of the initialization state. Probably it is.

cuviper: It also lets us drop the `isize::MAX` limit, though I doubt that was ever a practical problem, and we get to use whatever the standard library decides is the optimal way to walk through the slice.  My performance testing on `vec_collect` actually looked a little favorable here, but it varied enough from run to run that I didn't want to quote that.

PR #171, opened 2016-11-30
Document and test the minimum rustc explicitly

### LINKED ISSUE #145: Error compiling Rayon 0.5.0
```
/Users/Home/.cargo/registry/src/github.com-88ac128001ac3a9a/rayon-0.5.0/src/scope/test.rs:88:34: 88:38 error: expected identifier, found keyword `impl`
/Users/Home/.cargo/registry/src/github.com-88ac128001ac3a9a/rayon-0.5.0/src/scope/test.rs:88     pub fn iter<'s>(&'s self) -> impl Iterator<Item=&'s T> + 's
                                                                                                                              ^~~~
/Users/Home/.cargo/registry/src/github.com-88ac128001ac3a9a/rayon-0.5.0/src/scope/test.rs:88:39: 88:47 error: expected one of `!`, `(`, `::`, `<`, `where`, or `{`, found `Iterator`
/Users/Home/.cargo/registry/src/github.com-88ac128001ac3a9a/rayon-0.5.0/src/scope/test.rs:88     pub fn iter<'s>(&'s self) -> impl Iterator<Item=&'s T> + 's
```

cuviper:
Rayon currently requires 1.12.0 to allow building with `impl Trait`
syntax in some of the tests, as they must parse even in normal builds.

Closes #145.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: rust
 rust:
+  - 1.12.0
   - stable
   - nightly
 script:
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ part, adding calls to Rayon should not change how your programs works
 at all, in fact. However, if you operate on mutexes or atomic
 integers, please see the [notes on atomicity](#atomicity).
 
+Rayon currently requires `rustc 1.12.0` or greater.
+
 ### Contribution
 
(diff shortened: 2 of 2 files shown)


nikomatsakis:
OK, so do we agree on this general policy?

- We document our minimum rustc.
- We make modest efforts to avoid raising the minimum, but if there is a feature or syntax we want that is on stable, we will raise the minimum.
- However, raising the minimum is considered a breaking change.

I have mixed feelings about the last point. I guess it indisputably is a breaking change, but it would seem a shame to go to Rayon 2.0 just because we want to use some feature on stable. It would also seem a shame to not get to use the feature. But for now let's go with the "full disclosure" thing. We should think about the best way to advertise when new major versions of Rayon are available, too.

cuviper:
> - However, raising the minimum is considered a breaking change.
> 
> I have mixed feelings about the last point. I guess it indisputably is a breaking change, but it would seem a shame to go to Rayon 2.0 just because we want to use some feature on stable. It would also seem a shame to not get to use the feature.

I do consider it a breaking change, but there are reasonable people on both sides of that point, as seen in the closed rust-lang/rfcs#1619.  As you say, it's unfortunate to be stuck on an old feature set.

The main issue to me is that Cargo doesn't have a notion of rustc compatibility for a crate, so having it select the highest semver-compatible crate is only safe if rustc is manually treated as part of semver.  There are a few RFCs floating around on this point, so if Cargo learns to only select crate versions that are compatible with the local rustc, then you could bump Rayon's requirement without hurting older users.  (Assuming the minimum baseline has a Cargo with this ability.)

nikomatsakis:
Let's try that policy for now. Can always change. We're not even on 1.0 yet, so worrying about 2.0 seems premature. =)

PR #172, opened 2016-12-01
Refactor threadpool's inner workings

nikomatsakis:
This is groundwork for better handling of asynchronous tasks and futures integration. 

There are a few changes:

- We no longer track the local stack depth. Instead, we pop and use pointer comparisons to decide if we popped the right-hand-side task. This scales gracefully to unbalanced pushes/pops (which can occur with `spawn_async` and with blocking).
- The main loop is now using the same code as the steal loop.
- Individual worker threads will now go to sleep when stealing if new data doesn't seem to be showing up. This is somewhat tunable: you can configure thresholds of how long they will yield before attempting to sleep.

I did some benchmarking throughout. I should do more though as I've done various refactorings and cleanup to the core algorithms in preparing this PR. In any case, [here is a plot](https://people-mozilla.org/~nmatsakis/spindown.svg) charting the effect of these changes on two join microbenchmarks that I added, since I wanted to ensure that this core underlying API remained solid. The plot shows five runs of various key commits: the most important commit is the final one, which is the effect of all the commits together. (The hash doesn't *quite* match because I dropped an inconsequential commit in the process.) The first two are basically a baseline. (This chart was created with the [cargo-chrono tool](https://github.com/nikomatsakis/cargo-chrono) that I implemented just for this purpose...)

Some things to do before landing:

- [x] Double check on panic tests and panic propagation.
- [x] Some kind of termination test cases -- do we even have any?
- [x] Write up good docs for the sleep module.
- [x] Tighten up some of the atomic orderings.
- [x] Gather more benchmark data.


nikomatsakis:
I did more measurements using a variety of benchmarks. Mostly they tell the same story: overall this change is a bit faster. The one exception is quick-sort.

Here is a plot of running a select set of tests, 10 times each, for various notable commits:

https://people-mozilla.org/~nmatsakis/2016.12.02-rayon-pr172.svg

This is the same data, but using medians of the 10 runs (with error bars):

https://people-mozilla.org/~nmatsakis/2016.12.02-rayon-pr172-medians.svg

Finally, I re-ran the results with quick-sort and found that indeed the same shift appears:

https://people-mozilla.org/~nmatsakis/2016.12.02-rayon-pr172-quicksort-medians.svg

nikomatsakis:
Ok, found a rather large oversight -- somewhere in my rebase I lost a critical call to `tickle()` that would cause the workers to wake up when other workers pushed new jobs. Amazing that this didn't have a more dramatic effect on performance, really. I am now re-measuring the data in light of fixing this problem -- but at least for quicksort, the results are encouraging (the perf is back to where it was).

nikomatsakis:
Revised data plots look pretty good:

- [all points](https://people-mozilla.org/~nmatsakis/2016.12.03-rayon-pr172-all.svg)
- [medians](https://people-mozilla.org/~nmatsakis/2016.12.03-rayon-pr172-medians.svg)

nikomatsakis:
Running mergesort with more data doesn't seem to show any performance loss:

```
# this branch, two runs:
seq: sorted 250000000 ints: 27.862516 s
par: sorted 250000000 ints: 2.594257 s
speedup: 10.74x

seq: sorted 250000000 ints: 27.71634 s
par: sorted 250000000 ints: 2.6546683 s
speedup: 10.44x

# master, two runs:
seq: sorted 250000000 ints: 27.862516 s
par: sorted 250000000 ints: 2.594257 s
speedup: 10.74x

seq: sorted 250000000 ints: 27.71634 s
par: sorted 250000000 ints: 2.6546683 s
speedup: 10.44x
```

nikomatsakis:
I have tweaked the memory orderings and written up my thoughts on why they are what they are. I also did measurements with everything set to `SeqCst` and saw no particular impact on performance -- so, in the `sleep` module at least, it may be worth forgoing complex reasoning and just using `SeqCst` everywhere. But it seems good to think rigorously and try for better `Ordering` values where possible.

nikomatsakis:
Rebased. 

@emilio I'm curious if you can easily run some servo benchmarks here to see if there is any effect from this branch. I'd like to land this soon, in any case.

emilio:
I think I can, but not in the really short-term (I need to do a bunch of classwork this week, though I'll let you know if I manage to get some spare time to do this :P).

FWIW what I use to do is downloading a local copy of a big DOM tree (the [html single-page spec](https://html.spec.whatwg.org/) uses to work fine), and then profile Servo loading that page locally (network makes stuff less deterministic).

To profile you can use Servo's `-p <period>` option, though in my experience it's not really reliable, so I always end up falling back to `perf`. When measuring rayon's performance, you usually care about the `top_down_dom` function, or related ones in the style system.

nikomatsakis:
OK, I've settled on how I think we should handle the atomic orderings and written some docs. I also rebased and cleaned up the history. The main "to do" item now is to add some termination and panic propagation test cases. I changed some of the panic recovery behavior in this branch -- I would prefer to say "normalized" -- and we ought to test that. (For example, if you call `join(a, b)` and `a` panics, it used to be nondeterminstic whether `b` would execute. This is no longer the case: `b` will always execute.)

nikomatsakis:
I now added several tests around panic propagation and tried to document the behavior with some precision. =)

nikomatsakis:
OK, I added some thread-pool tests and did some final cleanup. I think this branch is ready to land.

nikomatsakis:
Here is [a plot of measurements for these changes][plot], showing before/after with the final rebased version, normalized to 100:

![plot][plot]

plot: https://people-mozilla.org/~nmatsakis/rayon-pr172.svg

There are only two regressions (the colors aren't working right, so you have to hover the mouse):

- merge-sort, ~8%
- parallel-find-last, ~5%

When I measured the merge-sort with a larger input, I found the regression disappeared. I can't really account for the 8% we see here in the `#[bench]` results. I'm also not sure why `parallel-find-missing` seems to be so much faster.

nikomatsakis:
Re-ran the measurements. Looks the same. Some charts:

- https://people-mozilla.org/~nmatsakis/rayon-pr172-1.svg (normalized medians)
- https://people-mozilla.org/~nmatsakis/rayon-pr172-1-points.svg (all data points)

PR #173, opened 2016-12-03
Reorder use super::* / super::internal::*

cuviper:
Nightly started warning that many `use super::internal::*` imports under
`par_iter` were unused -- see rust-lang/rust#38104.  It turns out that
all those modules also had `use super::*`, and `par_iter/mod.rs` has its
own `use internal::*`, so apparently this now adds to what the other
modules get.

But we can't just remove those lines, because it doesn't work for older
rustc.  Instead we can reorder it so `super::internal::*` comes first,
and nothing looks unused to the linter.

### DIFF
--- a/src/par_iter/chain.rs
+++ b/src/par_iter/chain.rs
@@ -1,5 +1,5 @@
-use super::*;
 use super::internal::*;
+use super::*;
 use std::cmp::min;
 use std::iter;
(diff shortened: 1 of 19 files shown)


nikomatsakis:
Ah, I was wondering what was up with those warnings.

PR #176, opened 2016-12-06
Remove the fake specialization on chain collect

### LINKED ISSUE #175: `.chain` ends up with an unindexed collector
The rayon-using code [here](https://gitlab.kitware.com/utils/rust-git-hooks/blob/bf4d8edd769cf491bf92df07ee0004d0931ab9e8/src/run.rs#L77) is causing this error after merging of #166:

```
panicked at 'internal error: entered unreachable code: CollectConsumer must be indexed!', /home/boeckb/.cargo/git/checkouts/rayon-2af7db7e4550ecfd/e2d8593dd0d870d54b1c7cd92342560f830f035d/src/par_iter/collect/consumer.rs:94
```

cuviper:
`ChainIter::drive_unindexed` can't deal with the fake specialization
used by `CollectConsumer`, as it has to directly call `split_off`.  So
we must not report an `opt_len`, and thus iterators with any chaining
can't take advantage of that optimization.

However, it would be a regression if we couldn't even use `collect_into`
with chains, so we need to let that call a direct indexed `drive` again.
So refactor the entry points with `collect_into` calling `drive`, and
still `special_collect_into` calling `drive_unindexed`, with all of
their common code in a new `Collect` wrapper.

Fixes #175.

### DIFF
--- a/src/par_iter/chain.rs
+++ b/src/par_iter/chain.rs
@@ -35,10 +35,15 @@ impl<A, B> ParallelIterator for ChainIter<A, B>
     }
 
     fn opt_len(&mut self) -> Option<usize> {
-        match (self.a.opt_len(), self.b.opt_len()) {
-            (Some(a_len), Some(b_len)) => a_len.checked_add(b_len),
-            _ => None,
-        }
+        // NB: Even though we could compute the indexed length as below,
+        // we can't support collect's faux `UnindexedConsumer` in our
+        // `drive_unindexed`, so we must leave this un-"specialized".
+        //
+        // match (self.a.opt_len(), self.b.opt_len()) {
+        //     (Some(a_len), Some(b_len)) => a_len.checked_add(b_len),
+        //     _ => None,
+        // }
+        None
     }
 }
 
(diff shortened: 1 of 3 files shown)


mathstuf:
Fixes #175 for me. What optimization is being missed out on here?

cuviper:
For `collect_into` on indexed iterators, we can write directly to the target vector in the appropriate positions.  For `collect` with any arbitrary parallel iterator, which may be unindexed, we instead fold into intermediate vectors in each parallel job, then aggregate them all into a final vector, which means there's extra allocation and copying involved.

Ideally, we would use specialization to let `collect` use `collect_into`'s strategy when given an indexed iterator, but specialization in Rust isn't stable yet.  Instead, we figured out a hack with `opt_len` and this `special_collect_into` to approximate the same effect.  This is sort of lying to the type system, and you found a case in `chain` where this can't work, so my solution here is to just disable that on `chain`.

So the loss is that `collect` on iterators with any `chain` will have go through intermediate vectors.  If you can structure your code to use `collect_into`, *proving* that it's indeed indexed, then you can still get the optimized way.  Once we get true Rust specialization, we'll be able to do this transparently.

mathstuf:
Ah, sounds good. Since I basically have just some `Vec<String>` in a structure (inserting into the vector should just be a handful of pointer copies) and each parallel job involves (usually) forking external processes, I suspect I'm still getting a net benefit from rayon here. Specialization will just be gravy on top :) .

In any case, my fold later is about merging those string vectors together anyways.

Thanks for the explanation.

nikomatsakis:
Seems ok for now.

PR #183, opened 2016-12-19
Add #[inline] on the primitive numeric type impls

### LINKED ISSUE #181: Need #[inline] on reduce operations
A bunch of trivial non-generic functions here are non inlinable: https://github.com/nikomatsakis/rayon/blob/master/src/par_iter/reduce.rs

bluss:
A small improvement that makes it possible for the sum loops to optimize.

(Note that floating point reductions are not possible to autovectorize, but integers are.)

Fixes #181

### DIFF
--- a/src/par_iter/reduce.rs
+++ b/src/par_iter/reduce.rs
@@ -131,9 +131,11 @@ pub const SUM: &'static SumOp = &SumOp;
 macro_rules! sum_rule {
     ($i:ty, $z:expr) => {
         impl ReduceOp<$i> for SumOp {
+            #[inline]
             fn start_value(&self) -> $i {
                 $z
             }
+            #[inline]
             fn reduce(&self, value1: $i, value2: $i) -> $i {
                 value1 + value2
             }
@@ -161,9 +163,11 @@ pub const PRODUCT: &'static ProductOp = &ProductOp;
 macro_rules! product_rule {
     ($i:ty, $z:expr) => {
         impl ReduceOp<$i> for ProductOp {
+            #[inline]
             fn start_value(&self) -> $i {
                 $z
             }
+            #[inline]
             fn reduce(&self, value1: $i, value2: $i) -> $i {
                 value1 * value2
             }


cuviper:
Looks good to me!

> Note that floating point reductions are not possible to autovectorize

Is it possible to relax those rules?  Like a `gcc -ffast-math` sort of thing?  We're already giving precision by assuming associativity for parallel splits, so some local shuffling would be just another drop in the bucket.

bluss:
Yes it is, there's only intrinsics for it (fadd_fast, fmul_fast), so it's an unstable feature so far.

bluss:
Doc: http://llvm.org/docs/LangRef.html#fast-math-flags

It's a bit fuzzy and scary unfortunately. Things we want like arbitrary operand reordering is tangled up with (the reserved right to) ignore nans.

bluss:
IME (the experience is very limited, it's the code below), NaN are propagated from the input into the output

```rust
#![crate_type="lib"]
#![feature(core_intrinsics)]

use std::intrinsics::{fadd_fast as fadd, fmul_fast as fmul};

#[no_mangle]
pub fn dot3_fast(x: &[f32], y: &[f32], z: &[f32]) -> f32 {
    let mut sum = 0.;
    for (&a, (&b, &c)) in x.iter().zip(y.iter().zip(z)) {
        unsafe {
            sum = fadd(sum, fadd(fmul(a, a), fadd(fmul(b, b), fmul(c, c))))
        }
    }
    sum
}

#[test]
fn test_nan_in_is_nan_out() {
    let xs = [1.; 128];

    let dot = dot3_fast(&xs, &xs, &xs);
    assert_eq!(dot, 128. * 3.);

    let mut xs = [[1.1; 128]; 3];
    let nan = 0./0.;

    let mut fail = Vec::new();
    for i in 0..xs.len() {
        for j in 0..3 {
            xs[i][j] = nan;
            let dot = dot3_fast(&xs[0], &xs[1], &xs[2]);
            if !dot.is_nan() {
                fail.push((i, j));
            }
            xs[i][j] = 1.1;
        }
    }
    assert_eq!(&fail, &[]);
}
```

Issue #111, opened 2016-10-19
deprecate existing weight APIs


nikomatsakis:
Per [this comment](https://github.com/nikomatsakis/rayon/pull/106#issuecomment-254469425), we should deprecate the existing weight APIs. I think we probably still want some room for manual tuning, but it should be simpler by default -- and if we are going to let people give fine-grained numbers, it should be tied to the length (like seq. cutoff) rather than some abstract notion of weight. Ideally though we'd add these APIs **along with representative benchmarks** where they are needed.

Eh2406:
As a quick test I took a work project and cargo updated to 0.4.3. With `weight_max` it runes 32s. Then I removed the `weight_max` and it runs in 48s. (each best of 3 runs, wall clock time.) Did I test this correctly? Is there any info I can get so as to be helpful?  Unfortunately I can't share the example.  :-(

cuviper:
@Eh2406 can you characterize the workload somehow?  Is the work well balanced for each item, or skewed heavily?  Maybe you can synthesize something that performs similarly?

For possible tunables related to length, I think we'll want both a minimum (don't bother splitting below this length) and a maximum (always split at least down to this length), where `weight_max()` corresponds to asking for a maximum length 1.

Eh2406:
so I am running on a `intel core i5-5200u @ 2.2GHz 2.19 GHz`
I Isolated the first stage, the stages are run sequentially but each one calls into_par_iter.
The first stage, actually looks good.
with `weight_max` 61s, 56s, 51s
without 46s, 58s, 59s
So let me try to isolate the second stage, as so far the new algorithm seems comparable.

fkjogu:
I really dislike the `weights` mechanism, since it relies on an arbitrary threshold. Which makes it meaningless as a measure, since the expensiveness of task can also vary arbitrarily. So I'm in favor removing it.

However, I'd like to retain some control, how the work is split. So I'd propose to simply introduce an optional way, to specify the number of threads used.

Something like `.par_iter().jobs(42).whatever()`.

It is minimally intrusive, but offers precise manual control, if needed.

nikomatsakis:
I think @cuviper has some thoughts here and wanted to take a crack at this. I'm going to assign to them and remove the Help-Wanted label for now.

PR #184, opened 2016-12-19
Add method fold_with to Producer and UnindexedProducer

bluss:
This method allows the producers to customize the inner loop. This is
useful for iterators that can provide a more efficient inner loop than
the default for loop.

Example: multidimensional arrays.

### DIFF
--- a/src/par_iter/internal.rs
+++ b/src/par_iter/internal.rs
@@ -28,6 +28,22 @@ pub trait Producer: IntoIterator + Send + Sized {
     /// Split into two producers; one produces items `0..index`, the
     /// other `index..N`. Index must be less than `N`.
     fn split_at(self, index: usize) -> (Self, Self);
+
+    /// Iterate the producer, feeding each element to `folder`, and
+    /// stop when the folder is full (or all elements have been consumed).
+    ///
+    /// The provided implementation is sufficient for most iterables.
+    fn fold_with<F>(self, mut folder: F) -> F
+        where F: Folder<Self::Item>,
+    {
+        for item in self {
+            folder = folder.consume(item);
+            if folder.full() {
+                break;
+            }
+        }
+        folder
+    }
 }
 
 /// A consumer which consumes items that are fed to it.
@@ -95,6 +111,22 @@ pub trait UnindexedConsumer<ITEM>: Consumer<ITEM> {
 pub trait UnindexedProducer: IntoIterator + Send + Sized {
     fn can_split(&self) -> bool;
     fn split(self) -> (Self, Self);
+
+    /// Iterate the producer, feeding each element to `folder`, and
+    /// stop when the folder is full (or all elements have been consumed).
+    ///
+    /// The provided implementation is sufficient for most iterables.
+    fn fold_with<F>(self, mut folder: F) -> F
+        where F: Folder<Self::Item>,
+    {
+        for item in self {
+            folder = folder.consume(item);
+            if folder.full() {
+                break;
+            }
+        }
+        folder
+    }
 }
 
 /// A splitter controls the policy for splitting into smaller work items.
@@ -207,14 +239,7 @@ pub fn bridge_producer_consumer<P, C>(len: usize, mut producer: P, mut consumer:
                      || helper(len - mid, splitter, right_producer, right_consumer));
             reducer.reduce(left_result, right_result)
         } else {
-            let mut folder = consumer.into_folder();
-            for item in producer {
-                folder = folder.consume(item);
-                if folder.full() {
-                    break;
-                }
-            }
-            folder.complete()
+            producer.fold_with(consumer.into_folder()).complete()
         }
     }
 }
@@ -245,13 +270,6 @@ fn bridge_unindexed_producer_consumer<P, C>(mut splitter: Splitter,
                  || bridge_unindexed_producer_consumer(splitter, right_producer, right_consumer));
         reducer.reduce(left_result, right_result)
     } else {
-        let mut folder = consumer.into_folder();
-        for item in producer {
-            folder = folder.consume(item);
-            if folder.full() {
-                break;
-            }
-        }
-        folder.complete()
+        producer.fold_with(consumer.into_folder()).complete()
     }
 }
--- a/src/par_iter/chain.rs
+++ b/src/par_iter/chain.rs
@@ -185,6 +185,17 @@ impl<A, B> Producer for ChainProducer<A, B>
              ChainProducer::new(0, a_right, b_right))
         }
     }
+
+    fn fold_with<F>(self, mut folder: F) -> F
+        where F: Folder<A::Item>,
+    {
+        folder = self.a.fold_with(folder);
+        if folder.full() {
+            folder
+        } else {
+            self.b.fold_with(folder)
+        }
+    }
 }
 
 impl<A, B> IntoIterator for ChainProducer<A, B>


bluss:
This allows parallel producers to customize the inner loop. This is a temporary measure until iterators gain this natively (one [proposal is the fold_ok method](https://internals.rust-lang.org/t/pre-rfc-fold-ok-is-composable-internal-iteration/4434)).

This is marked WIP because I haven't been able to use this on UnidexedProducer parallel iterators to make a real difference.

This is because for very lightweight operations, I need to set the weight very low; but the unindexed producer has no way of communicating the weight to the splitter (?), so it always splits the producer down to the limit used by `fn can_split()`.

I do have success using ParallelIterator with UnindexedProducer if I do something heavy enough. For example: if the lower limit for can_split is 128 elements and the operation is `f64::exp` it's a win.

Replacing that expensive function call with something that autovectorizes has big gains:

```rust
fn fastexp(x: f64) -> f64 {
    let x = 1. + x/1024.;
    x.powi(1024)
}
```

However, only if the can split threshold is hardcoded to a larger limit. (Something like at least 10 000 elements at least, it turns out, for “fastexp”).

bluss:
As an example of potential gains, the PR enables fastexp to be autovectorized in ndarray (with the parallel traits implemented, not yet integrated into ndarray), for a case like this:

```rust
    let mut a = Array2::<f64>::zeros((M, N));
    let mut a = a.slice_mut(s![.., ..-1]);
    a.view_mut().into_par_iter().for_each(|x| *x = fastexp(*x));
```

The inner loop customization is needed because the array is not contiguous (due to the slicing), yet each row is contiguous, so the inner loop can still be subject to loop optimizations if we circumvent the general `Iterator::next`.

`.weight()` has no effect here it seems? I think a way to impact the unindexedproducer's splitting is needed.

The WIP parallel iterator implementation is here: https://github.com/bluss/rust-ndarray/compare/par-iter...par-iter-elements?expand=1

src/par_iter/internal.rs
```diff
@@ -28,6 +28,20 @@ pub trait Producer: IntoIterator + Send + Sized {
...
+    fn fold_using<F>(self, mut folder: F) -> F::Result
+        where F: Folder<Self::Item>,
+    {
+        for item in self {
+            folder = folder.consume(item);
+            if folder.full() {
+                break;
+            }
+        }
+        folder.complete()
```
cuviper: I think it may be more useful to leave `complete` out of this.  At least for `ChainProducer`, that would let it call `fold_using` separately on its two parts, recursively if there are multiple levels of chains, then just let the bridge functions complete it at the end.

cuviper:
I might bikeshed this to something like `fold_with`, not sure, but anyway I like the idea.  It's also interesting that the `full` concept could be applied to the short-circuiting `fold_ok` -- very nice!

As for tuning parameters, note that we're trying to move away from `weight` -- see #111.  This is why I didn't bother adding the `cost` function to the new `UnindexedProducer` at all.  But we probably do want *some* kind of tunables, and I think a minimum split size would address what you're wanting here.  My rough idea is that we need both a minimum (jobs should be at least this big) and a maximum (jobs should be no bigger, e.g. used to fit the working size in cache).

> so it always splits the producer down to the limit used by fn can_split().

This shouldn't *always* happen unless your workload is very small.  The "thief" splitter works by first splitting roughly to the number of job threads, and then each time a job is stolen that gets split to the number of threads again.  Towards the end of task, there probably will be a lot of eager stealing and thus splitting, maybe down to the minimum, but for most of a busy workload they should stay relatively large.

Automatic tuning is hard -- if you have input on manual tuning parameters, please chime in!

src/par_iter/internal.rs
bluss: Neat. I know that idea :D https://github.com/rust-lang/rust/pull/37315

cuviper: I knew I had seen it somewhere! :smile:

bluss:
One important piece of tuning turns out to be (maybe not surprisingly) using an explicit configuration to use one thread per core (so not counting hyperthreading).

bluss:
Sure. I'll squash this. I removed the WIP tag since it is an improvement when using a thread per core.

bluss:
I'll squash when you think it's ready, I mean.

src/par_iter/internal.rs
```diff
@@ -95,6 +111,22 @@ pub trait UnindexedConsumer<ITEM>: Consumer<ITEM> {
...
+
+    /// Iterate the producer, feeding each element to `folder`, and
+    /// stop when the folder is full (or all elements have been consumed).
+    ///
+    /// The provided implementation is sufficient for most iterables.
+    fn fold_with<F>(self, mut folder: F) -> F
+        where F: Folder<Self::Item>,
+    {
+        for item in self {
+            folder = folder.consume(item);
```
nikomatsakis: At some point, we discussed whether we should have `Producer: UnindexedProducer`, but we didn't have a good reason for it (even though it logically makes sense). The existence of this duplicate code might provide such a reason.

bluss:
Squashed commits together.

src/par_iter/internal.rs
bluss: I'm new to the internals, maybe split_at can be removed and only split is left?

bridge_producer_consumer is this today:

```rust
            let mid = len / 2;
            let (left_producer, right_producer) = producer.split_at(mid);
            let (left_consumer, right_consumer, reducer) = consumer.split_at(mid);
            let (left_result, right_result) =
                join(|| helper(mid, splitter, left_producer, left_consumer),
                     || helper(len - mid, splitter, right_producer, right_consumer));
            reducer.reduce(left_result, right_result)
```

It could change to something like:

```
            let (left_producer, right_producer) = producer.split();
            let mid = left_producer.len();
            let (left_consumer, right_consumer, reducer) = consumer.split_at(mid);
```

so the difference is that an indexed producer can tell its length (and that it has a meaningful length or before / after ordering of the halves).

ndarray's unindexed producer can tell the length of each part perfectly. What sets it apart is that (1) it can't split exactly at len / 2 since it must split in chunks of some axis' size (2) It doesn't want to split so that it has a "before" and "after" half. It prefers an unordered split for performance.

bluss: About order:

Take a 3 × 4 array

```rust
[[1, 2, 3, 4],
 [5, 6, 7, 8],
 [9, 0, 1, 2]];
```

The logical order of the regular `.iter()` is defined to be "logical" row major, which means the elements in the order we read them: 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2. The memory order of the array is separated from the logical order of `.iter()`.

For parallelism, we guarantee no order and the unindexed producer for an array view will pick either of these two splits, depending on which preserves memory locality:

```rust
[[1, 2, | 3, 4],
 - - - - - - - - - v.split_at(Axis(0), 1) // best for row major memory order
 [5, 6, | 7, 8],
 [9, 0, | 1, 2]];
        \ v.split_at(Axis(1), 2)  // best for column major memory order
```

nikomatsakis: No, you need `split_at` in order to handle zip.

nikomatsakis: (That said, I did want to consider changing the structure so that the producer can offer a suggestion about where to split instead. But I haven't decided on the best way to do it.)

nikomatsakis: So the key distinctions are:

- can split at an arbitrary point
    - needed for `zip`
- knows exactly how many elements it will produce
    - needed for precise consumers

Today we tie those together, but they could be separated. Something like:

- UnindexedProducer
    - can split itself at some point
- KnownLengthProducer: UnindexedProducer
    - can tell you the number of items it will produce
- Producer: KnownLengthProducer
    - can further be split at an arbitrary point

we might then have two bridge methods:

- one takes a `KnownLengthProducer` and a `Consumer`
    - this is the one that today requires a `Producer`
    - it would instead invoke `producer.split()`, get resulting length, and invoke `consumer.split_at()`, as you suggested
- one takes a `UnindexedProducer` and a `UnindexedConsumer`
    - same as today

We would probably have to tweak the `ParallelIterator` hierarchy a bit too, but the idea is that `zip()` would require that its sub-iterators can produce a full `Producer` (or, potentially, that the LHS can produce a `KnownLengthProducer` and the RHS can produce a full `Producer`). It's `split()` implementation can then call `left.split()` and right.split_at(left_split.len())`.

Make sense?

cuviper: > At some point, we discussed whether we should have Producer: UnindexedProducer, but we didn't have a good reason for it (even though it logically makes sense). The existence of this duplicate code might provide such a reason.

Yes, although IIRC we were also wanting `impl<T: Producer> UnindexedProducer for T` to avoid duplicating `can_split`/`split` boilerplate, but that would really want specialization.  Plus that boilerplate would never actually be invoked.

In the case of just `fold_while`, perhaps we could make a `BaseProducer` common to both, so the boilerplate is a one-line empty `impl` in most cases.  Or we could extract this body to a shared free function just to avoid the duplication.  But this is only 6 lines, let's not overthink it. :)

cuviper: Yet another option is to add something like:
```rust
trait Folder {
    // ...
    fn consume_iter<I:IntoIterator>(self, iterator: I) -> Self {
        // ...
    }
}
```
... which might be useful to folks specializing `fold_with` too.

nikomatsakis:
OK, I propose we merge this as is, but experiment with the producer hierarchy a bit.

bluss:
Will this be in a point release, like 0.6.1 ?

nikomatsakis:
@bluss presumably

bluss:
Rayon is making big strides anyway, and that's exciting.

nikomatsakis:
@bluss I'll push a new release ASAP, no reason to wait I guess

PR #185, opened 2016-12-20
Scoped TLS API

bholley:
Per discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1323372

We need either this or #186, but not both.

CC @emilio

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -2,6 +2,7 @@ use latch::{LockLatch, SpinLatch};
 #[allow(unused_imports)]
 use log::Event::*;
 use job::StackJob;
+use std::cell::{Ref, RefCell, RefMut};
 use std::sync::Arc;
 use std::error::Error;
 use std::fmt;
@@ -239,5 +240,45 @@ impl ThreadPool {
 impl Drop for ThreadPool {
     fn drop(&mut self) {
         self.registry.terminate();
+    }
+}
+
+/// Stack-scoped thread-local storage for thread pools.
+pub struct ScopedTLS<'a, T: Send> {
+    _pool: &'a ThreadPool,
+    slots: Box<[RefCell<Option<T>>]>,
+}
+
+unsafe impl<'a, T: Send> Sync for ScopedTLS<'a, T> {}
+
+impl<'a, T: Send> ScopedTLS<'a, T> {
+    pub fn new(p: &'a ThreadPool) -> Self {
+        let mut v = Vec::new();
+        for _ in 0..p.registry.num_threads() {
+            v.push(RefCell::new(None));
+        }
+
+        ScopedTLS {
+            _pool: p,
+            slots: v.into_boxed_slice(),
+        }
+    }
+
+    pub fn borrow(&self) -> Ref<Option<T>> {
+        let idx = current_thread_index();
+        self.slots[idx].borrow()
+    }
+
+    pub fn borrow_mut(&self) -> RefMut<Option<T>> {
+        let idx = current_thread_index();
+        self.slots[idx].borrow_mut()
+    }
+}
+
+fn current_thread_index() -> usize {
+    unsafe {
+        let curr = WorkerThread::current();
+        assert!(!curr.is_null());
+        (*curr).index()
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,7 @@ pub use api::InitError;
 pub use api::dump_stats;
 pub use api::initialize;
 pub use api::join;
+pub use api::ScopedTLS;
 pub use api::ThreadPool;
 #[cfg(feature = "unstable")]
 pub use scope::{scope, Scope};


bholley:
Use-case is in https://github.com/servo/servo/pull/14642

src/api.rs
```diff
@@ -241,3 +242,43 @@ impl Drop for ThreadPool {
...
+
+    pub fn borrow_mut(&self) -> RefMut<Option<T>> {
+        let idx = current_thread_index();
+        self.slots[idx].borrow_mut()
+    }
+}
+
+fn current_thread_index() -> usize {
+    unsafe {
+        let curr = WorkerThread::current();
```
nikomatsakis: One catch is that we don't know the worker-thread is associated with the thread-pool that the `ScopedTLS` corresponds to. My inclination would be to use the machinery I introduced in https://github.com/nikomatsakis/rayon/pull/141 (which I should probably land at some point...), which basically maintains a linked list of registries (which is expected to be of length 1, really) to find the appropriate one. In that case, you could just create the `ScopedTLS` without giving a `ThreadPool` (since the common case is not to have one, but instead to use the global, default thread-pool).

But another option would be to assert in this method that the worker belongs to the given thread-pool (and perhaps modify `new` so that it takes no arguments, in which cases it uses the default registry, which would either be the global registry or -- if in a worker -- the current thread-pool).

nikomatsakis:
OK, hmm. I think I'm inclined to go with a tweaked form of #186 for the time being, though longer term I might like to have something like this TLS API available. Basically because #186 doesn't seem to be committing to very much and is simpler to add in the short term. But I'm curious to hear what @cuviper has to say.

nikomatsakis:
@cuviper, for context:

- Servo project has a need to do some caching. They want this to be per-worker-thread. One option is that they expose the current worker thread index as well as the number of threads. I am slightly nervous about this since I imagine that in the future we may want to make the number of threads more dynamic, but on the other hand it's a pretty minor bit of information and may be of general utility.
- Alternatively, we could offer the TLS API ourselves, which also seems like a useful thing to have around (and then we could adjust the implementation to cope with a growing number of threads if that were ever a possibility). But I worry that to "do it right" would require a bit of iteration (though I guess we can certainly flag the API with the unstable feature flag as a warning).

src/api.rs
```diff
@@ -241,3 +242,43 @@ impl Drop for ThreadPool {
...
+pub struct ScopedTLS<'a, T: Send> {
+    _pool: &'a ThreadPool,
+    slots: Box<[RefCell<Option<T>>]>,
+}
+
+unsafe impl<'a, T: Send> Sync for ScopedTLS<'a, T> {}
+
+impl<'a, T: Send> ScopedTLS<'a, T> {
+    pub fn new(p: &'a ThreadPool) -> Self {
+        let mut v = Vec::new();
```
Manishearth: `Vec::with_capacity` please

cuviper:
> One option is that they expose the current worker thread index as well as the number of threads. I am slightly nervous about this since I imagine that in the future we may want to make the number of threads more dynamic, but on the other hand it's a pretty minor bit of information and may be of general utility.

I'm OK with this, and I think we can just document the caveat that this information may become stale if dynamic threads are used.  And in the future, dynamic threads should probably have an opt-out, or even be opt-in only.

> Alternatively, we could offer the TLS API ourselves, which also seems like a useful thing to have around (and then we could adjust the implementation to cope with a growing number of threads if that were ever a possibility). But I worry that to "do it right" would require a bit of iteration (though I guess we can certainly flag the API with the unstable feature flag as a warning).

It seems like something that could be neat, but I agree we'd need more deliberation here, and it sounds like they're a bit eager.  Going with #186 for now doesn't preclude a TLS API later, especially once they've proven its worth in Servo code.  (I'm assuming they'd use the index to do the same thing on their own.)

Was the [`scoped-tls`](https://crates.io/crates/scoped-tls) crate considered?  This may not need direct Rayon help at all...

nikomatsakis:
@cuviper I think that crate won't suffice because the cache should persist across Rayon jobs -- i.e., they are using the scoped API -- but I'm not 100% sure. Anyway, I agree, let's go with #186 for now. Going to close this and move the conversation over there, I think I had a few tweaks I was going to suggest to that PR.

bholley:
@cuviper |scoped-tls| requires the type to be static, right? That won't work for us, because we have a type with non-static lifetime.

cuviper:
@bholley I don't think so, but I have no personal experience with it.  It appears to only use `&'static` on the `ScopedKey`, and the docs say, "There are no restrictions on what types can be placed into a scoped variable".

bholley:
Oh, I see. I think I misread it.

Niko's comment above still applies though, since the TLS needs to be set up by the parent thread in order to persist across jobs.

nikomatsakis:
@bholley I guess the key question is whether one job needs to be able to insert a cache entry that other, sibling jobs will be able to use (and I think they do, no?)

bholley:
By 'sibling jobs' do you mean subsequently-spawned jobs that get scheduled on the same thread? If so, then yes, that's the reason for using TLS here (otherwise we'd just pass it around explicitly).

cuviper:
As long as the jobs nested, either through explicit recursion or by stealing jobs while waiting for some completion, then I think `scoped-tls` would do the job.  But if a worker thread completed its queue entirely, then its scope would be ended, probably bouncing in and out as it steals more jobs and finishes them. So... probably not what you want.

PR #186, opened 2016-12-20
Expose the number of threads in the pool and the index of the current thread

bholley:
Per discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1323372

We need either this or #185, but not both.

CC @emilio

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -234,6 +234,21 @@ impl ThreadPool {
             job_a.into_result()
         }
     }
+
+    /// Returns the number of threads in the thread pool.
+    pub fn num_threads(&self) -> usize {
+        self.registry.num_threads()
+    }
+
+    /// Returns the index of the current thread in the thread pool. Panics if
+    /// the current thread is not in this thread pool.
+    pub fn current_thread_index(&self) -> usize {
+        unsafe {
+            let curr = WorkerThread::current();
+            assert!(!curr.is_null());
+            (*curr).index()
+        }
+    }
 }
 
 impl Drop for ThreadPool {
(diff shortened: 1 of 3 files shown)


src/api.rs
```diff
@@ -234,6 +234,21 @@ impl ThreadPool {
...
         }
     }
+
+    /// Returns the number of threads in the thread pool.
+    pub fn num_threads(&self) -> usize {
+        self.registry.num_threads()
+    }
+
+    /// Returns the index of the current thread in the thread pool. Panics if
+    /// the current thread is not in this thread pool.
```
nikomatsakis: This actually panics if the current thread is not a Rayon thread, but doesn't check that the current thread is in *this* thread-pool. You would have to compare the registries; I forget if the `RegistryId` code which would enable this is in master, I think it is only in https://github.com/nikomatsakis/rayon/pull/141

nikomatsakis: (That said, I think I would have expected this to just be a free function. And I'd probably prefer for it to just return `Option<usize>` rather than panic.)

nikomatsakis:
OK, I adjusted it to return an `Option` and check the current registry, but I didn't make it a free fn. I think I'm inclined to add two free fns, one to get the number of threads for the "current" thread-pool and one to get the index within that thread-pool, but @bholley mentioned on IRC that it'd be useful for them to double-check that they are in fact in the expected thread-pool, so I left that out for now. Also I have to run to lunch so shorter on time. =)

@bholley take a look...

bholley:
lgtm!

nikomatsakis:
One thing that is I guess missing are some tests...

nikomatsakis:
I'll merge this for now, though I think free fns would also be nice.

PR #187, opened 2016-12-20
tsp: normalize the direction of the benchmarked path

cuviper:
Rarely, I've seen this case fail with a result that's exactly the
reverse of the expected path.  I suppose there's no harm in going the
other way.  So we just detect if the given path is `lt` when reversed,
and then normalize it before asserting its value.


nikomatsakis:
Heh. Interesting.

PR #188, opened 2016-12-20
Add a string `par_split` method for `char` separators

cuviper:
This should give the exact same result at `str::split`, just in parallel!  But we only support a plain `char` for the separator right now, versus the fancy `str::Pattern` trait.  The test cases are taken from the `str::split` docstring, as well as a split on `test.rs` itself for something larger.

### DIFF
--- a/src/par_iter/string.rs
+++ b/src/par_iter/string.rs
@@ -37,6 +37,12 @@ pub trait ParallelString {
     /// Returns a parallel iterator over substrings separated by a
     /// given character, similar to `str::split`.
     fn par_split(&self, char) -> ParSplit;
+
+    /// Returns a parallel iterator over substrings terminated by a
+    /// given character, similar to `str::split_terminator`.  It's
+    /// equivalent to `par_split`, except it doesn't produce an empty
+    /// substring after a trailing terminator.
+    fn par_split_terminator(&self, char) -> ParSplitTerminator;
 }
 
 impl ParallelString for str {
@@ -46,6 +52,10 @@ impl ParallelString for str {
 
     fn par_split(&self, separator: char) -> ParSplit {
         ParSplit::new(self, separator)
+    }
+
+    fn par_split_terminator(&self, terminator: char) -> ParSplitTerminator {
+        ParSplitTerminator::new(self, terminator)
     }
 }
 
@@ -185,5 +195,66 @@ impl<'a> IntoIterator for ParSplit<'a> {
             Iterator::last(&mut empty);
             empty.chain(Some(chars))
         }
+    }
+}
+
+
+// /////////////////////////////////////////////////////////////////////////
+
+pub struct ParSplitTerminator<'a> {
+    splitter: ParSplit<'a>,
+    endpoint: bool,
+}
+
+impl<'a> ParSplitTerminator<'a> {
+    fn new(chars: &'a str, terminator: char) -> Self {
+        ParSplitTerminator {
+            splitter: ParSplit::new(chars, terminator),
+            endpoint: true,
+        }
+    }
+}
+
+impl<'a> ParallelIterator for ParSplitTerminator<'a> {
+    type Item = &'a str;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+        where C: UnindexedConsumer<Self::Item>
+    {
+        bridge_unindexed(self, consumer)
+    }
+}
+
+impl<'a> UnindexedProducer for ParSplitTerminator<'a> {
+    fn split(&mut self) -> Option<Self> {
+        self.splitter.split().map(|right| {
+            let endpoint = self.endpoint;
+            self.endpoint = false;
+            ParSplitTerminator { splitter: right, endpoint: endpoint }
+        })
+    }
+}
+
+impl<'a> IntoIterator for ParSplitTerminator<'a> {
+    type Item = &'a str;
+    type IntoIter = Chain<Split<'a, char>, OptionIter<&'a str>>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        let skip_last = if self.endpoint {
+            let chars = self.splitter.chars;
+            let terminator = self.splitter.separator;
+            chars.is_empty() || chars.ends_with(terminator)
+        } else {
+            false
+        };
+
+        let mut iter = self.splitter.into_iter();
+
+        if skip_last {
+            // eat the empty trailing substring
+            iter.next_back();
+        }
+
+        iter
     }
 }
--- a/src/par_iter/string.rs
+++ b/src/par_iter/string.rs
@@ -240,7 +240,10 @@ impl<'a> UnindexedProducer for ParSplitTerminator<'a> {
         self.splitter.split().map(|right| {
             let endpoint = self.endpoint;
             self.endpoint = false;
-            ParSplitTerminator { splitter: right, endpoint: endpoint }
+            ParSplitTerminator {
+                splitter: right,
+                endpoint: endpoint,
+            }
         })
     }
 }
@@ -280,12 +283,15 @@ impl<'a> ParallelIterator for ParLines<'a> {
     fn drive_unindexed<C>(self, consumer: C) -> C::Result
         where C: UnindexedConsumer<Self::Item>
     {
-        self.0.par_split_terminator('\n').map(|line| {
-            if line.ends_with('\r') {
-                &line[..line.len() - 1]
-            } else {
-                line
-            }
-        }).drive_unindexed(consumer)
+        self.0
+            .par_split_terminator('\n')
+            .map(|line| {
+                if line.ends_with('\r') {
+                    &line[..line.len() - 1]
+                } else {
+                    line
+                }
+            })
+            .drive_unindexed(consumer)
     }
 }
--- a/src/par_iter/string.rs
+++ b/src/par_iter/string.rs
@@ -59,17 +59,15 @@ impl<'a> ParallelIterator for ParChars<'a> {
 }
 
 impl<'a> UnindexedProducer for ParChars<'a> {
-    fn can_split(&self) -> bool {
-        // This is pessimistic, as we only *know* there are multiple characters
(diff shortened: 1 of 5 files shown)


cuviper:
In the worst case, this will scan the entire string from one thread before determining that there are no separators at all.  I don't love this.  The only alternative I came up with was something like:

* add a new `par_char_indices` (which we might want anyway)
* filter that for separators and collect those into a vector
* iterate that vector in parallel to produce substrings

So that would do a more-parallel search, at the expense of an O(n) intermediate vector.  I think the current approach is probably good enough for a first shot, and the edge cases are probably too rare to worry about.

cuviper:
Another option is to use a parallel `find` or `find_first`, but that won't give us any control to avoid scanning the same characters again after each split.

nikomatsakis:
> So that would do a more-parallel search, at the expense of an O(n) intermediate vector. I think the current approach is probably good enough for a first shot, and the edge cases are probably too rare to worry about.

Yeah, I think this is ok, at least for now. We could add a note about it to the documentation -- basically saying that the method is intended for cases where the separator is common.

src/par_iter/string.rs
```diff
@@ -68,3 +81,103 @@ impl<'a> IntoIterator for ParChars<'a> {
...
+        // All done, now we are two!
+        (left_split, right_split)
+    }
+}
+
+impl<'a> IntoIterator for ParSplit<'a> {
+    type Item = &'a str;
+    type IntoIter = Chain<Once<&'a str>, Split<'a, char>>;
+
+    fn into_iter(self) -> Self::IntoIter {
```
nikomatsakis: I wonder if adopting the `fold_with()` method would make this sort of thing easier, since we wouldn't have to return the iterator back out.

src/par_iter/string.rs
```diff
@@ -68,3 +81,103 @@ impl<'a> IntoIterator for ParChars<'a> {
...
+
+    fn split(self) -> (Self, Self) {
+        let ParSplit { chars, separator, first } = self;
+        let first = first.expect("error: splitting with no separator present!");
+
+        // First find a suitable UTF-8 boundary in the unsearched region.
+        let char_index = find_char_midpoint(&chars[first..]) + first;
+
+        // Find a separator in reverse, towards the `first` that we know exists.
+        let index = chars[first..char_index]
```
nikomatsakis: I wonder if it is really worth the trouble to maintain the invariant that we know there exists a `first`? I initially imagined we would not have a field `first`. I guess the problem then is that we can't know whether `can_split` should return true? We could modify `can_split()` to take `&mut self`, so that it can update the producer's internal state (or perhaps remove it entirely, and instead have `split` return `Option<(Self, Self)>`). In that case, we could just do one search before the split, something like:

```rust
let char_index = find_char_midpoint(&chars);
let opt_index = chars[char_index..].find(separator).map(|i| i + char_index).or_else(|| /* search left */ );
...
```

It seems like a shame to have to do two searches sequentially before you can split, basically.

src/par_iter/string.rs
cuviper: It is a good candidate to implement a custom `fold_with`, but we'll still have to have `IntoIterator` to satisfy that default implementation of `fold_with`.  (unless we cheat with `unimplemented!()`)

src/par_iter/string.rs
cuviper: I started down this road, using `Cell` for `can_split` updates, but otherwise basically what you suggest.  It was easy if I didn't care about re-scanning memory we've already covered, but I was getting tripped up trying to avoid that.  Especially, scanning both forward and backwards meant there were two "already scanned" indexes to maintain.

It also doesn't help that `can_split` is called before `Splitter::try`.  Either way we order it, one of those will be doing work that the other might nullify, which is why I tried to keep `can_split` minimal.  But I'm prescanning `first`, so maybe that's moot.  Hrm.  At least `Splitter::try` is well-bounded, so maybe we _should_ do that first.

I'll play more and see if I can make this a bit lazier...

nikomatsakis: we should make a benchmark :)

cuviper:
I rebased, then simplified `UnindexedProducer` by removing `can_split` and making `split` return an `Option`.  This made it easier to limit how much we have to actually scan at a given time, as you suggested.  Then I did a little refactoring, including removing those associated types which were getting cumbersome, and not really necessary.  (And with an eye to the future, they can't support `Fn` separators unless we get ATC.)  Finally, I added `par_split_terminator` and `par_lines`, which build up nicely from `par_split`.

cc @bluss for the `UnindexedProducer` change, since you're trying to use that too.

The next big step will be `Fn` separators, which will allow more generality and especially enable `par_split_whitespace`.  I'd like to figure out how to generalize this into `par_split` without referencing `Pattern` for `find` et al., since that's still unstable, but I have a few ideas.  And anyway, this can be a separate PR later.

nikomatsakis:
Nice! I like the way this looks.

bluss:
the split change looks good to me. I haven't put it through a test or anything, but I think it's implementable all the same. (Removes the panicking case for trying to split too small, too.)

nikomatsakis:
@cuviper needs rebase, but I like this version better

cuviper:
Rebased.

I also tried adding `fold_with`, but I didn't like how much I had to duplicate `into_iter` to make it work, so I haven't committed that yet.  However, it occurs to me that `UnindexedProducer` doesn't _need_ `IntoIterator` anymore, if they all update to `fold_with`.  `Producer` still needs it for `zip` though.  So how would you feel about making that change?  (in a separate PR)

nikomatsakis:
@cuviper 

> However, it occurs to me that UnindexedProducer doesn't need IntoIterator anymore, if they all update to fold_with. Producer still needs it for zip though. So how would you feel about making that change? (in a separate PR)

I feel good about it, I think.

nikomatsakis:
I think I'd feel *better* about it if we had the full specialization stuff working, since then we could do:

```
default impl<T: IntoIterator> UnindexedProducer for T {
    fn fold_with() { ... }
}
```

But oh well.

bluss:
With full specialization I think fold with could be on an entirely separate trait

nikomatsakis:
OK, so I'm 👍 to landing. I guess the only question is whether this is a "breaking change" since it changes the internal traits. I think ... I've decided we probably want to say "yes", since there are people like @bluss poking about at those traits, and any other answer could make cargo's compatibility algorithms into a total mess.

nikomatsakis:
(I wonder if we can somehow refactor such that those internals are exposed in a distinct crater that can be versioned separately.)

bluss:
I have been vacillating between the rayon dependency choices for ndarray the last days.

There are three options for us:

- rayon is an internal dependency. We provide methods like `Array::par_map(f) -> Array` and `Array::par_map_inplace(f)` and expose nothing from rayon. This is the most compatible and least powerful
- rayon and ndarray are integrated in a separate crate. This provides parallel versions of array views and various array iterators. The integration crate can be versioned separately. It effectively lives a life at double pace, stepping one major version for each of rayon's and ndarray's major versions.
  - Advantage: Flexibility of using the Parallelterator implementations directly
  - Drawback: No `IntoParallelIterator` for arrays and arrayviews themselves
- ndarray depends on rayon directly.
  - Advantage: Flexibility of using the Parallelterator implementations directly
  - Advantage: usual rayon traits implemented on arrays, arrayviews, axis iterator
  - Drawback: ndarray itself must step its major version when rayon does.

Rayon isn't the only thing we are integrating with! All of these libraries want to share public types or traits with ndarray:

- rustc-serialize
- serde
- rand
- rayon

ndarray potentially has to live a quintuple speed life (major version if any of those four publish a major version). Forget about 1.0! If we go 1.0, we'll soon be 20.0 too!

`rand` is the only one that's completely separated out, it has less drawbacks, since we only use it for constructors, so ndarray-rand provides an extension trait. For rayon I think we are leaning towards the latter alternative, full integration.

bluss:
On the direct implied question, it's not on my map that you break versioning in the internal parts. If you do, be abundantly clear in which APIs are internal and can break at any time. You could for example be clear in the documentation [“don't implement this trait”](https://bluss.github.io/rust-ndarray/master/ndarray/trait.Dimension.html), and even better, use private marker types to make it impossible the private-only traits outside of the crate.

bluss:
I think this works for closing a trait?

```rust
mod private {
    /// If this type is pub but not publically rechable, third parties can't name it and can't implement the trait.
    pub struct PrivateMarker;
} 

pub trait MyTrait {
    /// This trait is private; this method exists to make it impossible
    /// to implement outside the crate.
    fn private_trait(&self) -> private::PrivateMarker;
}
```

cuviper:
I think closing `ParallelString` makes sense, so we can add new methods without it being a breaking change. But I think @nikomatsakis is referring to `UnindexedProducer`, which should remain open if we want to support third-party iterators like ndarray.  We'll need to settle such questions of "internalness" before 1.0.

bluss:
Hopefully the API questions can be clarified as soon as possible. There is now so much uncertainty being cast on stability existing at all for this, that I can't have ndarray depend on rayon until it's clarified that there exists a versioned interface.

I'm ready for breaking changes and new versions, even often, as long as crate versions are used properly. There is existing ndarray code using UnindexedProducer, but it hasn't been merged yet.

cuviper:
I like the idea of splitting implementation into a separate crate, especially if we assume that most people will just use the traits, not implement them.  So aim for a 1.0 `rayon` interface fairly soon which maintains stability, but have a `rayon-impl` which may still bump semver relatively frequently.  We can use the same trick @bluss suggested for privacy, such that you have to import `rayon-impl` yourself to actually implement anything, but plain users don't have to care.

There may be coordination challenges if multiple implementers have to update, but again, I expect this to be a rarer case then simple rayon use.

nikomatsakis:
@bluss the idea that @withoutboats has been kicked around for "automatic features" are also pretty relevant here

nikomatsakis:
I think the bottom line is that, for semver, we have to be scrupulous. If we want things to change in breaking ways without a new major version for most folks, we should factor them into a separate crate, or else enforce it somehow.

PR #189, opened 2016-12-20
Implement find_first/last, position_first/last

### LINKED ISSUE #151: support `find_first`, `find_last`, `position_first`, and `position_last`
We currently have `find_any` and `position_any`. These are similar to the `find` and `position` methods on `Iterator` but they don't always find the *first* match. It'd be nice to support compatible versions. I would want to give them different names so as to discourage people from using them, since `find_any` will be more efficient.

Implementing this is similar to [the existing `find_any` code](https://github.com/nikomatsakis/rayon/blob/c4674dbe7b6a14a228150153a58150ecb9df1a3c/src/par_iter/find.rs) but a bit trickier. The way that `find_any` works is that it makes a consumer [with a `full()` method](https://github.com/nikomatsakis/rayon/blob/c4674dbe7b6a14a228150153a58150ecb9df1a3c/src/par_iter/find.rs#L53-L55). `full()` returns true if execution should stop. We use a single shared atomic and, when a matching item is found, [we set this flag to true](https://github.com/nikomatsakis/rayon/blob/c4674dbe7b6a14a228150153a58150ecb9df1a3c/src/par_iter/find.rs#L85), signaling others to halt.

To make this work for `find_first`, we need to halt not when *any* solution is found, but if a solution has been found *in the sections to your left*. The easiest way I can think to do this is to have [each call to `split_off`](https://github.com/nikomatsakis/rayon/blob/c4674dbe7b6a14a228150153a58150ecb9df1a3c/src/par_iter/find.rs#L62-L64), which breaks the consumer into two pieces, allocate an `Arc<AtomicBool>` (or something) for those two pieces to share. The left half will signal if it finds something by writing true; the right half will stop when either it finds something, or it sees that true has been written. The reduce method will then act [just like it does now](https://github.com/nikomatsakis/rayon/blob/c4674dbe7b6a14a228150153a58150ecb9df1a3c/src/par_iter/find.rs#L105), preferring a value from the left half over the right half.

schuster:
Fixes #151 

This is my first PR to Rayon, so any and all feedback is welcome.

### DIFF
--- a/src/par_iter/find_first_last.rs
+++ b/src/par_iter/find_first_last.rs
@@ -23,6 +23,14 @@ enum MatchPosition {
     Rightmost,
 }
 
+// Returns true if pos1 is a better match than pos2 according to MatchPosition
+fn better_position(pos1: usize, pos2: usize, mp: MatchPosition) -> bool {
+    match mp {
+        MatchPosition::Leftmost => pos1 < pos2,
+        MatchPosition::Rightmost => pos1 > pos2,
+    }
+}
+
 pub fn find_first<PAR_ITER, FIND_OP>(pi: PAR_ITER, find_op: FIND_OP) -> Option<PAR_ITER::Item>
     where PAR_ITER: ParallelIterator,
           FIND_OP: Fn(&PAR_ITER::Item) -> bool + Sync
@@ -61,6 +69,13 @@ impl<'f, FIND_OP> FindConsumer<'f, FIND_OP> {
             best_found: best_found,
         }
     }
+
+    fn current_index(&self) -> usize {
+        match self.match_position {
+            MatchPosition::Leftmost => self.lower_bound.get(),
+            MatchPosition::Rightmost => self.upper_bound
+        }
+    }
 }
 
 impl<'f, ITEM, FIND_OP> Consumer<ITEM> for FindConsumer<'f, FIND_OP>
@@ -85,10 +100,7 @@ impl<'f, ITEM, FIND_OP> Consumer<ITEM> for FindConsumer<'f, FIND_OP>
     fn into_folder(self) -> Self::Folder {
         FindFolder {
             find_op: self.find_op,
-            boundary: match self.match_position {
-                MatchPosition::Leftmost => self.lower_bound.get(),
-                MatchPosition::Rightmost => self.upper_bound
-            },
+            boundary: self.current_index(),
             match_position: self.match_position,
             best_found: self.best_found,
             item: None,
@@ -96,13 +108,11 @@ impl<'f, ITEM, FIND_OP> Consumer<ITEM> for FindConsumer<'f, FIND_OP>
     }
 
     fn full(&self) -> bool {
-        let best = self.best_found.load(Ordering::Relaxed);
-        match self.match_position {
-            // can stop consuming if the best found index so far is *strictly*
-            // better than anything this consumer will find
-            MatchPosition::Leftmost => best < self.lower_bound.get(),
-            MatchPosition::Rightmost => best > self.upper_bound
-        }
+        // can stop consuming if the best found index so far is *strictly*
+        // better than anything this consumer will find
+        better_position(self.best_found.load(Ordering::Relaxed),
+                        self.current_index(),
+                        self.match_position)
     }
 }
 
@@ -161,11 +171,9 @@ impl<'f, FIND_OP: 'f + Fn(&ITEM) -> bool, ITEM> Folder<ITEM> for FindFolder<'f, 
     }
 
     fn full(&self) -> bool {
-        let best_found = self.best_found.load(Ordering::Relaxed);
-        match self.match_position {
-            MatchPosition::Leftmost => best_found < self.boundary,
-            MatchPosition::Rightmost => best_found > self.boundary,
-        }
+        better_position(self.best_found.load(Ordering::Relaxed),
+                        self.boundary,
+                        self.match_position)
     }
 }
 
--- a/src/par_iter/find_first_last.rs
+++ b/src/par_iter/find_first_last.rs
@@ -1,5 +1,6 @@
 use std::cell::Cell;
 use std::sync::atomic::{AtomicUsize, Ordering};
+use super::internal::*;
 use super::*;
 use super::len::*;
 
--- a/src/par_iter/find_first_last.rs
+++ b/src/par_iter/find_first_last.rs
@@ -246,20 +246,20 @@ impl<'f, FIND_OP: 'f + Fn(&ITEM) -> bool, ITEM> Folder<ITEM> for FindFolder<'f, 
             // Continuously try to set best_found until we succeed or we
             // discover a better match was already found.
             let mut current = self.best_found.load(Ordering::Relaxed);
-            let mut exchange_result = Result::Err(0);
-            while !better_position(current, self.boundary, self.match_position) &&
-                  exchange_result.is_err() {
-                exchange_result = self.best_found.compare_exchange_weak(current,
-                                                                        self.boundary,
-                                                                        Ordering::Relaxed,
-                                                                        Ordering::Relaxed);
-                if exchange_result.is_err() {
-                    current = self.best_found.load(Ordering::Relaxed);
+            loop {
+                if better_position(current, self.boundary, self.match_position) {
+                    break;
                 }
-            }
-
-            if exchange_result.is_ok() {
-                self.item = Some(item);
+                match self.best_found.compare_exchange_weak(current,
+                                                            self.boundary,
+                                                            Ordering::Relaxed,
+                                                            Ordering::Relaxed) {
+                    Ok(_) => {
+                        self.item = Some(item);
+                        break;
+                    },
+                    Err(v) => current = v,
+                }
             }
         }
         self
--- a/src/par_iter/find_first_last.rs
+++ b/src/par_iter/find_first_last.rs
@@ -1,0 +1,182 @@
+use std::cell::Cell;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use super::*;
+use super::len::*;
+
+// The consumer for find_first/find_last has fake indexes representing the lower
+// and upper bounds of the "range" of data it consumes. This range does not
+// correspond to indexes from the consumed iterator but rather indicate the
+// consumer's position relative to other consumers. The purpose is to allow a
+// consumer to know it should stop consuming items when another consumer finds a
+// better match.
+
+// An indexed consumer could specialize to use the real indexes instead, but we
+// don't implement that for now. The only downside of the current approach is
+// that in some cases, iterators very close to each other will have the same
+// range and therefore not be able to stop processing if one of them finds a
+// better match than the others.
+
+#[derive(Copy, Clone)]
+enum MatchPosition {
+    Leftmost,
+    Rightmost,
+}
+
+pub fn find_first<PAR_ITER, FIND_OP>(pi: PAR_ITER, find_op: FIND_OP) -> Option<PAR_ITER::Item>
+    where PAR_ITER: ParallelIterator,
+          FIND_OP: Fn(&PAR_ITER::Item) -> bool + Sync
+{
+    let best_found = AtomicUsize::new(usize::max_value());
+    let consumer = FindConsumer::new(&find_op, MatchPosition::Leftmost, &best_found);
+    pi.drive_unindexed(consumer)
+}
+
+pub fn find_last<PAR_ITER, FIND_OP>(pi: PAR_ITER, find_op: FIND_OP) -> Option<PAR_ITER::Item>
+    where PAR_ITER: ParallelIterator,
+          FIND_OP: Fn(&PAR_ITER::Item) -> bool + Sync
+{
+    let best_found = AtomicUsize::new(0);
+    let consumer = FindConsumer::new(&find_op, MatchPosition::Rightmost, &best_found);
+    pi.drive_unindexed(consumer)
+}
+
+struct FindConsumer<'f, FIND_OP: 'f> {
+    find_op: &'f FIND_OP,
+    lower_bound: Cell<usize>,
+    upper_bound: usize,
+    match_position: MatchPosition,
+    best_found: &'f AtomicUsize,
+}
+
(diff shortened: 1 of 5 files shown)


cuviper:
It looks like you're missing some imports for the stable compilers, possibly related to ordering like #173.
(IOW, try adding `use super::internal::*;` _before_ `use super::*;`)

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,183 @@
...
+          FIND_OP: Fn(&ITEM) -> bool + Sync
+{
+    fn split_off(&self) -> Self {
+        // Upper bound for one consumer will be lower bound for the other. This
+        // overlap is okay, because only one of the bounds will be used for
+        // comparing against best_found; the other is kept only to be able to
+        // divide the range in half
+        let old_lower_bound = self.lower_bound.get();
+        let median = old_lower_bound + ((self.upper_bound - old_lower_bound) / 2);
+        self.lower_bound.set(median);
```
cuviper: So, it's important to your algorithm that the result from `split_off` is always used for the left/lower side by whatever is calling it.  Near as I can tell, that's always true in practice, but I don't think we've ever specified it must be.  It's just generally convenient to call in order `(consumer.split_off(), consumer)` when splitting in two.

I don't think there's a problem here, but we need to be careful, and this deserves a comment at least.

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,183 @@
...
+impl<'f, FIND_OP: 'f + Fn(&ITEM) -> bool, ITEM> Folder<ITEM> for FindFolder<'f, ITEM, FIND_OP> {
+    type Result = Option<ITEM>;
+
+    fn consume(mut self, item: ITEM) -> Self {
+        if (self.find_op)(&item) {
+            // This may sometimes set best_found to a worse index than it was
+            // before, depending on timing. This means more consumers will
+            // continue to run than necessary, but the reducer will still ensure
+            // the correct value is returned.
+            self.best_found.swap(self.boundary, Ordering::Relaxed);
```
cuviper: I think we can do better than that with a `compare_and_swap` loop, repeating until you either get a successful swap or find that someone else swapped something "better" than you anyway.

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    fn complete(self) -> Self::Result {
+        self.item
+    }
+
+    fn full(&self) -> bool {
+        let best_found = self.best_found.load(Ordering::Relaxed);
+        match self.match_position {
+            MatchPosition::Leftmost => best_found < self.boundary,
```
cuviper: For `Leftmost` only, isn't `self.item.is_some()` also sufficient?

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,183 @@
+use std::cell::Cell;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use super::internal::*;
+use super::*;
+use super::len::*;
+
+// The consumer for find_first/find_last has fake indexes representing the lower
```
nikomatsakis: 👍 to leaving some comments on how things work =)

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,183 @@
...
+// and upper bounds of the "range" of data it consumes. This range does not
+// correspond to indexes from the consumed iterator but rather indicate the
+// consumer's position relative to other consumers. The purpose is to allow a
+// consumer to know it should stop consuming items when another consumer finds a
+// better match.
+
```
nikomatsakis: Total nit: I'd like a `//` here to indicate that these are two paragraphs in the same comment.

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,183 @@
...
+// An indexed consumer could specialize to use the real indexes instead, but we
+// don't implement that for now. The only downside of the current approach is
+// that in some cases, iterators very close to each other will have the same
+// range and therefore not be able to stop processing if one of them finds a
+// better match than the others.
```
nikomatsakis: Hmm, can you elaborate here?

src/par_iter/find_first_last.rs
nikomatsakis: Perhaps we can rename `split_off()` to `split_left()` or `split_off_left()` or `split_left_off()`, something like that.

src/par_iter/find_first_last.rs
nikomatsakis: Yes, seems like it'd be worth using a compare-exchange here. I'd probably use `compare_exchange_weak`, but that's a minor thing.

src/par_iter/find_first_last.rs
nikomatsakis: Wait, am I confused? Why is `self.item()` sufficient? Don't we want to also stop if some other folder (to our left) has found an item?

src/par_iter/mod.rs
```diff
@@ -575,6 +576,28 @@ pub trait ParallelIterator: Sized {
...
+    /// Once a match is found, all attempts to the left of the match
+    /// will be stopped, while attempts to the right must continue in case
+    /// a later match is found.
+    fn find_last<FIND_OP>(self, predicate: FIND_OP) -> Option<Self::Item>
+        where FIND_OP: Fn(&Self::Item) -> bool + Sync {
+        find_first_last::find_last(self, predicate)
+    }
+
     #[doc(hidden)]
     #[deprecated(note = "parallel `find` does not search in order -- use `find_any`")]
```
nikomatsakis: let's update this deprecated note to say: "use `find_any`, `find_first`, or `find_last`" -- we might also consider calling `find_first()`, not sure.

src/par_iter/find_first_last.rs
nikomatsakis: OK, I read the code now, so I understand, but I think an example would be great. Basically just saying: we start out with an iterator over 0..usize::max, then split this range in half, and so forth.

Also, should we use `u64` just to get that much more resolution?

Also, I think we need a test for the case where we have an iterator that exceeds the resolution of the false indices. Or at least some careful comments explaining why it is correct. =)

nikomatsakis:
This is nice =)

src/par_iter/find_first_last.rs
schuster: @nikomatsakis I think he's saying it's *also* sufficient, which is correct (and a case I didn't think about). If we've already found the item, we need to make sure we stop before we find some other item slightly to the right.

So @cuviper, I think you're right that that logic should be added to both `full` and `consume`.

src/par_iter/find_first_last.rs
schuster: Good point; I'll add a comment. +1 to renaming `split_off()`.

src/par_iter/find_first_last.rs
schuster: Thanks, I'll work on that explanation.

I think `AtomicU64` is not stable yet (tracked [here](https://github.com/rust-lang/rust/issues/32976)). Would it be worth postponing merging this PR until it's stable? Of course, we should be able to make that change later, because the exact size of the integer used is private to this module.

I agree that a test would be useful to validate my hand-waving in the comments. Any thoughts on how to write that test? I don't yet know enough about the splitting logic to know how to force that case. One possibility is to start with a `FindConsumer` whose upper and lower bounds are the same, but that requires either making the `FindConsumer` struct public so it can be used in the test module, or writing a test directly in this module (which none of the rest of this project does).

nikomatsakis: > I think AtomicU64 is not stable yet (tracked here). Would it be worth postponing merging this PR until it's stable? Of course, we should be able to make that change later, because the exact size of the integer used is private to this module.

Nah, just use a usize. It'll make it easier to write a test. :)

As for the test... well, you could... perhaps write something that drives the splitting by hand? i.e., take a u64 range, create a producer from it, and manually call split etc? The test can be internal to this module if it needs access to private state...

...at minimum some thorough comments for how that case is handled would be a good start.

src/par_iter/find_first_last.rs
cuviper: Right, it's an additional case for stopping.

nikomatsakis: yes, ok. I agree (this seems to relate to having some tests that cover this case)

schuster:
Just pushed with updates to address all feedback.

schuster:
Actually, I just found a bug in the range calculation. Stand by for a fix.

schuster:
Fixed. Turned out to be a problem with the test itself, not the code.

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,319 @@
...
+// To solve this, we assign each consumer a lower and upper bound for an
+// imaginary "range" of data that it consumes. The initial consumer starts with
+// the range 0..usize::max_value(). The split divides this range in half so that
+// one resulting consumer has the range 0..(usize::max_value() / 2), and the
+// other has (usize::max_value() / 2)..usize::max_value(). Every subsequent
+// split divides the range in half again until it cannot be split anymore
+// (i.e. its length is 1), in which case the split returns two consumers with
+// the same range. In that case both consumers will continue to consume all
+// their data regardless of whether a better match is found, but the reducer
+// will still return the correct answer.
```
nikomatsakis: Great comment =) Very clear.

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,319 @@
...
+fn same_range_consumers_return_correct_answer() {
+    let find_op = |x: &i32| x % 2 == 0;
+
+    let first_found = AtomicUsize::new(usize::max_value());
+    let first_consumer = FindConsumer::new(&find_op,
+                                           MatchPosition::Leftmost,
+                                           &first_found);
+
+    // split until we have an indivisible range
+    let bits_in_usize = usize::min_value().count_zeros();
```
nikomatsakis: cute

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    let left_first_folder = left_first_folder.consume(0).consume(1);
+    assert_eq!(left_first_folder.boundary, right_first_folder.boundary);
+    // expect not full even though a better match has been found because the
+    // ranges are the same
+    assert!(!right_first_folder.full());
+    let right_first_folder = right_first_folder.consume(2).consume(3);
+    assert_eq!(first_reducer.reduce(left_first_folder.complete(),
+                                    right_first_folder.complete()),
+               Some(0));
```
nikomatsakis: one nit -- while we are testing all this, maybe we can keep around also the consumer where we expect `full()` to be true? For example:

```
for i in 0..bits_in_usize - 1 {
    first_consumer.split_off();
}

let second_folder = first_consumer.split_off().into_folder();
...
assert!(!right_first_folder.full()); 
assert!(second_folder.full());
```

The name `second_folder` is kind of weak though...but whatever =)

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    // same test, but for find_last
+    let last_found = AtomicUsize::new(0);
```
nikomatsakis: can you break this into a distinct `#[test]`?

nikomatsakis: also, as a style nit, at least in rayon I would tend to convert this into a directory module (i.e., `find_first_last/mod.rs` and put the tests in `find_first_last/test.rs`). But .... I don't really care about this. Maybe just move them to the end of the file or something. 

(I would prefer not to have an inline `mod test { ... }` module, as many people prefer. I find managing the imports more annoying that way. I'm idiosyncratic like that I guess.)

src/par_iter/find_first_last.rs
```diff
@@ -0,0 +1,319 @@
...
+            // discover a better match was already found.
+            let mut current = self.best_found.load(Ordering::Relaxed);
+            let mut exchange_result = Result::Err(0);
+            while !better_position(current, self.boundary, self.match_position) &&
+                  exchange_result.is_err() {
+                exchange_result = self.best_found.compare_exchange_weak(current,
+                                                                        self.boundary,
+                                                                        Ordering::Relaxed,
+                                                                        Ordering::Relaxed);
+                if exchange_result.is_err() {
```
nikomatsakis: Actually, `compare_exchange_weak` gives you back the value that you failed to swap in its `Err` variant. So you could rewrite this as:

```rust
let mut current = self.best_found.load(Ordering::Relaxed);
loop {
    if better_position(current, self.boundary, self.match_position) {
        break;
    }
    match self.best_found.compare_exchange_weak(current, self.boundary, Relaxed, Relaxed) {
        Ok(_) => {
            self.item = Some(item);
            break;
        }
        Err(v) => current = v,
    }
}
```

This seems mildly clearer (and more efficient) to me (I also prefer the `loop+break` formulation over the `while` loop; it's a nit, but I find it easier for me to parse what is happening).

nikomatsakis:
So one question is whether we should rename `split_off()` in this PR to reflect its new role. We can probably do it in a later PR.

src/par_iter/find_first_last.rs
schuster: Do you mean that you want to see a consumer from outside of the exhausted range, so that we can check that it's full when the consumer from inside the range finds a match? That wasn't quite what your example code does (you'd have to move the creation of `second_folder` above the loop, or something like that), but I think that's what you're going for.

src/par_iter/find_first_last.rs
schuster: This was actually the most direct way I could find to get the number of bits in a usize, believe it or not. Is there some other way that I missed?

src/par_iter/find_first_last.rs
nikomatsakis: Yes, that's what I meant.

src/par_iter/find_first_last.rs
schuster: Agreed, I also find your formulation easier to understand.

schuster:
A later PR makes sense to me, too - it's a logically separate change. I could submit that PR as soon as this one goes through.

nikomatsakis:
Looks great thanks @schuster !

PR #196, opened 2016-12-31
Updated num_cpus dependency

Aaronepower:
Updating allows for rayon to be compiled for fuchsia and emscripten.


nikomatsakis:
looks good. 👍

PR #199, opened 2017-01-02
Fix some compiler warnings from find_first/last

schuster:
I just found a few compiler warnings I forgot about from find_first/find_last that I went ahead and fixed.


nikomatsakis:
Thanks :)

PR #201, opened 2017-01-03
Give one last tickle to terminated registries

### LINKED ISSUE #200: Sometimes `thread_pool::test::workers_stop` doesn't stop
As noted in https://github.com/nikomatsakis/rayon/pull/194#issuecomment-269992219, there was a Travis failure where `thread_pool::test::workers_stop` didn't stop, hanging until Travis's 10 minute timeout.  I restarted that build and it completed.  However, running that test in a loop locally, I did eventually catch a hung test.

Here's gdb `thread info` and backtraces: [gdb.txt](https://github.com/nikomatsakis/rayon/files/681799/gdb.txt).  Looks like the threads are:

- Id 1 (LWP 11147): the main thread, waiting for test completion.
- Id 12 (LWP 11148): the thread created by the `#[test]` harness, stuck in a `LockLatch::wait()` under `wait_until_stopped()`
- Id 2-11: 10 remaining threads (out of 22) from this test's thread pool, all asleep.  I guess they missed a tickle?

I'll try to keep this process alive, in case there's more information we want to extract...

cuviper:
`Registry::terminate` was merely setting `terminate_latch`, leaving
sleeping threads in an eternal slumber.  Normally that would just amount
to a resource leak, but `thread_pool::test::workers_stop` will stall
waiting for every thread to stop.  We just need a `tickle` to make sure
every thread will wake up and exit.

Fixes #200.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -181,6 +181,7 @@ impl Registry {
     /// extant work is completed.
     pub fn terminate(&self) {
         self.terminate_latch.set();
+        self.sleep.tickle(usize::MAX);
     }
 }
 
--- a/src/thread_pool/test.rs
+++ b/src/thread_pool/test.rs
@@ -34,3 +34,22 @@ fn workers_stop() {
     // should lead to worker threads stopping
     registry.wait_until_stopped();
 }
+
+#[test]
+fn sleeper_stop() {
+    use std::{thread, time};
+
+    let registry;
+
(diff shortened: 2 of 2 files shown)


cuviper:
I added a simple test which reliably triggered the problem before, and passes now.

nikomatsakis:
Ah, nice!

Issue #142, opened 2016-11-15
Feature Request: ability to name Rayon threads


jwilm:
Currently the thread pool used by Rayon inherits their names from the process title. We like to track CPU utilization by thread in our application; it would be nice if Rayon added a configuration parameter to name the threads.

I'd be willing to make a PR if the feature would be accepted, but I wanted to check that's the case first.

Thanks!

nikomatsakis:
@jwilm seems like a reasonable feature, sure!

I imagine it would be done by modifying the `Configuration` to include a "base name" for the Rayon threads or something like that, and then modifying the `Registry::new()` function (in `src/thread_pool.rs`) to give the threads names.

martinhath:
I suppose this can be closed, as #202 is merged 😄

jwilm:
Done! Thanks for implementing it :D

PR #202, opened 2017-01-09
Add support for naming threads

martinhath:
This PR adds support for specifying names for rayon threads, and fixes issue #142.

Most notable things are:
 - `Registry::new` takes the entire `Configuration`, not just the number of threads
 - The thread name is `format!("{}{}", name, i)`

I found it difficult to test the thread naming. Any ideas?

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -104,13 +104,11 @@ impl Registry {
 
         for (index, worker) in workers.into_iter().enumerate() {
             let registry = registry.clone();
+            let mut b = thread::Builder::new();
             if let Some(name) = configuration.base_thread_name(index) {
-                let _ = thread::Builder::new()
-                    .name(name)
-                    .spawn(move || unsafe { main_loop(worker, registry, index) });
-            } else {
-                thread::spawn(move || unsafe { main_loop(worker, registry, index) });
+                b = b.name(name);
             }
+            b.spawn(move || unsafe { main_loop(worker, registry, index) });
         }
 
         registry
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -78,7 +78,7 @@ unsafe fn init_registry(config: Configuration) {
 }
 
 impl Registry {
-    pub fn new(configuration: Configuration) -> Arc<Registry> {
+    pub fn new(mut configuration: Configuration) -> Arc<Registry> {
         let limit_value = match configuration.num_threads() {
             Some(value) => value,
             None => {
@@ -104,9 +104,9 @@ impl Registry {
 
         for (index, worker) in workers.into_iter().enumerate() {
             let registry = registry.clone();
-            if let Some(name) = configuration.base_thread_name() {
+            if let Some(name) = configuration.base_thread_name(index) {
                 let _ = thread::Builder::new()
-                    .name(format!("{}{}", name, index))
+                    .name(name)
                     .spawn(move || unsafe { main_loop(worker, registry, index) });
             } else {
                 thread::spawn(move || unsafe { main_loop(worker, registry, index) });
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -73,18 +73,20 @@ pub fn get_registry_with_config(config: Configuration) -> &'static Registry {
 /// function. Declared `unsafe` because it writes to `THE_REGISTRY` in
 /// an unsynchronized fashion.
 unsafe fn init_registry(config: Configuration) {
-    let registry = leak(Arc::new(Registry::new(config.num_threads())));
+    let registry = leak(Arc::new(Registry::new(config)));
     THE_REGISTRY = Some(registry);
 }
 
 impl Registry {
-    pub fn new(num_threads: Option<usize>) -> Arc<Registry> {
-        let limit_value = match num_threads {
+    pub fn new(configuration: Configuration) -> Arc<Registry> {
+        let limit_value = match configuration.num_threads() {
             Some(value) => value,
-            None => match env::var("RAYON_RS_NUM_CPUS") {
-                Ok(s) => usize::from_str(&s).expect("invalid value for RAYON_RS_NUM_CPUS"),
-                Err(_) => num_cpus::get(),
-            },
+            None => {
+                match env::var("RAYON_RS_NUM_CPUS") {
+                    Ok(s) => usize::from_str(&s).expect("invalid value for RAYON_RS_NUM_CPUS"),
+                    Err(_) => num_cpus::get(),
+                }
+            }
         };
 
         let (inj_worker, inj_stealer) = deque::new();
@@ -102,7 +104,13 @@ impl Registry {
 
         for (index, worker) in workers.into_iter().enumerate() {
             let registry = registry.clone();
-            thread::spawn(move || unsafe { main_loop(worker, registry, index) });
+            if let Some(name) = configuration.base_thread_name() {
+                let _ = thread::Builder::new()
+                    .name(format!("{}{}", name, index))
+                    .spawn(move || unsafe { main_loop(worker, registry, index) });
+            } else {
+                thread::spawn(move || unsafe { main_loop(worker, registry, index) });
+            }
         }
 
         registry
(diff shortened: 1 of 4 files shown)


cuviper:
> The thread name is format!("{}{}", name, i)

I wonder if we could take a closure instead? Like `FnMut(index: usize) -> String` -- I'm not sure how well lifetimes can fit into the builder pattern though, so it might have to be `'static`.

cuviper:
> I found it difficult to test the thread naming. Any ideas?

Maybe do a par_iter collect of `std::thread::current().name()` into a hash set, then compare expectations.

martinhath:
So I replaced `base_thread_name: Option<String>` with `get_base_thread_name: Option<Box<FnMut(usize) -> String>>`. I messed around with `&Fn*`s, but didn't really get anything to work, and the one thing I did get to work, was seemingly impossible to call(?) 😛 

I think the API is fairly OK to use, but I'm not sure about which tradeoffs there are here. 

Also, I put the test in `rayon/tests/run-pass`. Is this okay?

emilio:
So we finally really want to expose thread indexes?

cuviper:
I'm OK with the boxing, since this is a short-lived allocation, not in some core loop.  Let's see what @nikomatsakis thinks too.

> Also, I put the test in rayon/tests/run-pass. Is this okay?

I think it's OK, but `src/thread_pool/test.rs` is another option.  You would apply the naming in a private thread pool in that case, same as those existing tests.

cuviper:
> So we finally really want to expose thread indexes?

I'm not worried about that myself.  But perhaps we should expose it more directly too, if we're going to make it visible in the thread names anyway.

jwilm:
> So we finally really want to expose thread indexes?

Seems like if a user wanted a counter with the thread name, it could be added without exposing rayon's internal ids. A closure like `Fn() -> String` could manage a counter itself.

src/registry.rs
```diff
@@ -102,7 +104,13 @@ impl Registry {
 
         for (index, worker) in workers.into_iter().enumerate() {
             let registry = registry.clone();
-            thread::spawn(move || unsafe { main_loop(worker, registry, index) });
+            if let Some(name) = configuration.base_thread_name(index) {
+                let _ = thread::Builder::new()
```
nikomatsakis: I think I would rather see this factored out:

```rust
let mut b = thread::Builder::new();
if let Some(n) = ... {
    b.name(name);
}
b.spawn(...);
```

nikomatsakis:
I'm not too worried about people scraping thread indices from the thread name. I also don't think the overhead of boxing is a particularly big deal. I wonder though if it is really necessary to go all out and supply an arbitrary closure though -- I am a bit nervous at introducing execution of arbitrary code into the initialization path. For example, what if the closure panics? I don't think we've analyzed what would happen in that code-path if it had to unwind. Seems like it might be simpler to start out with just appending the index onto the name (as originally proposed).

cuviper:
> For example, what if the closure panics? I don't think we've analyzed what would happen in that code-path if it had to unwind.

Whether or not we abandon the closure idea, we should analyze this.  Even `thread::spawn` can panic!

nikomatsakis:
> Whether or not we abandon the closure idea, we should analyze this. Even thread::spawn can panic!

Hmm, I suppose it probably can under the proper conditions.

martinhath:
Just so I am on the same page as you on the thread panic thing: the issue is that if the supplied closure panics on, say `index == 4`, it is not obvious what will happen to the parallel computation (surely it won't complete succesfully)?

nikomatsakis:
@martinhath well, for example, what will happen to the other threads that have already started? will they ever terminate?

nikomatsakis:
> I'm not too worried about people scraping thread indices from the thread name.

Still, we should probably leave a note indicating that the precise format of thread names is not considered a "stable" part of the interface.

nikomatsakis:
So, given that `spawn()` could also panic, I agree that this panic safety thing is a pre-existing issue and I filed #205 about it. I still think running an arbitrary closure exacerbates the problem but I withdraw my objection. =)

As for leaking thread indices, we already decided to leak them for custom thread-pools; we didn't yet add any global APIs for doing so, but I expect we will eventually, so I'm not that worried about leaking them at startup. 

I still suspect that most people will just want to set a prefix and append numbers to it, so we might want a convenience function for that, though using this API at all is rather rare.

src/registry.rs
```diff
@@ -102,7 +104,11 @@ impl Registry {
 
         for (index, worker) in workers.into_iter().enumerate() {
             let registry = registry.clone();
-            thread::spawn(move || unsafe { main_loop(worker, registry, index) });
+            let mut b = thread::Builder::new();
+            if let Some(name) = configuration.base_thread_name(index) {
+                b = b.name(name);
+            }
+            b.spawn(move || unsafe { main_loop(worker, registry, index) });
```
cuviper: `Builder::spawn()` returns a `Result` -- shouldn't we check that?
(I'm confused why `#[must_use]` didn't flag this...)

cuviper: Oh, it actually did in travis:

    warning: unused result which must be used, #[warn(unused_must_use)] on by default

Now I'm surprised this isn't default-deny, but perhaps we should set that ourselves.

martinhath: Will simply `.unwrap()` the call suffice? Failing to spawn a thread seems a serious enough issue to shut down everything to me. 

I haven't really understood the initialization routine of rayon yet; is the initialization performed in the main thread, or simply in the thread first to call eg. `.par_iter()`, such that a `panic` in the init logic will cause only that thread to panic?

martinhath: Maybe if this requires some thinking and logic to handle well, it's better to do it in a PR related to #205 ?

nikomatsakis: > Now I'm surprised this isn't default-deny, but perhaps we should set that ourselves.

Yeah, we should set `#[deny(warnings)]` I guess -- it's easy to miss things when you're reviewing.

nikomatsakis: @martinhath it does seem like this is more about #205; you could probably add something like

```
b.spawn(...).unwrap(); // FIXME(#205) -- handle this error
```

cuviper:
So @nikomatsakis expressed hesitation on gitter, whether we really need this to be a closure.  My response is that the thread name is such a cosmetic thing that it's a classic "color of the bikeshed" sort of problem, and there's not really any reason for Rayon to have an opinion about it.  Some users may want $NAME_$INDEX, some may want just $NAME, who knows!  So a closure punts the question back to the user who wanted a thread name in the first place, since by default we don't name at all.

@martinhath My first comment about this was a soft "I wonder if ...", and you immediately ran with it.  That's fine, but do you personally have an opinion about it?

If we keep the closure, then calling this a "base" thread name is a misnomer, since the closure returns the exact name we're going to set.  I'd just drop the "base" qualifier everywhere.

martinhath:
> do you personally have an opinion about it?

While I see there are concerns wrt. running an arbitrary closure in the init. path, I prefer the closure API to the initial API, simply becasue it gives the user more freedom. I also think the API is more self explanatory with the closure; the user would perhaps guess that the thread index is appended to the given string, but the user would have to check the docs to be sure (one could argue that this part of the API is obscure enough for any user to have checked the docs anyways).

```rust
let c = Configuration::new().set_thread_name("thread-");
let c = Configuration::new().set_thread_name(|i| format!("thread-{}", i) );
```

Is is a little more to write with the closure API, but I do think it makes up for it by being clearer.


I definitely agree that namings are a little off here 😁 The change was more "uuh, well let's see what I can get working" than anything else.

nikomatsakis:
I'm persuaded. Let's do this.

martinhath:
But I still think we should follow @cuviper's suggestion of changing the naming from `base_thread_name` to `thread_name` 😛

nikomatsakis:
@martinhath ah, yeah, missed that :) follow-up PR?

emilk:
For anyone stumbling on this thread today, this is how you name your threads in rayon 1.7:

``` rs
    rayon::ThreadPoolBuilder::new()
        .thread_name(|i| format!("rayon-{i}"))
        .build_global()
        .unwrap();
```

PR #203, opened 2017-01-09
link to the video from RustBeltRust

nikomatsakis:


cuviper:
Nice!  And thanks for the shoutout on the splitter PR. :)

PR #206, opened 2017-01-13
Rename `base_thread_name` to `thread_name`

martinhath:
This was missed while merging #202 😄


src/configuration.rs
```diff
@@ -65,15 +65,15 @@ impl Configuration {
     }
 
     /// Get the base thread name for the given index.
```
cuviper: Nit: update the docstring too, please.

cuviper:
Thanks!

PR #209, opened 2017-01-15
Fork into `rayon-core` and a stable `rayon` facade

cuviper:
The root crate is now `rayon-core`, with all functionality, and under
`rayon-stable/` is the new `rayon` crate that only publicizes select
really-stable interfaces -- those we feel are nearly ready for 1.0.  For
the most part, this means the APIs for *using* rayon, while we can
continue tweaking the APIs for *implementing* rayon traits in the core
with proper pre-1.0 semver bumps.

(The directories are structured this way just to minimize the code churn.)

Most of the traits are now split, e.g. `ParallelIterator` and a new
`ParallelIteratorImpl`, so only the former is publicized in `rayon`.
Methods like `drive` and `with_producer`, anything that implementors
have to provide, are isolated in `*Impl`.

### DIFF
--- a/rayon-stable/src/lib.rs
+++ b/rayon-stable/src/lib.rs
@@ -1,0 +1,13 @@
+extern crate rayon_core;
+
+pub mod par_iter;
+
+pub use rayon_core::prelude;
+
+pub use rayon_core::Configuration;
+pub use rayon_core::InitError;
+pub use rayon_core::dump_stats;
+pub use rayon_core::initialize;
+pub use rayon_core::ThreadPool;
+pub use rayon_core::join;
+pub use rayon_core::{scope, Scope};
--- a/src/par_iter/imp.rs
+++ b/src/par_iter/imp.rs
@@ -1,0 +1,79 @@
+use super::*;
+
+/// The `ParallelIterator` implementation interface.
+pub trait ParallelIteratorImpl: Sized {
+    type Item: Send;
+
+    /// Internal method used to define the behavior of this parallel
+    /// iterator. You should not need to call this directly.
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result where C: UnindexedConsumer<Self::Item>;
+
+    /// Returns the number of items produced by this iterator, if known
+    /// statically. This can be used by consumers to trigger special fast
+    /// paths. Therefore, if `Some(_)` is returned, this iterator must only
+    /// use the (indexed) `Consumer` methods when driving a consumer, such
+    /// as `split_at()`. Calling `UnindexedConsumer::split_off_left()` or
+    /// other `UnindexedConsumer` methods -- or returning an inaccurate
+    /// value -- may result in panics.
+    ///
+    /// This is hidden & considered internal for now, until we decide
+    /// whether it makes sense for a public API.  Right now it is only used
+    /// to optimize `collect`  for want of true Rust specialization.
+    #[doc(hidden)]
+    fn opt_len(&mut self) -> Option<usize> {
+        None
+    }
+}
+
+impl<T> ParallelIterator for T where T: ParallelIteratorImpl {}
+
+
+/// A trait for parallel iterators items where the precise number of
+/// items is not known, but we can at least give an upper-bound. These
+/// sorts of iterators result from filtering.
+pub trait BoundedParallelIteratorImpl: ParallelIterator {
+    fn impl_upper_bound(&mut self) -> usize;
+
+    /// Internal method used to define the behavior of this parallel
+    /// iterator. You should not need to call this directly.
+    fn drive<'c, C: Consumer<Self::Item>>(self, consumer: C) -> C::Result;
+}
+
+impl<T> BoundedParallelIterator for T
+    where T: BoundedParallelIteratorImpl
+{
+    fn upper_bound(&mut self) -> usize {
+        self.impl_upper_bound()
+    }
+}
+
+
+/// A trait for parallel iterators items where the precise number of
+/// items is known. This occurs when e.g. iterating over a
+/// vector. Knowing precisely how many items will be produced is very
+/// useful.
+pub trait ExactParallelIteratorImpl: BoundedParallelIterator {
+    /// Produces an exact count of how many items this iterator will
+    /// produce, presuming no panic occurs.
+    fn impl_len(&mut self) -> usize;
+}
+
+impl<T> ExactParallelIterator for T
+    where T: ExactParallelIteratorImpl
+{
+    fn len(&mut self) -> usize {
+        self.impl_len()
+    }
+}
+
+
+/// An iterator that supports "random access" to its data, meaning
+/// that you can split it at arbitrary indices and draw data from
+/// those points.
+pub trait IndexedParallelIteratorImpl: ExactParallelIterator {
+    /// Convert this parallel iterator into a producer that can be
+    /// used to request the items.
+    fn with_producer<CB: ProducerCallback<Self::Item>>(self, callback: CB) -> CB::Output;
+}
+
+impl<T> IndexedParallelIterator for T where T: IndexedParallelIteratorImpl {}
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -577,7 +577,8 @@ pub trait ParallelIterator: Sized {
     /// will be stopped, while attempts to the left must continue in case
     /// an earlier match is found.
     fn find_first<FIND_OP>(self, predicate: FIND_OP) -> Option<Self::Item>
-        where FIND_OP: Fn(&Self::Item) -> bool + Sync {
+        where FIND_OP: Fn(&Self::Item) -> bool + Sync
+    {
         find_first_last::find_first(self, predicate)
     }
 
@@ -588,12 +589,14 @@ pub trait ParallelIterator: Sized {
     /// will be stopped, while attempts to the right must continue in case
     /// a later match is found.
     fn find_last<FIND_OP>(self, predicate: FIND_OP) -> Option<Self::Item>
-        where FIND_OP: Fn(&Self::Item) -> bool + Sync {
+        where FIND_OP: Fn(&Self::Item) -> bool + Sync
+    {
         find_first_last::find_last(self, predicate)
     }
 
     #[doc(hidden)]
-    #[deprecated(note = "parallel `find` does not search in order -- use `find_any`, `find_first`, or `find_last`")]
+    #[deprecated(note = "parallel `find` does not search in order -- \
+                 use `find_any`, `find_first`, or `find_last`")]
     fn find<FIND_OP>(self, predicate: FIND_OP) -> Option<Self::Item>
         where FIND_OP: Fn(&Self::Item) -> bool + Sync
     {
@@ -874,7 +877,8 @@ pub trait IndexedParallelIterator: ExactParallelIterator {
     }
 
     #[doc(hidden)]
-    #[deprecated(note = "parallel `position` does not search in order -- use `position_any`, `position_first`, or `position_last`")]
+    #[deprecated(note = "parallel `position` does not search in order -- \
+                 use `position_any`, `position_first`, or `position_last`")]
     fn position<POSITION_OP>(self, predicate: POSITION_OP) -> Option<usize>
         where POSITION_OP: Fn(Self::Item) -> bool + Sync
     {
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -238,6 +238,8 @@ pub trait ParallelIterator: ParallelIteratorImpl {
     /// Example:
     ///
     /// ```
+    /// # extern crate rayon_core as rayon;
+    /// # fn main() {
     /// // Iterate over a sequence of pairs `(x0, y0), ..., (xN, yN)`
     /// // and use reduce to compute one pair `(x0 + ... + xN, y0 + ... + yN)`
     /// // where the first/second elements are summed separately.
@@ -248,6 +250,7 @@ pub trait ParallelIterator: ParallelIteratorImpl {
     ///            .reduce(|| (0, 0), // the "identity" is 0 in both columns
     ///                    |a, b| (a.0 + b.0, a.1 + b.1));
     /// assert_eq!(sums, (0 + 5 + 16 + 8, 1 + 6 + 2 + 9));
+    /// # }
     /// ```
     ///
     /// **Note:** unlike a sequential `fold` operation, the order in
@@ -372,6 +375,8 @@ pub trait ParallelIterator: ParallelIteratorImpl {
     /// map/reduce, you might try this:
     ///
     /// ```
+    /// # extern crate rayon_core as rayon;
+    /// # fn main() {
     /// use rayon::prelude::*;
     /// let s =
     ///     ['a', 'b', 'c', 'd', 'e']
@@ -380,6 +385,7 @@ pub trait ParallelIterator: ParallelIteratorImpl {
     ///     .reduce(|| String::new(),
     ///             |mut a: String, b: String| { a.push_str(&b); a });
     /// assert_eq!(s, "abcde");
+    /// # }
     /// ```
     ///
     /// Because reduce produces the same type of element as its input,
@@ -389,6 +395,8 @@ pub trait ParallelIterator: ParallelIteratorImpl {
     /// do this instead:
     ///
     /// ```
+    /// # extern crate rayon_core as rayon;
+    /// # fn main() {
     /// use rayon::prelude::*;
     /// let s =
     ///     ['a', 'b', 'c', 'd', 'e']
@@ -398,6 +406,7 @@ pub trait ParallelIterator: ParallelIteratorImpl {
     ///     .reduce(|| String::new(),
     ///             |mut a: String, b: String| { a.push_str(&b); a });
     /// assert_eq!(s, "abcde");
+    /// # }
     /// ```
     ///
     /// Now `fold` will process groups of our characters at a time,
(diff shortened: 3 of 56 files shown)


src/par_iter/imp.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// iterator. You should not need to call this directly.
+    fn drive<'c, C: Consumer<Self::Item>>(self, consumer: C) -> C::Result;
+}
+
+impl<T> BoundedParallelIterator for T
+    where T: BoundedParallelIteratorImpl
+{
+    fn upper_bound(&mut self) -> usize {
+        self.impl_upper_bound()
+    }
```
cuviper: I don't love this indirection myself, but it seemed necessary if we want to keep `upper_bound` in the public API.  I wanted to avoid having to implement *two* bounded traits, instead of just requiring `*Impl` and everything is fine.  The `impl_` prefix is just to avoid calling ambiguity for anyone who has both traits in scope.  Ditto `ExactParallelIterator::len()` below.

nikomatsakis:
Exciting. I'll take a look soon.

bluss:
I am reading this with curiosity, it's a hard problem to solve, and I would love to have new ideas.

Because rayon has a public reexport of `ParallelIterator` from `rayon-core`, updating the version of the rayon-core dependency in rayon is a breaking change (doesn't matter if the trait doesn't change, if the version is incompatible). From a cursory look this achieves separation of the codebases, but does not change the version story since it's using a so called public dependency.

bluss:
Ah so rayon-core can be stabilized quickly then. That makes sense.

nikomatsakis:
> Ah so rayon-core can be stabilized quickly then. That makes sense. (Edit: But then all the impls must be in rayon, not rayon-core)

Wait, I'm still a bit confused. Is the idea that we should put `rayon-core` at 1.0 but have a big warning that it does not obey semver, except for those parts that are publicly re-exported?

src/par_iter/mod.rs
```diff
@@ -20,44 +20,49 @@ use self::enumerate::Enumerate;
...
 
+pub use self::from_par_iter::FromParallelIterator;
 pub use self::string::ParallelString;
 
-pub mod find;
-pub mod find_first_last;
-pub mod chain;
-pub mod collect;
-pub mod enumerate;
-pub mod filter;
```
nikomatsakis: While I definitely think we should make these modules private, I think we also need to continue to publicly export types like `par_iter::filter::Filter` -- that is, the return types of iterator adapters, much like the corresponding types in libstd are public. This perhaps amounts to making things like `use self::filter::Filter` above into `pub use`.

nikomatsakis:
OK, so, I remain a bit confused. I would have expected the public bits to be *moved* into `rayon-stable` rather than re-exported. 

In any case, while I still think the idea of splitting into `rayon`, `rayon-core`, and (eventually) `rayon-pool` makes sense, I want to game out what's going to happen just a bit more. Basically the idea is that 99% of consumers just want the public surface API so they use `rayon` at 1.0 and they are happy. Meanwhile, rayon 1.0 depends on rayon-core 0.1 (say). Then rayon 1.1 adds some nice features and we bump rayon-core to 0.2 because it's backwards incompatible. Since this is a private dependency, nobody cares.

The exception to this will be @bluss. If ndarray is using rayon-core 0.1 to create `ParallelIterator` APIs, even if they are hidden, and then you try to combine that with rayon 1.1, you will wind up with two copies of rayon-core (0.1 and 0.2). This...is actually fine, except for the thread-pool. 

But if we manage to separate out the *thread-pool* into rayon-pool, hopefully *that* can go to 1.0 fairly quickly. (The thread-pool interface is relatively stable.) That would mean that rayon-core 0.1 and 0.2 would hopefully both depend on the same version number.

(I've been waiting to separate out the thread-pool interface until we land the futures work and so forth, since it is the only thing lately that has required refactoring, but it seems to me it should happen before 1.0.)

bluss:
I'm not sure it's possible to create the crate structure that meets the requirements here. I haven't had much time to experiment though, so I'm just raising in the problems I see without a full understanding of the whole map of possibilities.

src/par_iter/mod.rs
cuviper: OK, sure, we can publicize those too.

cuviper:
@bluss 
> Because rayon has a public reexport of ParallelIterator from rayon-core, updating the version of the rayon-core dependency in rayon is a breaking change (doesn't matter if the trait doesn't change, if the version is incompatible).

Hmm.  I think I see how this is technically true, but I'm not sure what a break would look like in practice.  Can you sketch out a scenario?

> Ah so rayon-core can be stabilized quickly then.

Er, no, the intention is the opposite!  Stabilize `rayon`, and keep iterating `rayon-core`.  It's quite possible that I've entirely missed the mark. :/

@nikomatsakis 
> The exception to this will be @bluss. If ndarray is using rayon-core 0.1 to create ParallelIterator APIs, even if they are hidden, and then you try to combine that with rayon 1.1, you will wind up with two copies of rayon-core (0.1 and 0.2). This...is actually fine, except for the thread-pool.

Ah, I think this gets closer to the break that @bluss refers too.  If my project is using `rayon=1` and `ndarray`, then `cargo update` *should* be smart enough to resolve that we're stuck on `rayon=1.0` to use the same `rayon-core`.  But if I force my project to `rayon=1.1` for some new feature I want, then we'll get the duped `rayon-core`, and more importantly `ndarray` no longer implements *my* `ParallelIterator`.

(Except in your "even if they are hidden" case, then that's not a problem.  The thread pool is it's own problem, which I agree seems solvable by an additional stable crate.)

Anyways this still feels to me like a better situation than we have now, where impl changes have to be a hard bump regardless of whether most users would see a difference.  But technically, it's a break.  Could we allow this flavor of break under the domain of minor bumps, 1.x -> 1.(x+1), or must it be a major break?

cuviper:
> I would have expected the public bits to be moved into rayon-stable rather than re-exported.

FWIW, I went down this road at first.  The problem I ran into was something like this:

```rust
// in `rayon-core`
pub trait ParallelIteratorImpl {...}
```

```rust
// in `rayon` (stable)
pub trait ParallelIterator: ParallelIteratorImpl {...}
impl<T: ParallelIteratorImpl> ParallelIterator for T {...}

pub trait IntoParallelIterator {...}
impl<T: ParallelIterator> IntoParallelIterator for T {...}

impl<T: Send> IntoParallelIterator for Vec<T> {...}
```

The compiler complains that these are overlapping `IntoParallelIterator` implementations.  There's no *actual* overlap, but I think the fact that `Vec<T>` and `ParallelIteratorImpl` are both external means we can't guarantee it's not implemented, right?  So there's a *possible* path for `Vec<T>` to implement `ParallelIteratorImpl`, implying `ParallelIterator`, implying `IntoParallelIterator`, making the direct `IntoParallelIterator` a coherence problem.

We'd have to break the chain, probably at `impl<T: ParallelIteratorImpl> ParallelIterator for T`, which would force implementors to write both `ParallelIteratorImpl` and an empty `ParallelIterator`.

bluss:
crate1 depends on rayon-core 0.1 directly

```
use rayon_core::ParallelIterator;

fn foo<I: ParallelIterator>(iter: I) { 
}
```

crate2 depends on rayon 1.0 (which uses rayon-core 0.1) and depends on crate1:

```
crate1::foo([1, 2, 3].par_iter())
```

Now if you upgrade crate2 to use rayon 1.1 (which uses rayon-core 0.2), crate1 will be compiled with rayon-core 0.1 and crate2 with rayon-core 0.2 and there are two incompatible ParallelIterator (doesn't matter if they have identical definition). This is because `ParallelIterator` is used in interchange between crates.

cuviper:
Another thing I've been thinking about, but didn't try, is to remove the supertrait relationship.  The `ParallelIterator` trait would not provide definitions for any of its methods, but only `ParallelIteratorImpl` would fill those out.  Something like:

```rust
 // no mention of Impl here
pub trait ParallelIterator {
    fn count(self) -> usize; // etc
}

pub trait ParallelIteratorImpl {...}
impl<T: ParallelIteratorImpl> ParallelIterator for T {
    fn count(self) -> usize {...}
}
```

I think that will document a little cleaner, but doesn't solve the semver problems.

cuviper:
> `fn foo<I: ParallelIterator>(iter: I)`

OK, yeah, that's an even more direct path to sorrow.

cuviper:
Here's another possibility (but this may be getting silly)

```
rayon-core :: ParallelIterator
 |  \
 |   rayon-impl :: ParallelIteratorImpl, Producer, etc.
 |  /
rayon :: IntoParallelIterator, and impls for std types
```

So `rayon-core` and `rayon` would fast-track to 1.0, while `rayon-impl` keeps iterating.

Maybe `rayon-impl` could even more generally be a `rayon-unstable` test bed.

bluss:
I notice that ParallelIterator is structured in a way so that it delegates to a producer impl, and this producer impl does not have to be public in any way (it's not a return type or associated type or so). This is neat, I wonder if that kind of design can be extended so that it is useful here.

cuviper:
`Producer` et al. actually are still exposed, and these are generally what I'm trying to hide in `*Impl`:

- `ParallelIterator::drive_unindexed` has an `UnindexedConsumer` parameter
- `BoundedParallelIterator::drive` has a `Consumer` parameter
- `IndexedParallelIterator::with_producer` has a `ProducerCallback` parameter and returns its `Output`
  - `ProducerCallback::callback` has a `Producer` parameter

Only `UnindexedProducer` escapes as an internal detail, and that's mainly because the other traits offer no means to stack these together, like how a `MapProducer` is built on another `Producer` in `with_producer`.

nikomatsakis:
Sorry I've been quiet here. I've been wanting to pull this branch and kind of study it in a bit more detail, but I haven't had a chance. This does seem like one of the biggest 1.0 priorities though.

nikomatsakis:
@cuviper 

> Anyways this still feels to me like a better situation than we have now, where impl changes have to be a hard bump regardless of whether most users would see a difference. But technically, it's a break. Could we allow this flavor of break under the domain of minor bumps, 1.x -> 1.(x+1), or must it be a major break?

Re-reading the comments, I didn't really understand what you mean here. What is "this" in the sentence, the current PR? And I'm not sure what "technically, it's a break" means either. I think perhaps what you mean is that if rayon-stable bumps from 1.0 to 1.1 and in the process changes from rayon-core 0.1 to rayon-core 0.2, that is technically a breaking change, but one you are tempted to allow? I can see the motivation to allow it, for sure, but it feels like cheating, since the traits *are* part of the public interface. I'd have to think over whether it can cause visible breakage if you are only using rayon-stable, seems... at least unlikely to do so.

What all of this is driving home for me though is that separating the thread-pool is the most important thing. After all, if we jump from Rayon 1.0 to Rayon 2.0, that's not the end of the world, so long as both of them depend on Rayon-ThreadPool 1.0, right?

To that end, perhaps that is worth exploring a bit. I'd like to land my "async spawn" branch first, but I could make a branch.

rayon-stable/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+extern crate rayon_core;
+
+pub mod par_iter;
+
+pub use rayon_core::prelude;
+
+pub use rayon_core::Configuration;
+pub use rayon_core::InitError;
+pub use rayon_core::dump_stats;
```
nikomatsakis: this method does not need to be public, I think, and perhaps should be removed =)

nikomatsakis:
I've been pondering over this. I don't have an answer but I have a few thoughts I'd like to get input on. First, I am not sure if `rayon-stable` and `rayon-core` is the right way to break down the crates. But I'll actually get to that later, let's continue with those names just for a bit, because I to come to my second point. I think there is a core trade-off and I am not convinced that we can get around it:

- If the `ParallelIterator` traits can be implemented by anyone outside of `rayon-stable`, then it will be a breaking change to make `rayon-stable` switch from `rayon-core = 0.1` to `rayon-core = 0.2`.
    - Put another way, if the `ParallelIterator` traits can be implemented from outside of `rayon-stable`, then `rayon-core` cannot be an internal dependency.
- If however the `ParallelIterator` traits **cannot** be implemented outside of `rayon-stable` (e.g., via various privacy tricks), then we can make `rayon-core` into a true internal dependency.
    - But then @bluss would have to reproduce the whole facade.
    - Or, alternatively, we encourage @bluss to hack on ndarray integration in `rayon-stable` under a feature flag (`with_ndarray` or some such) until we feel we are ready to "bless" the internal traits and make them publicly accessible.

In my head, what I want is some way that "implementing" the `ParallelIterator` traits kind of "opts in" to rayon-core or something but I think this is a fantasy. If we want ndarray to be able to implement the (stable) `ParallelIterator` traits, it is going to have to offer *some* way to coerce itself into the (unstable) `Producer` or `Consumer` traits, so that the `foreach` methods etc can operator on it. That means that it will have a dependency on `rayon-core=0.1`, in order that it implements the `ParallelIterator` traits defined in rayon 1.0. So when we bump to `rayon-core=0.2` and release `rayon=1.1`, ndarray's implementation of the traits will no longer compile with Rayon 1.1. I can't see any way to wave our hands and call that not a breaking change. And it seems likely to affect users. For now, @bluss is the only one I know of daring enough to implement the parallel iterator traits themselves, but this can't last (and we don't want it to).

If, on the other hand, we decide to keep `Consumer` and `Producer` *internal*, this is easy enough to achieve. We can simply make them inaccessible and keep the existing `#[doc(hidden)]` methods like `drive_consumer`. But now if someone wants to implement `ParallelIterator`, they will indeed find that they cannot. (I think by the way that these tricks are quite obviously a feature request for having more ability to customize the coherence rules; in this case, to "opt-in" to a more stringent orphan check, but I guess that's obvious enough.)

**TL;DR I think we can't have it both ways:** If @bluss and others can implement `ParallelIterator` themselves, ithen some amount of internal details must be stable. **So we should decide whether we want to support that, or if we want to encourage @bluss to iterate "in tree".** Obviously @bluss you have a vote here. =)

That leads me to another thought. In terms of separating out the thread-pool, which is the most important thing, it might make sense to do it along a slightly different boundary than I had initially thought:

- rayon-threadpool:
    - hides unsafe details of the thread-pool
    - exports only the safe `join`, `scope`, and the `spawn_async` APIs
        - maybe also `Configuration` and friends
- rayon:
    - re-exports rayon-threadpool APIs
    - adds the `ParallelIterator` traits

My motivation here is that the `join`, `scope`, etc APIs are not going to change in **backwards incompatible** ways. Much **more** likely is that their **implementation** will change dramatically; i.e., the internal details of how the thread-pool works. But if that is all encapsulated within the `rayon-threadpool` crate (we may want a different name), those can be semver-compatible updates.  Meanwhile, if people are using rayon 1.0 and rayon 2.0, but both of them depend on rayon-threadpool 1.x, who cares really.

(Originally, I wanted to move the "unsafe threadpool" abstraction into its own crate, basically, but I think that is premature, and will result in more problems, not less.)

nikomatsakis:
@cuviper do you agree with my analysis then? Shall I take a stab at the split I'm recommending so we can compare it as an alternative?

cuviper:
Right, so I guess the conclusion is I'm trying to build a stable interface that basically extends a less stable one, and this is a foundation of sand.  Bah, humbug.  😄 

But yes, I agree that we have a more stable foundation in the threadpool, join, etc. -- basically everything but `par_iter` -- and it makes sense to separate those.  The implementation may change, even dramatically, but we should be able to commit to a 1.x series ~forever.  We should take a fine-toothed comb over it though to be sure we have all the bounds we can anticipate needing.

I'm also still inclined to call that base crate `rayon-core`, but naming is classically subjective.  And FWIW there's no actual `rayon-stable` crate in this PR -- that's what I called the directory, but the crate is still just `rayon`.

I'm happy to welcome ndarray implementations here, if @bluss would like, to lessen the maintenance burden if we do change things.  Probably behind a feature so the dependency is optional, but maybe default enabled.  Of course, this assumes its implementation doesn't need private access within ndarray, but given ndarray-parallel exists separately, I expect it's ok.

But in the course of world domination, we can't expect to integrate all implementors forever...

bluss:
Separating out the thread pool sounds great, that improves compatibility all around. Thanks for inviting me to vote @nikomatsakis, I just don't know much about things that I haven't tried.

First we note that a feature `with_ndarray` also makes ndarray a public dependency of rayon. And ndarray is still in the “we are iterating fast and doing lots of versions” phase. This is not desireable either. I think that I would only want to develop in tree if that solution truly worked differently and fixed the versioning story, but it doesn't.

What I can think of is that rayon could mark all the APIs internal, and tell that they are not versioned (so regular dependencies will break on them), and then there is a single rayon-contrib crate that uses that unstable interface and is upgraded in lock step. I guess rayon-contrib only has major versions (never compatible upgrades) and public dependencies on rayon and each of the crates it integrates with.

nikomatsakis:
@cuviper I'm not sure where I got `rayon-stable` from, but anyhow I think `rayon-core` and `rayon` is a fine name for the crates. As good as any.

-----

@bluss 

> And ndarray is still in the “we are iterating fast and doing lots of versions” phase. This is not desireable either. 

Indeed.

> What I can think of is that rayon could mark all the APIs internal, and tell that they are not versioned (so regular dependencies will break on them), and then there is a single rayon-contrib crate that uses that unstable interface and is upgraded in lock step.

Hmm. I'm not *thrilled* about just declaring things as "internal" -- but I guess it's ok. But I don't quite understand the role of the rayon-contrib crate here. It's upgraded in lock step with what?

bluss:
rayon-contrib upgrades in lock step with rayon (but that's just one of the reasons it will upgrade, it will upgrade for dependencies too). It becomes hard to handle correctly, I guess.

nikomatsakis:
Closing in favor of https://github.com/nikomatsakis/rayon/pull/227

PR #211, opened 2017-01-16
Require `UnindexedProducer::fold_with` without `IntoIterator`           

cuviper:
Our `string::ParSplit` can write a better `fold_with` than its iterator, because it can bypass the chain-Option iterator logic.  However, this duplicates some slightly tricky logic if we still keep the trait requirement for `IntoIterator`.  We can drop that and just *require* `fold_with`, without a default implementation.  It's trivial now for those that do have `IntoIterator` to use `Folder::consume_iter`.

### DIFF
--- a/src/par_iter/internal.rs
+++ b/src/par_iter/internal.rs
@@ -33,16 +33,10 @@ pub trait Producer: IntoIterator + Send + Sized {
     /// stop when the folder is full (or all elements have been consumed).
     ///
     /// The provided implementation is sufficient for most iterables.
-    fn fold_with<F>(self, mut folder: F) -> F
+    fn fold_with<F>(self, folder: F) -> F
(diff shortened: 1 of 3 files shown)


nikomatsakis:
Ah, I like this.

nikomatsakis:
Hmm, I wonder if this would make it easier to support `.rev()` as well? 

cc @schuster

nikomatsakis:
Ah, I guess probably not, as this is specific to `UnindexedProducer`. Still, there may be room for some sort of transform like this.

cuviper:
Right, `Producer` still needs iterators for `zip()`, at least.

nikomatsakis:
@cuviper yeah, I mean, the whole *point* of producer is to be pull access, for certain. So never mind. =)

PR #213, opened 2017-01-21
Fix typo

hniksic:
Noticed the typo because I copied and pasted the licensing blurb into a [library of my own.](https://github.com/hniksic/rust-subprocess) :)


nikomatsakis:
Thanks =)

PR #214, opened 2017-01-22
Add `par_split(Fn)` and `par_split_whitespace` for strings

### LINKED ISSUE #133: `par_split_whitespace` for string

cuviper:
Both `par_split` and `par_split_terminator` can now accept `Fn(char) -> bool` for the separator, in addition to a simple `char`.  This is somewhat modeled after the standard library `Pattern`, but that's not stable so we can't use it directly.  From this, `par_split_whitespace` follows easily.

This also adds a "private" method to `ParallelString`, as suggested by @bluss, so we can add methods without worrying about breaking third-party implementations.

Fixes #133.

### DIFF
--- a/src/par_iter/string.rs
+++ b/src/par_iter/string.rs
@@ -103,6 +103,30 @@ impl Pattern for char {
     }
 }
 
+impl<FN: Sync + Fn(char) -> bool> Pattern for FN {
+    fn find_in(&self, chars: &str) -> Option<usize> {
+        chars.find(self)
+    }
+
+    fn rfind_in(&self, chars: &str) -> Option<usize> {
+        chars.rfind(self)
+    }
+
+    fn is_suffix_of(&self, chars: &str) -> bool {
+        chars.ends_with(self)
+    }
+
+    fn fold_with<'ch, F>(&self, chars: &'ch str, folder: F, skip_last: bool) -> F
+        where F: Folder<&'ch str>
+    {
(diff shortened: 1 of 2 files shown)


nikomatsakis:
This looks really nice 👍

PR #216, opened 2017-01-26
Require `Send` output in `map`, `cloned`, `filter_map`

### LINKED ISSUE #204: Consider tightening bounds in ParallelIterator.
One example:

```rust
fn map<MAP_OP, R>(self, map_op: MAP_OP) -> Map<Self, MapFn<MAP_OP>>
where MAP_OP: Fn(Self::Item) -> R + Sync
```

The requirement that `R: Send` comes later, for the implementation of trait `MapOp`. This kind of indirection leads to error messages coming at later stages and being less clear. (I have not tested this idea yet.)

cuviper:
All three of these methods lacked any constraints on their output types,
even though it has to be `Send` for the returned adaptor to itself be a
`ParallelIterator`.  This sometimes leads to error messages in usage
that give no indication that missing `Send` is the real problem.  If we
add that requirement up front, it should be a more helpful error.

Fixes #204.

### DIFF
--- a/src/par_iter/mod.rs
+++ b/src/par_iter/mod.rs
@@ -176,7 +176,8 @@ pub trait ParallelIterator: Sized {
     /// Applies `map_op` to each item of this iterator, producing a new
     /// iterator with the results.
     fn map<MAP_OP, R>(self, map_op: MAP_OP) -> Map<Self, MapFn<MAP_OP>>
-        where MAP_OP: Fn(Self::Item) -> R + Sync
+        where MAP_OP: Fn(Self::Item) -> R + Sync,
+              R: Send
     {
         Map::new(self, MapFn(map_op))
     }
@@ -184,7 +185,7 @@ pub trait ParallelIterator: Sized {
     /// Creates an iterator which clones all of its elements.  This may be
     /// useful when you have an iterator over `&T`, but you need `T`.
     fn cloned<'a, T>(self) -> Map<Self, MapCloned>
-        where T: 'a + Clone,
+        where T: 'a + Clone + Send,
               Self: ParallelIterator<Item = &'a T>
     {
         Map::new(self, MapCloned)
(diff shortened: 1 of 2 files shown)


cuviper:
> All three of these methods lacked any constraints on their output types,

Well, `cloned` did have *some* constraints, just not `Send`. :)

shepmaster:
> it should be a more helpful error.

Did you have a chance to see what the error would be for the originally-reported code?

cuviper:
I didn't try the originally-reported code exactly, but on the new compile-fail test I added, without the `R: Send` you just get this:

```
error: no method named `count` found for type `rayon::par_iter::map::Map<rayon::par_iter::option::OptionIter<&NoSend>, rayon::par_iter::map::MapFn<[closure@tests/compile-fail/no_send_par_iter.rs:17:14: 17:20]>>` in the current scope
  --> tests/compile-fail/no_send_par_iter.rs:18:10
   |
18 |         .count(); //~ ERROR no method named `count`
   |          ^^^^^
   |
   = note: the method `count` exists but the following trait bounds were not satisfied: `rayon::par_iter::map::MapFn<[closure@tests/compile-fail/no_send_par_iter.rs:17:14: 17:20]> : rayon::par_iter::map::MapOp<_>`, `rayon::par_iter::map::Map<rayon::par_iter::option::OptionIter<&NoSend>, rayon::par_iter::map::MapFn<[closure@tests/compile-fail/no_send_par_iter.rs:17:14: 17:20]>> : std::iter::Iterator`
```

Now you get:

```
error[E0277]: the trait bound `*const (): std::marker::Send` is not satisfied in `NoSend`
  --> tests/compile-fail/no_send_par_iter.rs:17:10
   |
17 |         .map(|&x| x) //~ ERROR Send` is not satisfied
   |          ^^^ within `NoSend`, the trait `std::marker::Send` is not implemented for `*const ()`
   |
   = note: `*const ()` cannot be sent between threads safely
   = note: required because it appears within the type `NoSend`

error: no method named `count` found for type `rayon::par_iter::map::Map<rayon::par_iter::option::OptionIter<&NoSend>, rayon::par_iter::map::MapFn<[closure@tests/compile-fail/no_send_par_iter.rs:17:14: 17:20]>>` in the current scope
  --> tests/compile-fail/no_send_par_iter.rs:18:10
   |
18 |         .count(); //~ ERROR no method named `count`
   |          ^^^^^
   |
   = note: the method `count` exists but the following trait bounds were not satisfied: `rayon::par_iter::map::MapFn<[closure@tests/compile-fail/no_send_par_iter.rs:17:14: 17:20]> : rayon::par_iter::map::MapOp<_>`, `rayon::par_iter::map::Map<rayon::par_iter::option::OptionIter<&NoSend>, rayon::par_iter::map::MapFn<[closure@tests/compile-fail/no_send_par_iter.rs:17:14: 17:20]>> : std::iter::Iterator`
```

So the confusing `MapOp` message is still there, but the first message gives a better hint what went wrong.

PR #217, opened 2017-01-26
Use `find_first` to short-circuit `cmp` and `partial_cmp`

cuviper:
Lexicographic comparisons are entirely decided by the first unequal
elements, which `find_first` can pull out for us.


cuviper:
Now that I look at this, I think it's also wrong that it doesn't take the length of the iterators into account.  When all else is equal, the longer one should be considered `Greater` overall.  This mistake is in `eq` and `ne` too.  I can push an update with a fix and tests for this later, or as a separate PR.

nikomatsakis:
@cuviper hmm, yeah, good point

cuviper:
OK, lengths should be correctly considered now.

PR #218, opened 2017-01-26
doc: the `Producer::split_at` index can equal the length

### LINKED ISSUE #215: `Producer::split_at` inconsistency
`Producer::split_at(index: usize)` says for a sequence of length `N`, index must be less than `N`.

```rust
    /// Split into two producers; one produces items `0..index`, the
    /// other `index..N`. Index must be less than `N`.
    fn split_at(self, index: usize) -> (Self, Self);
```

The ChainProducer impl does not follow this rule.

The reason I ask is that:

- It would be expected that `index <= N`, which is usual for splitting vectors.
- However `index < N` is very useful, since the split between the last element and the end is only easy to implement for sequences where the "pointer to one past the end" is allowed.
  - This is not the case for a reversed view of a vector (“stride -1”); we are not allowed to have a slice that points to the element before the start of the vector's memory, even if it has length 0.
  - (ndarray already implements splitting using the `index <= N` rule in some cases, but it has extra complexity to do so, and I ran into this when trying to implement `split_at` for more iterators that have exactly the negative stride case to care for).

I suppose the simple fix is to update docs to align with the implementation's perception of the `Producer` trait's interface.

cuviper:
It was documented that the index must be strictly less than the length,
but in practice it may also equal the length, especially in the current
`ChainProducer` that often carries a zero-length side around.

Fixes #215.

### DIFF
--- a/src/par_iter/internal.rs
+++ b/src/par_iter/internal.rs
@@ -26,7 +26,7 @@ pub trait Producer: IntoIterator + Send + Sized {
     fn cost(&mut self, len: usize) -> f64;
 
     /// Split into two producers; one produces items `0..index`, the
-    /// other `index..N`. Index must be less than `N`.
+    /// other `index..N`. Index must be less than or equal to `N`.
     fn split_at(self, index: usize) -> (Self, Self);
 
     /// Iterate the producer, feeding each element to `folder`, and


bluss:
Thanks!

PR #220, opened 2017-02-02
cleanup redundant pointer type in global variable

F001:
The `Arc` pointer behind a `&'static` reference seems useless.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -50,13 +50,13 @@ struct RegistryState {
 /// ////////////////////////////////////////////////////////////////////////
 /// Initialization
 
-static mut THE_REGISTRY: Option<&'static Arc<Registry>> = None;
+static mut THE_REGISTRY: Option<&'static Registry> = None;
 static THE_REGISTRY_SET: Once = ONCE_INIT;
 
 /// Starts the worker threads (if that has not already happened). If
 /// initialization has not already occurred, use the default
 /// configuration.
-pub fn global_registry() -> &'static Arc<Registry> {
+pub fn global_registry() -> &'static Registry {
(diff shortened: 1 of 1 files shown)


src/registry.rs
```diff
@@ -73,7 +73,7 @@ pub fn get_registry_with_config(config: Configuration) -> &'static Registry {
 /// function. Declared `unsafe` because it writes to `THE_REGISTRY` in
 /// an unsynchronized fashion.
 unsafe fn init_registry(config: Configuration) {
-    let registry = leak(Arc::new(Registry::new(config)));
+    let registry = leak(Registry::new(config));
     THE_REGISTRY = Some(registry);
```
cuviper: Huh.  Note that `Registry::new` returns `Arc<Registry>`, so after `leak` it was `registry: &'static Arc<Arc<Registry>>`.  I guess some auto-deref magic was letting this be written to `THE_REGISTRY: Option<&'static Arc<Registry>>`.

Now with your change, we have `registry: &'static Arc<Registry>` writing to `THE_REGISTRY: Option<&'static Registry>`, so there's still some auto-deref going on in forming `Some(registry)`.

I wonder, why shouldn't we just forget `fn leak` and use `THE_REGISTRY: Option<Arc<Registry>>` ?

cuviper: error[E0493]: destructors in statics are an unstable feature

That's why.

nikomatsakis:
I believe that this interacts with an open PR of mine (in a bad way). I think we always want to be able to get an `Arc<Registry>`, and only the master register is in a static.

F001:
Ok, I'll close this PR.

PR #222, opened 2017-02-13
Correct "either readers or writers"

torkleyy:


cuviper:
Yeah, that's more precise, thanks!

PR #224, opened 2017-02-23
Implement rev

### LINKED ISSUE #150: support `rev` on indexed producers
We should support the standard `rev()` combinator from sequential iterators. I am only envisioning supporting it for *indexed* parallel iterators. 

I think to do this we'd have to extend the `Producer` trait so that it can (at least sometimes? maybe always?) support a `rev()` method. I'm not 100% sure.

The basic steps though would be as follows:

- We would be modeling this change roughly on the [existing `Enumerate` combinator](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L6-L8); so we'll add a type like `Rev<M>`
- Like `Enumerate`, `Rev<M>` will implement the various parallel iterators traits, but [only if `M: IndexedParallelIterator`](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L16-L17)
- By far the most complex part is implementing [the `with_producer()` method](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L51-L72). As [described in the parallel iterator README](https://github.com/nikomatsakis/rayon/blob/master/src/par_iter/README.md#what-on-earth-is-producercallback), this method is basically a callback scheme:
    - The heart of it is that you [ask your base producer (the one that we are going to be skipping over) to call you back with a producer](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0//src/par_iter/enumerate.rs#L54).
    - But it can't use Rust's closures, so you have to [make a custom struct](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L56-L58) which [implements the `ProducerCallback` trait](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L60-L62).
    - Ultimately, [the `callback()` method of this struct will be invoked](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L67-L69) with the base producer.
    - At this point, `Enumerate` [wraps that producer it was given in its own producer](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L67-L68) and then [invokes the callback with the wrapper producer](https://github.com/nikomatsakis/rayon/tree/feacfc2d7baebb2df0c9a99699952db7d51185c0/src/par_iter/enumerate.rs#L69). I am not quite sure what `Rev` would do.

So this last point is where things get interesting. I suspect that `Rev` will actually wind up invoking `producer.reverse()` to reverse the producer it was given and pass that along. Since all producers support indexing at arbitrary points, probably they can all support reversing too; but we'll have to implement that method down the line, for all the producer impls (which aren't *that* many).

I'd rather this task "moderate" -- the complexity level is low, but it's some grunt work, and of course if you're not familiar with the system, that'll take a bit of getting used to.

schuster:
Fixes #150, using the approach @nikomatsakis suggested. A few notes on problems mentioned in that issue:
* Producer iterator types are required to implement `ExactSizeIterator` as well as `DoubleEndedIterator` (should be fine since producers are based on exact-size parallel iterators anyway).
* We were able to use `Range` inside `enumerate()` after all: since the iterators usable by `enumerate()` are required to be exact-size, their max index is actually `usize::MAX - 1`.
* I wrote a wrapper around `Chain` to make it an `ExactSizeIterator`. This isn't actually correct if adding the lengths of the iterators overflows, but it will work for most cases. This was already an issue anyway, since the same issue could cause issues for `Chain` as an `IndexedParallelIterator`

### DIFF
--- a/src/par_iter/chain.rs
+++ b/src/par_iter/chain.rs
@@ -211,7 +211,6 @@ impl<A, B> Producer for ChainProducer<A, B>
 
 pub struct Chain<A, B> {
     chain: iter::Chain<A, B>,
-    len: usize,
 }
 
 impl<A, B> Chain<A, B> {
@@ -219,10 +218,8 @@ impl<A, B> Chain<A, B> {
         where A: ExactSizeIterator,
               B: ExactSizeIterator<Item=A::Item>
     {
-        let len = a.len().checked_add(b.len()).expect("overflow");
         Chain {
             chain: a.chain(b),
-            len: len,
         }
     }
 }
@@ -234,14 +231,11 @@ impl<A, B> Iterator for Chain<A, B> // parameterized over the iterator types
     type Item = A::Item;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.len != 0 {
-            self.len -= 1;
-        }
         self.chain.next()
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        (self.len, Some(self.len))
+        self.chain.size_hint()
     }
 }
 
@@ -249,9 +243,6 @@ impl<A, B> ExactSizeIterator for Chain<A, B>
     where A: ExactSizeIterator,
           B: ExactSizeIterator<Item=A::Item>
 {
-    fn len(&self) -> usize {
-        self.len
-    }
 }
 
 impl<A, B> DoubleEndedIterator for Chain<A, B>
@@ -259,9 +250,6 @@ impl<A, B> DoubleEndedIterator for Chain<A, B>
           B: DoubleEndedIterator<Item=A::Item>
 {
     fn next_back(&mut self) -> Option<Self::Item> {
-        if self.len != 0 {
-            self.len -= 1;
-        }
         self.chain.next_back()
     }
 }
--- a/src/par_iter/chain.rs
+++ b/src/par_iter/chain.rs
@@ -163,6 +163,13 @@ impl<A, B> Producer for ChainProducer<A, B>
     where A: Producer,
           B: Producer<Item = A::Item>
 {
+    type Item = A::Item;
+    type IntoIter = Chain<A::IntoIter, B::IntoIter>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        Chain::new(self.a.into_iter(), self.b.into_iter())
+    }
+
     fn weighted(&self) -> bool {
         self.a.weighted() || self.b.weighted()
     }
@@ -199,14 +206,58 @@ impl<A, B> Producer for ChainProducer<A, B>
     }
 }
 
-impl<A, B> IntoIterator for ChainProducer<A, B>
-    where A: Producer,
-          B: Producer<Item = A::Item>
+/// ////////////////////////////////////////////////////////////////////////
+/// Wrapper for Chain to implement ExactSizeIterator
+
+pub struct Chain<A, B> {
+    chain: iter::Chain<A, B>,
+    len: usize,
+}
+
+impl<A, B> Chain<A, B> {
+    fn new(a: A, b: B) -> Chain<A, B>
+        where A: ExactSizeIterator,
+              B: ExactSizeIterator<Item=A::Item>
+    {
+        let len = a.len() + b.len();
+        Chain {
+            chain: a.chain(b),
+            len: len,
+        }
+    }
+}
+
+impl<A, B> Iterator for Chain<A, B> // parameterized over the iterator types
+    where A: Iterator,
+          B: Iterator<Item=A::Item>
 {
     type Item = A::Item;
-    type IntoIter = iter::Chain<A::IntoIter, B::IntoIter>;
 
-    fn into_iter(self) -> Self::IntoIter {
-        self.a.into_iter().chain(self.b)
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.len != 0 {
+            self.len -= 1;
+        }
+        self.chain.next()
+    }
+}
+
+impl<A, B> ExactSizeIterator for Chain<A, B>
+    where A: ExactSizeIterator,
+          B: ExactSizeIterator<Item=A::Item>
+{
+    fn len(&self) -> usize {
+        self.len
+    }
+}
+
+impl<A, B> DoubleEndedIterator for Chain<A, B>
+    where A: DoubleEndedIterator,
+          B: DoubleEndedIterator<Item=A::Item>
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        if self.len != 0 {
+            self.len -= 1;
+        }
+        self.chain.next_back()
     }
 }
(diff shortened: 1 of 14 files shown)


src/par_iter/rev.rs
```diff
@@ -0,0 +1,112 @@
...
+    base: P,
+    len: usize,
+}
+
+impl<P> Producer for RevProducer<P>
+    where P: Producer,
+{
+    type Item = P::Item;
+    type IntoIter = iter::Rev<P::IntoIter>;
+    fn into_iter(self) -> Self::IntoIter {
```
nikomatsakis: Nit: newline here

src/par_iter/internal.rs
```diff
@@ -16,7 +16,12 @@ pub trait ProducerCallback<ITEM> {
 /// A producer which will produce a fixed number of items N. This is
 /// not queryable through the API; the consumer is expected to track
 /// it.
-pub trait Producer: IntoIterator + Send + Sized {
+pub trait Producer: Send + Sized
+{
+    type Item;
+    type IntoIter: Iterator<Item = Self::Item> + DoubleEndedIterator + ExactSizeIterator;
+    fn into_iter(self) -> Self::IntoIter;
```
nikomatsakis: Nit: newline here

src/par_iter/chain.rs
```diff
@@ -199,14 +206,58 @@ impl<A, B> Producer for ChainProducer<A, B>
...
+    chain: iter::Chain<A, B>,
+    len: usize,
+}
+
+impl<A, B> Chain<A, B> {
+    fn new(a: A, b: B) -> Chain<A, B>
+        where A: ExactSizeIterator,
+              B: ExactSizeIterator<Item=A::Item>
+    {
+        let len = a.len() + b.len();
```
nikomatsakis: Nit: maybe we should use `a.len().checked_add(b.len()).expect("overflow")`?

What do you think @cuviper?

src/par_iter/enumerate.rs
```diff
@@ -89,6 +90,17 @@ pub struct EnumerateProducer<P> {
...
 {
+    type Item = (usize, P::Item);
+    type IntoIter = iter::Zip<Range<usize>, P::IntoIter>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        // Enumerate only works for ExactParallelIterators. Since those
+        // have a max length of usize::MAX, their max index is
+        // usize::MAX - 1, so the range 0..usize::MAX includes all
+        // possible indices
+        (self.offset..usize::MAX).zip(self.base.into_iter())
```
nikomatsakis: I think `self.offset..` suffices, in that case, unless you wanted to make the `usize::MAX` clear for stylistic reasons.

cuviper: `RangeFrom` is not exact nor double-ended.

src/par_iter/chain.rs
cuviper: I think it's fine to check, but if there's an overflow I expect we'll first have a problem in `ExactParallelIterator::len()`.  So let's check everywhere it matters, if we're going to check at all.

I wonder here though, do you really need to keep track the length at all?  You're paying a (small) computational cost to update a value that's only occasionally read.  Plus, I don't think you need it anyway, because `ExactSizeIterator::len()` has a default implementation that just checks and uses the value from `Iterator::size_hint()`, which should work fine here if you pass it through `Chain` wrapper.  In fact, the docs say:

> When implementing an `ExactSizeIterator`, You must also implement `Iterator`. When doing so, the implementation of `size_hint()` _must_ return the exact size of the iterator.

... which implies that you can't leave `size_hint()` to the default `(0, None)`.

src/par_iter/internal.rs
cuviper: I don't love this change to an embedded `IntoIterator`, but I see how it's necessary to get the additional trait bounds to propagate implicitly.  It's yet another case of rust-lang/rust#20671, and actually kind of interesting to point out that associated types constraints do get elaborated this way, but they don't if you write:

```rust
pub trait Producer: IntoIterator + Send + Sized
    where <Self as IntoIterator>::IntoIter: DoubleEndedIterator + ExactSizeIterator
{...}
```

*sigh*

src/par_iter/enumerate.rs
nikomatsakis: Oh, is that the problem? OK, seems fine.

src/par_iter/internal.rs
nikomatsakis: Yes, I hope to fix #20671 at some point in the not that distant future. I can't remember if, when I made a branch of this, I also inlined `IntoIter`? Presumably I did.

schuster: Yes, that's where I got the idea from. Otherwise I probably would have done the same as you and spent a long time diagnosing why the compiler wasn't propagating the constraint.

schuster: I'll leave a comment at this point noting that issue, so that hopefully whenever it's resolved we can go back and remove this inlining.

src/par_iter/chain.rs
schuster: @cuviper I'm not sure I understand: are you suggesting that this wrapper's `size_hint` just use use the result from the underlying `std::iter::Chain::size_hint`? The issue I see with that is I don't know whether that underlying value would always be accurate (since the only reason for this wrapper to exist is that the original `Chain` doesn't implement `ExactSizeIterator`).

schuster:
Fixed most of the nits from above; just need clarity on the length issue for the `Chain` wrapper.

src/par_iter/chain.rs
cuviper: AFAICS you didn't write a `size_hint` at all, which the docs say you must for `ExactSizeIterator`.  Then the default `ExactSizeIterator::len()` asserts that the lower and upper bounds are actually equal.  You can override `len` for performance of you know it's always accurate, but in this case I'd just defer to `Chain::size_hint` and let it assert.

schuster: Ah okay, so if the underlying `Chain` ever returns an imprecise result, the assertion will at least let us know. Sounds good enough to me.

schuster:
Length issue now fixed as well.

nikomatsakis:
Thanks @schuster !

PR #226, opened 2017-02-26
Add a new `rayon::split` to create arbitrary parallel iterators

cuviper:
This takes two arguments, the data and a closure to split it, and
implements a `ParallelIterator`.  The data will be divided using the
usual `par_iter` splitting heuristics, then may be mapped, reduced, etc.
however is needed.

The `fibonacci` and `quicksort` demos have been updated with `split`
examples, where the former splits `n` into `n-2` and `n-1`, and the
latter performs a sorting partition before splitting its slice.

### DIFF
--- a/rayon-demo/src/fibonacci/mod.rs
+++ b/rayon-demo/src/fibonacci/mod.rs
@@ -22,18 +22,28 @@ const INIT_FAILED: &'static str = "Rayon failed to initialize";
 const N: u32 = 32;
 const FN: u32 = 2178309;
 
+fn fib_iterative(n: u32) -> u32 {
+    let mut a = 0;
+    let mut b = 1;
+    for _ in 0..n {
+        let c = a + b;
+        a = b;
+        b = c;
+    }
+    a
+}
+
+fn fib_recursive(n: u32) -> u32 {
+    if n < 2 { return n; }
+
+    fib_recursive(n - 1) + fib_recursive(n - 2)
+}
+
 
 #[bench]
 /// Compute the Fibonacci number recursively, without any parallelism.
 fn fibonacci_recursive(b: &mut test::Bencher) {
-
-    fn fib(n: u32) -> u32 {
-        if n < 2 { return n; }
-
-        fib(n - 1) + fib(n - 2)
-    }
-
-    b.iter(|| assert_eq!(fib(test::black_box(N)), FN));
+    b.iter(|| assert_eq!(fib_recursive(test::black_box(N)), FN));
 }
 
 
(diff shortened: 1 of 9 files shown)


cuviper:
Benchmark results on my dual core (4-cpu) laptop:

```
test fibonacci::fibonacci_iterative                          ... bench:           6 ns/iter (+/- 0)
test fibonacci::fibonacci_join_1_2                           ... bench:  96,867,939 ns/iter (+/- 35,786,979)
test fibonacci::fibonacci_join_2_1                           ... bench:  93,908,533 ns/iter (+/- 36,924,257)
test fibonacci::fibonacci_recursive                          ... bench:  13,495,680 ns/iter (+/- 686,026)
test fibonacci::fibonacci_split_iterative                    ... bench:      34,358 ns/iter (+/- 7,128)
test fibonacci::fibonacci_split_recursive                    ... bench:   6,598,017 ns/iter (+/- 434,233)

test quicksort::bench::quick_sort_par_bench                  ... bench:  20,788,288 ns/iter (+/- 1,349,063)
test quicksort::bench::quick_sort_seq_bench                  ... bench:  40,737,106 ns/iter (+/- 5,660,104)
test quicksort::bench::quick_sort_splitter                   ... bench:  21,033,429 ns/iter (+/- 4,943,563)
```

src/par_iter/internal.rs
```diff
@@ -132,7 +132,7 @@ pub trait UnindexedProducer: Send + Sized {
     type Item;
 
     /// Split midway into a new producer if possible, otherwise return `None`.
-    fn split(&mut self) -> Option<Self>;
+    fn split(self) -> (Self, Option<Self>);
```
cuviper: This `UnindexedProducer::split` interface change was because I found it awkward to deal with lifetimes in quicksort with extra `&mut` indirection, whereas it was much easier to deal with by value.

cuviper:
I fixed the imports for 1.12.0 CI.  Note that nightly also failed on the first attempt with a stack overflow, but succeeded on the second -- heisenbug?

src/par_iter/splitter.rs
```diff
@@ -0,0 +1,66 @@
+use super::internal::*;
+use super::*;
+
+/// The `split` function takes arbitrary data and a closure that knows how to
+/// split it, and turns this into a `ParallelIterator`.
```
nikomatsakis: Should we mark this as `features = unstable`? Also, I think the docs could use with an example or two at least :)

nikomatsakis: (I'm not sure of what the "unstable" *policy* is -- probably somewhat we should chat about. Maybe rather than a cargo feature, just some doc comments might be good.)

nikomatsakis:
@cuviper interesting. I could imagine that happening, actually, in that the stack depth can be affected by how things get stolen.

nikomatsakis:
@cuviper seems like some performance measurements would be prudent, seeing as how this touches a core API, but I'm basically r+.

nikomatsakis:
Measurements look good!

- https://people-mozilla.org/~nmatsakis/rayon-pr226-points.svg (Raw data)
- https://people-mozilla.org/~nmatsakis/rayon-pr226-points-medians.svg (Medians normalized to 100)

Nothing appears to have gotten slower =)

nikomatsakis:
@cuviper if you rebase, r=me, so feel free to merge

PR #227, opened 2017-02-28
extract the "core APIs" into a rayon-core crate

nikomatsakis:
The parallel iterator APIs remain in the rayon crate.  This allows the rayon crate to advance a major version without introducing two thread-pools.

Friendly competition to https://github.com/nikomatsakis/rayon/pull/209.

Questions:

- [x] How should the `rayon-core` docs refer to APIs in code examples? i.e., should we write `rayon::spawn` (with `rayon_core` aliased to `rayon`), `rayon_core::spawn`, or just `spawn`? 

cc @cuviper


nikomatsakis:
@cuviper ok, finally got travis happy. (Couldn't see the logs before, which was a bit of an obstacle.)

src/lib.rs
```diff
@@ -7,47 +7,15 @@
...
-pub use thread_pool::ThreadPool;
-pub use join::join;
-pub use scope::{scope, Scope};
-#[cfg(feature = "unstable")]
-pub use spawn_async::spawn_async;
-#[cfg(feature = "unstable")]
-pub use spawn_async::spawn_future_async;
-#[cfg(feature = "unstable")]
-pub use future::RayonFuture;
+pub use rayon_core::*;
```
cuviper: I'd prefer explicit re-exports for two reasons:

1. It makes the API surface of a particular `rayon` version consistent, rather than depending on what `rayon-core` lies beneath.

2. It works better with rustdoc, which only writes globs as a single re-exports link, whereas direct `pub use rayon_core::foo` creates inline documentation.

nikomatsakis: yeah, I had a feeling this was probably a bad idea ... I figured we'd never want a discrepancy, but the rustdoc thing is definitely a KO argument

Cargo.toml
```diff
@@ -7,12 +7,15 @@ license = "Apache-2.0/MIT"
 repository = "https://github.com/nikomatsakis/rayon"
 documentation = "https://docs.rs/rayon/"
 
+[workspace]
+members = ["rayon-demo", "rayon-core"]
+
 [dependencies]
+rayon-core = { path = "rayon-core" }
```
cuviper: Note, you'll have to version this before it can be published.  And this is what we'll have to raise any time there's a new `rayon-core` interface that `rayon` want to use/reexport.

nikomatsakis: oh? I thought cargo rewrote `path` dependencies?

nikomatsakis: Maybe I'm thinking of something else.

cuviper: The path-dep is in effect for local builds, but when you publish to crates.io you have to explicitly say the version of `rayon-core` to depend on.  So you do both:

    rayon-core = { path = "rayon-core", version = "0.1" }

nikomatsakis:
@cuviper ok I added a version dependency and removed `pub use rayon_core::*`

src/par_iter/internal.rs
```diff
@@ -6,7 +6,6 @@
 use join;
```
cuviper: 1.12 CI says:

```
error[E0432]: unresolved import
 --> src/par_iter/internal.rs:6:5
  |
6 | use join;
  |     ^^^^
```

nikomatsakis: bother. seems like a bug in 1.12, but ... whatever.

Issue #228, opened 2017-02-28
cleanup public surface of parallel iterator API


nikomatsakis:
The parallel iterator public API contains a whole bunch of random modules and things (e.g., for each operation, like `reduce`). We should clean it up to be more minimal, only exporting the stuff we really need, and putting all the associated types in a flat space (probably still keeping the `par_iter` module though).

### Current status

Much progress has been made in #247.

This established some principles:

- the `par_iter` module is now called `iter`; it contains only things specific to the traits themselves
- parallel iterator types for existing collections and types (e.g., for slices) are named `FooIter` (e.g., `SliceIter`) and placed in top-level modules (e.g., `rayon::slice::SliceIter`)
- we generally are not exposing the means to "re-implement" the combinators like `map` unless we have a reason to do so (e.g., there is no public constructor on `iter::map::Map`, nor free function version of the `map()` method). This is because those methods are not expected to be overridden (and, if they are, it's unlikely that `Map` would be of use). We can revisit this if there is demand.

### Open questions to be resolved

- [x] What should we call `iter::internal` and should it move somewhere else in the hierarchy?
- [x] What should we do about the `MapOp` hierarchy in `iter`?
- [x] What should we do about the `ReduceOp` hierarchy in `iter`?
- [x] Should we make the `ChunksIter` stuff less general, more like `ParallelString`?
    - right now, its a trait that acts as though many types might implement it, though only `[]` does so, and it's not clear it's general enough for other things yet

Kerollmops:
I think that rewriting [Type Parameters to be more concise](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html) can be a good thing, no ?

nikomatsakis:
@Kerollmops I ... obviously strayed from the style guidelines intentionally, but I think I've decided that the longer names are not so readable. CamelCase I am actively against, though, because I tried it and found it kind of confusing (it was hard to tell which is a type parameter and which is a type).

nikomatsakis:
That said, it's not part of the public API.

nikomatsakis:
(In the sense that it's not a breaking change if we change them later.)

nikomatsakis:
@Kerollmops see https://github.com/nikomatsakis/rayon/issues/249

nikomatsakis:
@cuviper I moved the comments of yours that I did not address in #247 into a list of "unresolved questions" at the top of this issue.

cuviper:
[...]

PR #247, opened 2017-03-03
Simplify public API of the `par_iter` module

nikomatsakis:
This is a first stab at simplifying the public API of the `par_iter` module. Here are the rules I used:

- Expose everything "flat" in the `par_iter` module except for the `internal` traits (`Producer`, etc). 
- Expose only those things that appear in the public interface of the `ParallellIterator` traits.
    - In particular, things like `EnumerateProducer` do not need to be exposed, because the `ExactParallelIterator` trait uses the "double indirection" so that it is never named.
    - To aid in this goal, I adopted the convention of never using `foo::` in public API surface, so that I am forced to `pub use, and **only** using `foo::` inside method bodies, so that I never need to `use`.

I'm pretty happy with how it turned out, but I did run into some questions:

- [x] What should we call the `internal` module? That name seems not great.
- [x] Should the internal APIs remain `#[doc(hidden)]`?
- [x] Do we care that the `new` methods on types like `Enumerator` are publicly surfaced? We could avoid this with more effort (e.g., use a free-function).
- [x] Do we want to expose the "free function" form of things like `find()` (e.g., `find::find()`). I am inclined to say "no" unless there is a demand. Similarly for things like the producer types.
- [x] Naming convention: I used `Combinator` if it is a parallel iterator combinator (e.g., `Enumerate`, `Zip`) and `CollectionIter` if is a `ParallelIterator` implementation for a collection (e.g., `SliceIter`, `RangeIter`). Before we had a mix. This seems to be in the *spirit* of the libstd conventions. However, I am not crazy about the `CollectionIter` convention. I also considered `par_iter::collection::ParIter`, or `CollectionParIter`. Thoughts?


nikomatsakis:
cc #228 -- this is at least a partial fix, and maybe a complete one.

cc @cuviper

cuviper:
> What should we call the internal module? That name seems not great.

I agree it's not great, but I don't have better ideas. `imp` (implementor)? `low` (as in low-level)?

> Should the internal APIs remain #[doc(hidden)]?

No, we need to show them IMO, but we should make sure they're at the end of the list.

> Do we care that the new methods on types like Enumerator are publicly surfaced? We could avoid this with more effort (e.g., use a free-function).

It's not a *big* deal, but I would prefer a free function.

> Naming convention

I think your choice is fine, and the consistency is good.

src/par_iter/mod.rs
```diff
@@ -14,52 +14,69 @@
...
+mod slice;
+pub use self::slice::{SliceIter, ChunksIter};
+mod slice_mut;
+pub use self::slice_mut::{SliceIterMut, ChunksMutIter};
+mod string;
+pub use self::string::{ParallelString, Pattern, ParChars, ParSplit,
+                       ParSplitTerminator, ParLines,
+                       ParSplitWhitespace};
+mod map;
+pub use self::map::{Map, MapOp, MapFn, MapCloned, MapInspect};
```
cuviper: I don't love having this `Map` hierarchy exposed.  Can we instead have specific `Cloned`, `Inspect`, `Map` (formerly `MapFn`) types, and hide the common `Map` in implementation?

src/par_iter/mod.rs
```diff
@@ -14,52 +14,69 @@
...
-pub mod vec;
-pub mod option;
-pub mod collections;
-pub mod noop;
-pub mod rev;
+mod for_each;
+mod fold;
+pub use self::fold::Fold;
+mod reduce;
+pub use self::reduce::{ReduceOp, SumOp, ProductOp};
```
cuviper: I don't love the `ReduceOp` hierarchy either, but I think we might be able to do away with the whole lot in #252.

src/par_iter/mod.rs
```diff
@@ -14,52 +14,69 @@
...
+mod take;
+pub use self::take::Take;
+mod slice;
+pub use self::slice::{SliceIter, ChunksIter};
+mod slice_mut;
+pub use self::slice_mut::{SliceIterMut, ChunksMutIter};
+mod string;
+pub use self::string::{ParallelString, Pattern, ParChars, ParSplit,
+                       ParSplitTerminator, ParLines,
+                       ParSplitWhitespace};
```
cuviper: Should these get `String`-specific names somehow?  Or maybe this is a large enough chunk of functionality to actually deserve a submodule, just `pub mod string;` and be done.

src/par_iter/mod.rs
```diff
@@ -14,52 +14,69 @@
...
+mod reduce;
+pub use self::reduce::{ReduceOp, SumOp, ProductOp};
+mod skip;
+pub use self::skip::Skip;
+mod take;
+pub use self::take::Take;
+mod slice;
+pub use self::slice::{SliceIter, ChunksIter};
+mod slice_mut;
+pub use self::slice_mut::{SliceIterMut, ChunksMutIter};
```
cuviper: I mentioned I wanted to prototype making the slice extensions to look more like `ParallelString` -- if we do that, I think a `pub mod slice` makes sense too.

cuviper:
While I'm thinking of the module layout, we could flatten this too:

- `rayon`
    - `prelude` should be emphasized
    - `iter`  (`par_` is somewhat redundant)
    - `slice`
    - `string`
    - `internal` (wearing its new bikeshed color)

I think having modules as siblings is cleaner, and also makes a little bit of a test that the necessary functionality is all public, whereas submodules get to cheat a little.

cuviper:
FWIW, I'd be happy with iterating these points over multiple PRs, especially since one massive PR will be disruptive to any other changes we might be considering in the meantime.  So if you want, I say go ahead and merge this as a partial fix for #228, to be continued...

nikomatsakis:
@cuviper ok, I agree, I think I'll apply some of your changes, but land it roughly as-is, and we can then discuss the precise setup on #228.

nikomatsakis:
@cuviper 

ok, so I adjusted the module hierarchy more-or-less as you requested. The one thing I did not do is move or rename `internal` -- I think I prefer it to be caller `rayon::iter::internal` rather than just `rayon::internal`, since its contents are basically specific to the parallel iterator traits, right? That is, I view these as being "part of" the parallel iterator traits.

I also left `MapOp`, `ReduceOp`, and friends as they are.

So the current module hierarchy becomes:

- rayon
    - iter
    - option
    - prelude
    - range
    - slice
    - string
    - vec

nikomatsakis:
@cuviper ok, I think I addressed most everything. I'm going to merge this, presuming that the CI is happy.

Note: I did not leave "deprecated" paths for things that were moved around. That's a pain and I don't know how to do it. =)

Issue #246, opened 2017-03-03
[docs] the `Take` struct needs docs


nikomatsakis:
The `Take` struct implements the parallel iterator `take()` combinator. It needs some simple docs that say precisely that and link to `take()` method for more information.

Issue #245, opened 2017-03-03
[docs] the `Skip` struct needs documentation


nikomatsakis:
The `Skip` struct implements the parallel iterator `skip()` combinator. It needs some simple docs that say precisely that and link to `skip()` method for more information.

nikomatsakis:
Done in #248.

Issue #243, opened 2017-03-03
[docs] the `Fold` struct needs documentation


nikomatsakis:
The `Fold` struct implements the parallel iterator `fold()` combinator. It needs some simple docs that say precisely that and link to `fold()` method for more information.

nikomatsakis:
Done in #248.

Issue #242, opened 2017-03-03
[docs] the `FromParallelIterator` trait needs documentation


nikomatsakis:
`FromParallelIterator` is used to drive the `collect()` method. It defines collections that can be created by parallel iterators. Probably it just needs to say that and point to `collect()`.

nikomatsakis:
Done in #248.

PR #248, opened 2017-03-03
Add simple docs to various iterators with links.

leshow:
closes: #246 #245 #243 #242 #241 #240 #239 #238 #237

### DIFF
--- a/src/par_iter/from_par_iter.rs
+++ b/src/par_iter/from_par_iter.rs
@@ -5,6 +5,14 @@ use std::hash::{BuildHasher, Hash};
 use std::collections::LinkedList;
 use std::collections::{BinaryHeap, VecDeque};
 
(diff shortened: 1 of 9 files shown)


src/par_iter/from_par_iter.rs
```diff
@@ -5,6 +5,11 @@ use std::hash::{BuildHasher, Hash};
 use std::collections::LinkedList;
 use std::collections::{BinaryHeap, VecDeque};
 
+/// `FromParallelIterator` implements the conversion to a [`ParallelIterator`].
```
cuviper: This is a conversion *from* a parallel iterator, and it should be mentioned that this is used by the `collect` method.

leshow: Thanks for that, I've made that change and added the link to `collect()`

nikomatsakis:
This is awesome. Thanks!

Issue #232, opened 2017-03-03
[docs] `rayon_wait()` method on futures


nikomatsakis:
Write `/// foo` documentation for the `rayon_wait()` method on the `RayonFuture` type in `rayon-core`.

This method will wait for the future to complete. Unlike the standard `Future::wait` wait, it is integrated with the rayon thread-pool, and hence using `rayon_wait()` will not cause dead-locks. However, it should still be used with caution, as it forces rayon to keep your current stack frame blocks until the future completes, which may not be the most efficient use of resources. It is typically better to create and spawn a new future (i.e., instead of `let v = future.rayon_wait(); use(v)` do `scope.spawn_future(future.and_then(|v| use(v))`).

nikomatsakis:
I think this does not have to be on the 1.0 milestone, given that (a) the method has problems and (b) futures have been extracted.

nikomatsakis:
In fact, i'm inclined to close this issue altogether.

PR #250, opened 2017-03-04
Added the first-pass at `rayon_wait` method documentation

petezhut:
Figured I'd take a stab at adding some doc strings.  Should be a decent attempt at closing out #232

### DIFF
--- a/rayon-core/src/future/mod.rs
+++ b/rayon-core/src/future/mod.rs
@@ -82,6 +82,11 @@ pub fn new_rayon_future<'scope, F, S>(future: F, scope: S) -> RayonFuture<F::Ite
     }
 }
 
+/// Implements a wait method which should not cause race conditions.
+/// Caution should still be used since it is not the most optimum resource
+/// usage.  It is typically better to create and spawn a new future.
+/// Recommend using:
+///   scope.spawn_future(future.and_then(|v| use(v))
 impl<T, E> RayonFuture<T, E> {
     pub fn rayon_wait(mut self) -> Result<T, E> {
         // NB: End-user API!


rayon-core/src/future/mod.rs
```diff
@@ -82,6 +82,11 @@ pub fn new_rayon_future<'scope, F, S>(future: F, scope: S) -> RayonFuture<F::Ite
     }
 }
 
+/// Implements a wait method which should not cause race conditions.
```
nikomatsakis: A couple of nits about this sentence:

- It assumes that the reader knows what "a wait method" is; I think it might be better to start off with text similar to [Future::wait][w]. 
- The problem with `Future::wait` is not race conditions but rather that invoking it can cause deadlocks.

I'm torn between something like:

A version of `Future::wait` that is safe for use in the Rayon threadpool.

And something more informative like:

Blocks the current thread until this future is resolved, returning the value. This function is a version of [`Future::wait`][w] that is safe for use from within the Rayon thread-pool. The `Future::wait()` function does not coordinate with Rayon's scheduler and so calling it may lead to panics or deadlocks.

[w]: https://docs.rs/futures/0.1.8/futures/future/trait.Future.html#method.wait

rayon-core/src/future/mod.rs
```diff
@@ -82,6 +82,11 @@ pub fn new_rayon_future<'scope, F, S>(future: F, scope: S) -> RayonFuture<F::Ite
...
+/// Caution should still be used since it is not the most optimum resource
+/// usage.  It is typically better to create and spawn a new future.
+/// Recommend using:
+///   scope.spawn_future(future.and_then(|v| use(v))
```
nikomatsakis: Here the formatting doesn't quite seem right. Also, the example should be something that can actually execute, since our tests are actively tested by rustdoc. Perhaps something like this (untested):

    It is typically better to create and spawn a new future using `and_then`, like so:
    
    ```rust
    extern crate futures;
    scope(|s| {
        let future = s.spawn(futures::lazy(|| -> Result<usize, ()> {
            // insert expensive computation here
            Ok(22)
        });

        // instead of `let x = future.rayon_wait()`, try this:
        let future = s.spawn(future.and_then(|x| -> Result<usize, ()> {
            // code that executes after `future` is complete;
            // `x` is the result value
            Ok(x * 2)
        });
    ```
```

Issue #233, opened 2017-03-03
[docs] document the `Scope` struct


nikomatsakis:
The `scope::Scope` struct in `rayon-core` needs documentation. The use of scopes is reasonably well documented on the `scope::scope` function, so probably the type itself just needs to say something like "Represents a fork-join scope which can be used to spawn any number of tasks. See `scope()` for more information."

dethoter:
Can this issue be closed?

cuviper:
Ah, yes, thanks!

PR #251, opened 2017-03-04
Added the exact doc string recommended in the comment

petezhut:
Just added the struct docstring for #233

### DIFF
--- a/rayon-core/src/scope/mod.rs
+++ b/rayon-core/src/scope/mod.rs
@@ -16,7 +16,7 @@ use unwind;
 mod test;
 
 /// Represents a fork-join scope which can be used to spawn any number 
-/// of tasks. See scope() for more information.
+/// of tasks. See [`Scope()`] for more information.
 pub struct Scope<'scope> {
     /// thread where `scope()` was executed (note that individual jobs
     /// may be executing on different worker threads, though they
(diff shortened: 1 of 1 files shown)


leshow:
You've got to add an internal link to get `scope()` to link properly. I recently did a bunch of these, have a look at how to do it:

https://github.com/leshow/rayon/commit/c4928f73efe15f5f4b538a518455d719d5b8575f

petezhut:
I'll update it!

rayon-core/src/scope/mod.rs
```diff
@@ -15,6 +15,8 @@ use unwind;
 #[cfg(test)]
 mod test;
 
+/// Represents a fork-join scope which can be used to spawn any number 
+/// of tasks. See [`Scope()`] for more information.
```
nikomatsakis: This isn't quite right. For one thing, the name of the function is not capitalized, and you also need a link. Something like this is probably right:

    See [`scope()`][s] for more information.
    
    [s]: fn.join.html

petezhut: Ah.  I'm very unfamiliar with whatever parsing engine you are planning to use for doc generation, so this is a brand new experience.  I'll give it another shot!

nikomatsakis: @petezhut it's just markdown; you can check the output locally by running `cargo doc`, I think

PR #253, opened 2017-03-10
Fix compile-fail tests and broaden CI

cuviper:


cuviper:
We weren't running the compile-fail tests, nor any rayon-core tests.  😢

nikomatsakis:
d'oh

PR #254, opened 2017-03-10
Minor edits to description of work-stealing in README

mbrubeck:
When I was first learning about work stealing, the mention of "a central queue" here confused me.  I hope this expanded description is clearer.

[Rendered](https://github.com/mbrubeck/rayon/blob/docs/README.md#how-it-works-work-stealing)

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,15 +190,15 @@ and dynamically ascertain how much parallelism is available and
 exploit it. The idea is very simple: we always have a pool of worker
 threads available, waiting for some work to do. When you call `join`
 the first time, we shift over into that pool of threads. But if you
-call `join(a, b)` from a worker thread W, then W will place `b` into a
-central queue, advertising that this is work that other worker threads
-might help out with. W will then start executing `a`. While W is busy
-with `a`, other threads might come along and take `b` from the
-queue. That is called *stealing* `b`. Once `a` is done, W checks
-whether `b` was stolen by another thread and, if not, executes `b`
-itself. If `b` *was* stolen, then W can just wait for the other thread
-to finish.  (In fact, it can do even better: it can go try to find
-other work to steal in the meantime.)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -190,15 +190,16 @@ and dynamically ascertain how much parallelism is available and
...
+call `join(a, b)` from a worker thread W, then W will place `b` into
+its work queue, advertising that this is work that other worker
+threads might help out with. W will then start executing `a`.
+
+While W is busy with `a`, other threads might come along and take `b`
+from its queue. That is called *stealing* `b`. Once `a` is done, W
+checks whether `b` was stolen by another thread and, if not, executes
+`b` itself. If `b` *was* stolen, then W can just move on to the next
+job it ins queue.  If its queue is empty, it will go look at all the
+other thread's queues and try to steal work from them.
```
cuviper: The changes of the last two sentences are subtly incorrect.  `W` process its queue from the front, and other threads steal from the back.  So if `b` was stolen, this directly implies that the queue is already empty, because everything prior to `b` must have been stolen too.  There's never anything for `W` to move on to in its queue -- it always becomes a stealer at this point until `b` is done.

mbrubeck: Fixed, thanks!

cuviper:
After my first comment, I'm actually confusing myself now, because with `scope` it is in fact possible to push jobs on top of `b`, which may or may not get executed before `b` is completed and returned.  But I think the way you reworded your update is still OK.

The CI failure is the stack overflow we've been encountering.  It comes and goes -- very frustrating...

nikomatsakis:
@cuviper I've been thinking of integrating @alexcrichton's [stacker crate](https://docs.rs/stacker/0.1.2/stacker/) to grow the stack as needed (perhaps as something one can configure -- i.e., pick a stack size, with the default being "dynamic").

PR #258, opened 2017-03-11
add stack_size parameter to Configuration (fixes #179)

### LINKED ISSUE #179: ability to set stack size for worker threads
Similar to #142, which concerns thread names, can the `stack_size` of `std::thread::Builder` be exposed through the `Configuration` to set the stack size for rayon's worker threads?

manuel-woelker:
Unfortunately I haven't been able to find a neat way to unit test this.
My first attempt was to force a stack overflow via recursion, but that segfaults the entire process (less than ideal for a unit test).
Using `getrlimit()` only seemed to return the system wide limits.
It might be possible to do something with pthread functions...

I checked via valgrind using the following command:

`valgrind --trace-children=yes --tool=drd --show-stack-usage=yes cargo test --lib -- --nocapture stack_size 2>&1 | grep stack`

Fixes #179

### DIFF
--- a/tests/run-pass/stack_overflow_crash.rs
+++ b/tests/run-pass/stack_overflow_crash.rs
@@ -1,0 +1,51 @@
+extern crate rayon;
+
+use rayon::*;
+
+use std::process::{self, Command};
+use std::env;
+
+#[cfg(target_os = "linux")]
+use std::os::unix::process::ExitStatusExt;
+
+
+
+fn force_stack_overflow(depth: u32) {
+    let buffer = [0u8; 1024*1024];
+    if depth > 0 {
+        force_stack_overflow(depth - 1);
+    }
+}
+
+fn main() {
+    if env::args().len() == 1 {
+        // first check that the recursivecall actually causes a stack overflow, and does not get optimized away
+        {
+            let status = Command::new(env::current_exe().unwrap())
+                .arg("8")
+                .status()
+                .unwrap();
+            assert_eq!(status.code(), None);
+            #[cfg(target_os = "linux")]
+            assert!(status.signal() == Some(11 /*SIGABRT*/) || status.signal() == Some(6 /*SIGSEGV*/));
+        }
+
+
+        // now run with a larger stack and verify correct operation
+        {
+            let status = Command::new(env::current_exe().unwrap())
+                .arg("48")
+                .status()
+                .unwrap();
+            assert_eq!(status.code(), Some(0));
+            #[cfg(target_os = "linux")]
+            assert_eq!(status.signal(), None);
+        }
+    } else {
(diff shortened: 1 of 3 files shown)


cuviper:
Maybe something in tests/run-fail?

manuel-woelker:
The description in 'tests/run-fail' says that the tests are "expected to run and panic", whereas a stack overflow just segfaults instead of panicking.

nikomatsakis:
@manuel-woelker one idea would be to have a run-pass test that spawns itself giving different arguments (e.g., size of stack, amount of data to allocate on stack) and then check that the bad combinations fail with segfault or whatever. Here is an [example using this strategy from rust-lang/rust](https://github.com/rust-lang/rust/blob/master/src/test/run-pass/weird-exit-code.rs)

manuel-woelker:
@nikomatsakis @cuviper Thanks for the hints. I made a test verifying proper stack sizing as well as that a stack overflow can actually be caused (and is not optimized away).

tests/run-pass/stack_overflow_crash.rs
```diff
@@ -0,0 +1,50 @@
...
+    if env::args().len() == 1 {
+        // first check that the recursivecall actually causes a stack overflow, and does not get optimized away
+        {
+            let status = Command::new(env::current_exe().unwrap())
+                .arg("8")
+                .status()
+                .unwrap();
+            assert_eq!(status.code(), None);
+            #[cfg(target_os = "linux")]
+            assert_eq!(status.signal(), Some(6));
```
cuviper: How about `libc::SIGABRT`?  Same value, but more clear about what we're looking for.

manuel-woelker:
Thanks for the feedback. Added the `SIGABRT` and rebased onto master.

nikomatsakis:
@manuel-woelker thanks; I'm ready to merge, I'm just a bit frustrated by the fact that CI can't give me any feedback. Presumably this passes your tests locally?

cuviper:
OK, restarting the build got newer nightlies and got past the cargo errors, and now it hit the general stack overflows we've been fighting.  (not the stack overflows you're _trying_ to test. :wink:)  Can you rebase onto master to pick up the new default stack size?

manuel-woelker:
In the nightly build the signal triggered by the stack overflow is SIGSEGV (6) rather than SIGABRT (11).
I think it might be hitting guard pages. I changed the test to accept either.

Almost got it working now, only the unstable build complains about the use of the unstable `libc` feature.
My best guess is that the compiletest somehow ignores the libc crate...
cf. https://travis-ci.org/nikomatsakis/rayon/builds/213544997

Since there is no actual code involved, I used (commented) magic constants again for now.

nikomatsakis:
Seems good enough to me. Thanks @manuel-woelker!

PR #260, opened 2017-03-12
Replace weights with split length control

### LINKED ISSUE #111: deprecate existing weight APIs

cuviper:
This adds `with_min_len` and `with_max_len` on `IndexedParallelIterator` to control the length of the splits that will be used in each thread.  Rayon will try to split iterators to lengths within this range, without making guarantees about where in the range it will fall.  If the min and max are at odds, the min wins.
(e.g. with any `2 * min > max`, say min=10 and max=15, we can't split length=16 further.)

Fixes #111?

TODO:

- [x] Add targeted testing
- [ ] ~~Add "representative" benchmarks~~
- [x] Leave deprecated no-op placeholders for the old weight APIs?


src/iter/internal.rs
```diff
@@ -166,6 +172,34 @@ impl Splitter {
...
+
+impl LengthSplitter {
+    #[inline]
+    fn new(min: usize, max: usize, len: usize) -> LengthSplitter {
+        let mut splitter = LengthSplitter {
+            inner: Splitter::new(),
+            min: cmp::max(min, 1),
+        };
+
+        let max_splits = len / cmp::max(max, 1);
```
nikomatsakis: I feel this logic merits a comment =)

nikomatsakis: Ideally, with a few examples that show how different min/max/len combinations work out

nikomatsakis: I'm a bit confused by the `max_splits` computation here, actually. So in particular if the default is that `max` is `usize::MAX`, then wouldn't `max_splits` almost always be `0`? I guess tomorrow I'll have to trace this out in more detail.

cuviper: Perhaps it's poorly named - it represents the number of splits derived from the max, and just below it's only used if it's more than the normal splitter would have used.

I'll fill in a comment later. Thanks for looking it over!

cuviper:
OK, added some comments, and flipped the name of that variable.  Hope it makes sense now! :)

I still need to address the TODOs, but I threw in some quick additions to the join benchmarks.

```
test join_microbench::increment_all                          ... bench:      35,475 ns/iter (+/- 2,704)
test join_microbench::increment_all_atomized                 ... bench:   1,460,769 ns/iter (+/- 21,970)
test join_microbench::increment_all_max                      ... bench:      67,500 ns/iter (+/- 14,403)
test join_microbench::increment_all_min                      ... bench:      20,404 ns/iter (+/- 1,771)
test join_microbench::increment_all_serialized               ... bench:      34,150 ns/iter (+/- 1,349)
```

At least `set_min_len()` shows some value here.  I'll have to think about how to use max better, but my thought was that some workloads may want to be kept within the cache size, for instance.

src/iter/internal.rs
```diff
@@ -133,7 +133,14 @@ pub trait UnindexedProducer: Send + Sized {
 /// job is actually stolen into a different thread.
 #[derive(Clone, Copy)]
 struct Splitter {
+
```
nikomatsakis: Nit: extra newline? Is this what rustfmt produces?

src/iter/internal.rs
```diff
@@ -160,42 +167,68 @@ impl Splitter {
...
     fn new(min: usize, max: usize, len: usize) -> LengthSplitter {
         let mut splitter = LengthSplitter {
             inner: Splitter::new(),
             min: cmp::max(min, 1),
         };
 
-        let max_splits = len / cmp::max(max, 1);
-        if max_splits > splitter.inner.splits {
-            splitter.inner.splits = max_splits;
+        // Divide the given length by the max working lenght to get the minimum
```
nikomatsakis: Nit: "lenght"

nikomatsakis:
This all seems quite good. I do have one bikeshed-y question: the name `set_min_len` suggests to me an `&mut self` "setting" operation, but in fact it produces a wrapper. Maybe just `min_len(N)`? I guess it's not a verb, which is odd. I also thought of `with_min_len(N)`? 

I can't really find a comparable thing in the existing iterator API to guide the naming here.

nikomatsakis:
With respect to this question:

> Leave deprecated no-op placeholders for the old weight APIs?

Probably that would be the nice thing to do, but we should be sure to file an issue to remove them before we release the real 1.0.

src/iter/internal.rs
cuviper: Nah, that's my doing.  I go back and forth on preferring a blank line before doc comments, but I'll remove it if you find this nit-worthy. :)

src/iter/internal.rs
cuviper: ack

cuviper:
I added some tests, and renamed to `with_min_len` / `with_max_len`.

Incidentally, I tried `--jobs 1` on the demo tests to see if that would avoid the stack overflows, but it didn't help, so I reverted it.  It sure would be nice to know what's causing that...

cuviper:
The benchmark todo comes from your emphasized requirement in #111.  If you're satisfied as-is, so am I. 😀

nikomatsakis:
@cuviper heh, fair. Well, I'm not sure what such a benchmark would be just now, but at least we have an unrepresentative one. =)

Actually, I think that the "do a lot of tiny tiny work" isn't such a bad thing to measure, I guess.

cuviper:
I struck out the benchmark TODO, as I think we agreed on gitter that we don't really need to worry about that.  Ready to merge?

nikomatsakis:
@cuviper yep -- my only concern was the "merge from master" commit, did you want to rebase instead?

cuviper:
Merge commits don't bother me, but I can rebase later today if you prefer.

cuviper:
Rebased.

Issue #249, opened 2017-03-04
change type parameter names to be a single letter


nikomatsakis:
I think that the long names for type parameters (like `MAP_OP`) have not proven to be a boon to readability as I had hoped. At this point, I think we should just use single letters (like `M`), with documentation for anything if the role is not clear or needs to be understood by end-users.

I'm going to put this in the 1.0 milestone, but it could be removed since it's not a breaking change to do later. Just a polish thing.

leshow:
I can take some of these, after looking at the code though, I'd like to suggest using `F` instead of `M`, as `M` is already used as a type parameter in a lot of these, for example:

```
pub struct FlatMap<M: ParallelIterator, F> {
    base: M,
    map_op: F,
}
```

nikomatsakis:
@leshow 👍

nikomatsakis:
This is basically done in #261

PR #261, opened 2017-03-15
Shorten type param names

leshow:
I picked the most obvious letters to me. Mapping operations; `F` for function, for anything that returns a bool; `P` for predicate, etc. I made sure to change the associated lifetimes to match up and fix any formatting changes. Let me know if there's anything you'd like to change. 

#249

### DIFF
--- a/src/iter/filter_map.rs
+++ b/src/iter/filter_map.rs
@@ -7,16 +7,16 @@ use super::*;
 ///
 /// [`filter_map()`]: trait.ParallelIterator.html#method.filter_map
 /// [`ParallelIterator`]: trait.ParallelIterator.html
-pub struct FilterMap<M: ParallelIterator, P> {
-    base: M,
+pub struct FilterMap<I: ParallelIterator, P> {
+    base: I,
     filter_op: P,
 }
 
 /// Create a new `FilterMap` iterator.
 ///
 /// NB: a free fn because it is NOT part of the end-user API.
-pub fn new<M, P>(base: M, filter_op: P) -> FilterMap<M, P>
-    where M: ParallelIterator
+pub fn new<I, P>(base: I, filter_op: P) -> FilterMap<I, P>
+    where I: ParallelIterator
 {
     FilterMap {
         base: base,
@@ -24,9 +24,9 @@ pub fn new<M, P>(base: M, filter_op: P) -> FilterMap<M, P>
     }
 }
 
-impl<M, P, R> ParallelIterator for FilterMap<M, P>
-    where M: ParallelIterator,
-          P: Fn(M::Item) -> Option<R> + Sync,
+impl<I, P, R> ParallelIterator for FilterMap<I, P>
+    where I: ParallelIterator,
+          P: Fn(I::Item) -> Option<R> + Sync,
           R: Send
 {
     type Item = R;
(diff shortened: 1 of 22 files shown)


cuviper:
I'm not fond of changing `ITEM` to `I` -- that evokes "iterator" to my mind, especially when we're dealing with a library of iterators here.  I'd prefer plain `T` for items, as the idiomatic type placeholder in Rust.

src/iter/filter_map.rs
```diff
@@ -24,9 +24,9 @@ pub fn new<M, FILTER_OP>(base: M, filter_op: FILTER_OP) -> FilterMap<M, FILTER_O
     }
 }
 
-impl<M, FILTER_OP, R> ParallelIterator for FilterMap<M, FILTER_OP>
+impl<M, P, R> ParallelIterator for FilterMap<M, P>
```
nikomatsakis: I no longer recall why I used `M` for the "base" operations, but it doesn't seem terribly intuitive to me. I see that [the standard library uses `I`](https://doc.rust-lang.org/std/iter/struct.Filter.html), perhaps we should use that (after adopting @cuviper's suggestion to change `ITEM` to `T`)?

src/iter/filter_map.rs
```diff
@@ -59,25 +59,25 @@ impl<M, FILTER_OP, R> BoundedParallelIterator for FilterMap<M, FILTER_OP>
...
             base: base,
             filter_op: filter_op,
         }
     }
 }
 
-impl<'f, ITEM, MAPPED_ITEM, C, FILTER_OP> Consumer<ITEM> for FilterMapConsumer<'f, C, FILTER_OP>
-    where C: Consumer<MAPPED_ITEM>,
-          FILTER_OP: Fn(ITEM) -> Option<MAPPED_ITEM> + Sync + 'f
+impl<'p, I, MI, C, P> Consumer<I> for FilterMapConsumer<'p, C, P>
```
nikomatsakis: If we were using `T` for `Item`, I *think* I'd prefer `U` for `MAPPED_ITEM`. If we're going into single letter territory, might as well *go there*, I guess. Somewhat oddly (to my mind), the standard library [seems to adopt `B` for this purpose](https://doc.rust-lang.org/std/iter/struct.Map.html), though.

nikomatsakis:
OK, looking over the changes, I'm torn. I do think the existing names like like `MAP_OP` and `REDUCE_OP` are not great. But the single-letter names sure do feel cryptic, particularly when you have a whole bunch in sequence. Currently though we seem to have a mix, which is also not a win. I think all things considered I'm inclined to adopt this branch (after we finish bike-shedding the particular choices), since I'm on the fence and it aligns better with the prevailing style, but I'm mildly tempted to try and find better "one word" names (e.g., `FUNC` instead of `MAP_OP`) as an alternative.

leshow:
I think those are good suggestions, (`ITEM` -> `T`, `MAPPED_ITEM` -> `U`) I can move on those later today so we can see what it looks like.

There is a few other cases where I ended up with a 2 letter params.
`IDENTITY` -> `ID`
`OP` -> `OP`
`C_ITEM` -> `CI`
`PAR_ITER` -> `PI` (following the prev suggestion this could just be `I`)

For `C_ITEM`, it occupies the same position as `MAPPED_ITEM`, so I think that should also become `U` if we make the aforementioned change. I don't see the 2 letter params as a problem, but if you want to strictly pursue the 1 character generic name then those will have to find replacements.

About documentation on the param names, I'm not an expert Rust developer and this library is new to me so my opinion may be worthwhile here, having the name `MAPPED_ITEM` didn't help me, looking at the constraints in the `where` clause did though.

nikomatsakis:
Looks pretty good to me. I'm inclined to merge.

nikomatsakis:
@cuviper you want to give one more look, or are you satisfied?

cuviper:
I'm happy with this.

leshow:
Thank you!

PR #262, opened 2017-03-16
Use the standard Sum/Product traits

### LINKED ISSUE #252: Consider using the `std::iter::Sum`/`Product` traits
Instead of the current `SumOp`/`ProductOp`, we could use the same traits as `Iterator`, potentially enabling use with more than just the primitive types.

```rust
// Iterator:
fn sum<S>(self) -> S
    where S: Sum<Self::Item>

// ParallelIterator now:
fn sum(self) -> Self::Item
    where SumOp: ReduceOp<Self::Item>

// perhaps use Sum for folding, Add for reducing
fn sum(self) -> Self::Item
    where Self::Item: Sum + Add<Output=Self::Item>

// or even this, allowing sums by reference
// (but this is sometimes bad for type inference)
fn sum<S>(self) -> S
    where S: Sum<Self::Item> + Add<Output=S>
```

cuviper:
`SumOp` is replaced by `Sum + Add`, `ProductOp` by `Product + Mul`, and `ReduceOp` is now hidden.

Technically, we could get by without `Add`/`Mul`, but it's easier this way, and I find it unlikely that these would ever be missing from something that implements `Sum`/`Product`.

Fixes #252.

### DIFF
--- a/rayon-demo/src/pythagoras/mod.rs
+++ b/rayon-demo/src/pythagoras/mod.rs
@@ -22,7 +22,7 @@ fn par_euclid<FM, M, FN, N>(map_m: FM, map_n: FN) -> u32
           FN: Fn(Iter<u32>) -> N + Sync,
           N: ParallelIterator<Item=u32>,
 {
-    map_m((1u32 .. 2000).into_par_iter()).map(|m| {
+    map_m((1u32 .. 2000).into_par_iter()).map(|m| -> u32 {
         map_n((1 .. m).into_par_iter())
             .filter(|n| (m - n).is_odd() && m.gcd(n) == 1)
             .map(|n| 4000000 / (m*m + n*n))
@@ -32,7 +32,7 @@ fn par_euclid<FM, M, FN, N>(map_m: FM, map_n: FN) -> u32
 
 /// Same as par_euclid, without tweaking split lengths
 fn par_euclid_weightless() -> u32 {
-    (1u32 .. 2000).into_par_iter().map(|m| {
+    (1u32 .. 2000).into_par_iter().map(|m| -> u32 {
         (1 .. m).into_par_iter()
             .filter(|n| (m - n).is_odd() && m.gcd(n) == 1)
             .map(|n| 4000000 / (m*m + n*n))
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -65,6 +65,7 @@ pub use self::rev::Rev;
 mod len;
 pub use self::len::{MinLen, MaxLen};
 mod sum;
+mod product;
 
 #[cfg(test)]
 mod test;
@@ -430,10 +431,10 @@ pub trait ParallelIterator: Sized {
     /// Basically equivalent to `self.reduce(|| 1, |a, b| a * b)`,
     /// except that the type of `1` and the `*` operation may vary
     /// depending on the type of value being produced.
-    fn product(self) -> Self::Item
-        where Self::Item: Product
+    fn product<P>(self) -> P
+        where P: Send + Product<Self::Item> + Product
     {
-        reduce::reduce(self, reduce::PRODUCT)
+        product::product(self)
     }
 
     /// DEPRECATED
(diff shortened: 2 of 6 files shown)


cuviper:
I did try making this even more general like the `Iterator` trait does:
```rust
fn sum<S>(self) -> S
    where S: Sum<Self::Item> + Add<Output=S>
```

But this quickly proved to have the type inference problems I expected, as many of our existing tests couldn't decide what `S` to use without explicit annotation.

I also extended our use of `consume_iter()` here, as it's quite convenient with `sum()` and `product()`.  On that tangent, I think we might gain some performance by generally pushing more implementations of custom `fold_with()` and `consume_iter()`.

nikomatsakis:
> But this quickly proved to have the type inference problems I expected, as many of our existing tests couldn't decide what S to use without explicit annotation.

At first I was surprised by that signature, since it doesn't fit my mental model of the proper types for a "reduce" (which maps from T -> T) -- it seems more suitable for a `fold` -- but I guess that since you can always `Sum` an empty iterator, it works out fine. 

In any case, these inference problems are presumably the same for sequential iterators, right? It *might* make sense to adopt this more general form, to try and match sequential iterators to the extent possible.

nikomatsakis:
Seems good to me.

nikomatsakis:
I'm inclined to merge, though I think it may be worth adopting the "standard" inference signature, even if that requires more type annotations.

cuviper:
> At first I was surprised by that signature, since it doesn't fit my mental model of the proper types for a "reduce" (which maps from T -> T) -- it seems more suitable for a fold -- but I guess that since you can always Sum an empty iterator, it works out fine.

Yeah, it is like `fold` -- in fact, a naive implementation looks something like this:

```rust
fn sum<S>(self) -> S
    where S: Send + Sum<Self::Item> + Add<Output=S>
{
    self.fold(|| iter::empty().sum(),
              |sum, x| sum + iter::once(x).sum())
        .reduce(|| iter::empty().sum(), Add::add)
}
```

Used like this, `Sum` is little more than a glorified `Zero` trait.  But yet another thing I was considering was adding something like a `fold_iter` which would pass through the internal `consume_iter`, so you'd get a little nicer:

```rust
fn sum<S>(self) -> S
    where S: Send + Sum<Self::Item> + Add<Output=S>
{
    self.fold_iter(|iter| iter.sum())
        .reduce(|| iter::empty().sum(), Add::add)
}
```

But if we do this, we'd definitely want to make sure that `consume_iter` is used more often.

cuviper:
> Technically, we could get by without Add/Mul, but it's easier this way, and I find it unlikely that these would ever be missing from something that implements Sum/Product.

Rust 1.16 defies me -- `Result` lacks `Add`/`Mul`, but now has:
```rust
impl<T, U, E> Sum<Result<U, E>> for Result<T, E>
    where T: Sum<U>,
{/*...*/}

impl<T, U, E> Product<Result<U, E>> for Result<T, E>
    where T: Product<U>,
{/*...*/}
```

Supporting `Sum` alone would look something like `iter::once(x).chain(iter::once(y)).sum()` in place of `Add` for our reduction.  If we generalize `S`, we'd need both `S: Sum<Self::Item> + Sum` (aka `Sum<S>`).

nikomatsakis:
> I find it unlikely that these would ever be missing from something that implements Sum/Product....Rust 1.16 defies me

Heh, I remember thinking "oh, I bet **somebody** will"...

nikomatsakis:
@cuviper so I think we should merge this PR, but it needs a rebase now. I still kinda' think we should match the signature of `Iterator` as closely as possible.

cuviper:
Rebased, dropped add/mul, and generalized the output.

cuviper:
I still want to do an audit to push more implementations of `Producer::fold_with` and `Folder::consume_iter`, but we can do that separately.  That would only be an internal optimization, without impacting the API.

nikomatsakis:
Looks good to me, thanks. =)

bluss:
consume_iter/fold_with are great :+1:

PR #265, opened 2017-03-17
Forbid multiple calls to initialize

leodasvacas:
Remove `pub` from `global_registry`. `get_registry_with_config` renamed to `init_global_registry`. Forbid multiple calls to `intialize`, if the configurations were equal we allowed calls after the first as a no-op. However that's bogus because `Configuration` does not implement `PartialEq` and we would incorrectly assume that if the thread_num is equal then the configuration is equal.


rayon-core/src/configuration.rs
```diff
@@ -152,26 +152,12 @@ impl Configuration {
...
-
-    let registry = registry::get_registry_with_config(config);
-
-    if let Some(value) = num_threads {
-        if value != registry.num_threads() {
-            return Err(InitError::GlobalPoolAlreadyInitialized);
-        }
-    }
-
+    config.validate()?;
```
nikomatsakis: Unfortunately, we currently try and maintain compatibility with Rust 1.12.0, which I do not think had `?`; therefore it's better to use `try!`

nikomatsakis:
👍 to this change overall; I agree we shouldn't assume that `Configuration` can be `PartialEq` (now that it has closures, that doesn't really make sense)

leodasvacas:
Glad to know the change is welcome, replaced all question marks with try! including some in rayon-demo for consistency.

nikomatsakis:
@leodasvacas it seems the tests fail:

```
thread 'join::test::sort' panicked at 'expected InitOk, but got GlobalPoolAlreadyInitialized', rayon-core/src/join/test.rs:48
```

you can probably just remove that initialization call, seems useless there

leodasvacas:
@nikomatsakis Thanks for the heads up, I had also broken some benches, all green now.

nikomatsakis:
Looking good, thank you.

PR #266, opened 2017-03-18
Generalize the private trait marker and use it more

cuviper:
This moves the `private::PrivateMarker` to the top level, and adds `private_decl!{}` and `private_impl!{}` for us to apply this in a uniform way.

This marks several more traits as private using the new macros.  `string::Pattern` is only meant to be used by the already-private `ParallelString`.  I think `iter::MapOp`, and `iter::ReduceOp` should get completely hidden for #228, but we can at least close their implementations until then.  Similarly, I'd like the chunks traits to look more like `ParallelString`, but closing them off now is an incremental step.

I believe all remaining traits are desirable to leave open for outside implementations.

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -8,12 +8,6 @@ use iter::*;
 use iter::internal::*;
 use std::cmp::min;
 
-mod private {
-    /// If this type is pub but not publically rechable, third parties can't
-    /// name it and can't implement traits using it.
-    pub struct PrivateMarker;
-}
-
 
 /// Test if a byte is the start of a UTF-8 character.
 /// (extracted from `str::is_char_boundary`)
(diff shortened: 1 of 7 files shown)


src/string.rs
```diff
@@ -41,10 +35,7 @@ fn find_char_midpoint(chars: &str) -> usize {
 
 /// Parallel extensions for strings.
 pub trait ParallelString {
-    /// This trait is private; this method exists to make it impossible
-    /// to implement outside the crate.
-    #[doc(hidden)]
-    fn private_parallel_string(&self) -> private::PrivateMarker;
+    private_decl!{}
```
nikomatsakis: Can you add a comment to those traits which appear in the public API, indicating that they are not intended (nor permitted) to be implemented from outside of rayon?

nikomatsakis: Otherwise, LGTM.

cuviper: Done.

I was hoping to sneak that in with a `//!` in `private_decl!`, but that doesn't actually work.  Alas.

nikomatsakis:
This cargo CI failure is a real drag. These changes look good to me.

cuviper:
I'll rebase now that #260's stack tweak is merged -- let's see.

nikomatsakis:
Nice.

PR #271, opened 2017-03-21
add thread start and exit handlers to ThreadPool

froydnj:
Add handlers that are run when threads in the `ThreadPool` start and when they exit.


rayon-core/src/registry.rs
```diff
@@ -499,6 +503,19 @@ unsafe fn main_loop(worker: Worker<JobRef>, registry: Arc<Registry>, index: usiz
...
+    if let Some(ref handler) = registry.start_handler {
+        let registry = registry.clone();
+        match unwind::halt_unwinding(|| handler()) {
+            Ok(()) => {
+            }
+            Err(err) => {
+                registry.handle_panic(err);
+            }
+        }
+        registry.terminate()
```
nikomatsakis: OK, calling `registry.terminate()` in *this* location seems definitely wrong, since it will execute unconditionally. I could *imagine* that you might want to call it in the case of panic, but then it should move within the arm. I'm not entirely sure however if invoking `terminate` here will work; I'd have to review the logic.

rayon-core/src/configuration.rs
```diff
@@ -55,19 +55,30 @@ pub struct Configuration {
...
+
+    /// Closure invoked on worker thread exit.
+    exit_handler: Option<ExitHandler>,
 }
 
 /// The type for a panic handling closure. Note that this same closure
 /// may be invoked multiple times in parallel.
 pub type PanicHandler = Arc<Fn(Box<Any + Send>) + Send + Sync>;
 
+pub type StartHandler = Arc<Fn() + Send + Sync>;
```
nikomatsakis: I wonder if we should pass in the index of the thread, as we do for `get_thread_name`. Seems like probably yes.

nikomatsakis:
> The types for StartHandler and ExitHandler; perhaps these should be FnMut instead?

No, I think `Fn` is correct. These will be invoked from many different threads in parallel, after all!

> Handling panics from the thread start handler: I think the registry should just be terminated, but perhaps letting it run normally would be OK. 

Yes that's an interesting question. The `handle_panic` by default will abort the process anyhow, so there is no concern there. I'd be inclined to just carry on in the event that `handle_panic` returns normally. If the user *wants* to terminate themselves they can arrange it, though it'd be a bit involved. I'll sketch it out since it seems helpful for your idea of writing a test, as well :)

I imagine you might make a shared channel:

```rust
let (tx, rx) = channel.clone();

let init_closure = Arc::new(|| {
    tx.send(true); // everything went well!
});

let panic_handler = Arc::new(|err| {
    tx.send(false); // panic occurred
});

let pool = ThreadPool::new(NUM_THREADS);

let mut successful = 0;
for msg in rx {
    if msg {
        successful += 1;
        if successful == NUM_THREADS { break; }
    } else {
        // panic! tear down the threadpool.
        return;
    }
}

/* use pool here */
```

>  I had a go at implementing a small test for thread start handlers, but ran into problems being unable to mutate things from within the handler. See point 1.

You should be able to use either [atomics](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html) or a [mutex](https://doc.rust-lang.org/std/sync/struct.Mutex.html) to mutate data from the callback. Since the callback is running from many threads in parallel, though, we indeed will not permit arbitrary mutation -- but this is a feature, not a bug. =)

froydnj:
Thanks for the fast feedback!

> Yes that's an interesting question. The handle_panic by default will abort the process anyhow, so there is no concern there. I'd be inclined to just carry on in the event that handle_panic returns normally.

My thought here was that the user might have expected the exit handler to use data set up by the start handler, but registry termination is not quite the right approach to handle that.  Maybe it would be better to document that the exit handler only runs if the start handler didn't panic, and implement that in the code.

After your comments, and re-reading the docs for `std::sync::atomic`, I got the test to work properly.  I think I can get a reliable test for exit handlers to work based on your shared channel sketch, though it'd be convenient for testing and possibly for other applications to have a function `ThreadPool::shutdown` or somesuch to call, after which it is guaranteed that all threads in the thread pool have joined.  Maybe that would complicate the `ThreadPool` implementation unnecessarily, because you'd have to check whether the pool is shutdown or not on every API call.

I don't think you'd want to have the `Drop` implementation wait for all the threads to shutdown normally, but perhaps a `synchronous_drop` (naming suggestions welcome) flag in `Configuration` to enable people to ask for that if they so desire?

froydnj:
Updated with tests that work, a bit more polish, and rebased to current master.

nikomatsakis:
@froydnj 

> My thought here was that the user might have expected the exit handler to use data set up by the start handler, but registry termination is not quite the right approach to handle that. Maybe it would be better to document that the exit handler only runs if the start handler didn't panic, and implement that in the code.

Keep in mind that they must also specify the panic handler, or else they won't have to worry about recovery.

rayon-core/src/test.rs
```diff
@@ -30,3 +33,53 @@ fn worker_thread_index() {
...
+        .set_num_threads(16)
+        .set_start_handler(Arc::new(move |index| {
+            clone.fetch_add(1, Ordering::SeqCst);
+        }));
+    {
+        let pool = ThreadPool::new(conf).unwrap();
+    }
+
+    // We must have started at least one thread.
+    assert!(n_called.load(Ordering::SeqCst) > 0);
```
nikomatsakis: why must we have started at least one hread here? Perhaps it would be good to do something in the pool. For example:

```rust
pool.install(|| {
    .. // this is running in a worker thread, so we can certainly assert that the count is >0
});
```

nikomatsakis: but I think that personally I would be inclined to have them the startup cause messages to be sent, so that we can check that *all* threads start up. Though I'm not sure I'd want to guarantee that forever; but it *is* observable. Perhaps it's worth leaving a comment to the effect that threads may in the future start/stop at different times, but that whenever they start/stop, these handlers will be run.

froydnj:
> Keep in mind that they must also specify the panic handler, or else they won't have to worry about recovery.

@nikomatsakis I'm unsure of how to parse this.  Do you mean that if the start handler panics, for the exit handler to run at all must mean the user has installed a panic handler, that panic handler has run successfully, and we can therefore assume that they've ensured any state that the exit handler needs is in order?  That is, we don't have to do any special handling here?

froydnj:
The start handler tests have been updated to ensure that all the threads in the pool run the start handler.  Since the approach used for the start handler test was equally applicable to the exit handler test and IMHO is a bit nicer, I've updated the exit handler test to match.

nikomatsakis:
@froydnj 

> I'm unsure of how to parse this. Do you mean that if the start handler panics, for the exit handler to run at all must mean the user has installed a panic handler, that panic handler has run successfully, and we can therefore assume that they've ensured any state that the exit handler needs is in order? That is, we don't have to do any special handling here?

Yes, that is what I mean.

rayon-core/src/configuration.rs
```diff
@@ -140,6 +156,27 @@ impl Configuration {
         self
     }
 
+    /// Returns a copy of the current thread start callback.
+    pub fn start_handler(&self) -> Option<StartHandler> {
+        self.start_handler.clone()
+    }
+
+    /// Sets a callback to be invoked on thread start.
```
nikomatsakis: Let's expand this doc comment. For example, I would add a paragraph about panics:

"If this closure should panic, the panic will be passed to the panic handler. If that handler returns, then startup will continue as normal."

rayon-core/src/configuration.rs
```diff
@@ -140,6 +156,27 @@ impl Configuration {
...
+    pub fn set_start_handler(mut self, start_handler: StartHandler) -> Configuration {
+        self.start_handler = Some(start_handler);
+        self
+    }
+
+    /// Returns a copy of the current thread exit callback.
```
nikomatsakis: Similarly here.

rayon-core/src/test.rs
```diff
@@ -30,3 +32,58 @@ fn worker_thread_index() {
...
+        let _ = ThreadPool::new(conf).unwrap();
+        // Drop the pool so it stops the running threads.
+    }
+
+    // Wait for all the threads to have been scheduled to run.
+    barrier.wait();
+
+    // The handler must have been called on every exiting thread.
+    assert_eq!(n_called.load(Ordering::SeqCst), n_threads);
+}
```
nikomatsakis: Can you add a test for the panic behavior? You can specify a panic handler that increments an integer and returns, and just have a start/exit that panics, so we can validate that everything runs as it should

froydnj:
Thanks for the review!  Added some documentation and the requested test.

nikomatsakis:
@froydnj looks good to me; @cuviper, care to take a look?

nikomatsakis:
This is a change to `rayon-core` so probably good to get a few pairs of eyes on it.

nikomatsakis:
I was thinking that, as with all of the APIs that allow you to observe the number of threads and so forth, it'd be good to add a caveat to the documentation that -- at least if the number of threads requested is zero (i.e., the default) -- the current behavior regarding thread start/stop is unstable, and threads may in the future be changed to start/stop dynamically. But having an infallible handler that runs seems pretty harmless even at that time.

(The only change I could imagine here is that we might want to allow the handler to return `Err` so as to actually **request** termination, but I'm not sure I'm inclined to add that, it'd be more work for us to handle and I don't know what the use case is. Could be added backwards compatibly in the future anyhow.)

rayon-core/src/lib.rs
```diff
@@ -36,7 +36,9 @@ mod unwind;
 mod util;
 
 pub use configuration::Configuration;
+pub use configuration::ExitHandler;
```
cuviper: nit: you made the `rayon-core/src/lib.rs` and `test.rs` files executable -- please revert that.

rayon-core/src/registry.rs
```diff
@@ -520,6 +536,19 @@ unsafe fn main_loop(worker: Worker<JobRef>, registry: Arc<Registry>, index: usiz
 
     // Normal termination, do not abort.
     mem::forget(abort_guard);
+
+    // Inform a user callback that we exited a thread.
```
cuviper: We still reach this even if the regular user code panicked, right?

cuviper:
> the current behavior regarding thread start/stop is unstable, and threads may in the future be changed to start/stop dynamically. But having an infallible handler that runs seems pretty harmless even at that time.

Right, we should keep the expectation that your start/stop handlers will be called for dynamic threads too.

> The only change I could imagine here is that we might want to allow the handler to return Err so as to actually request termination

I would imagine that as a wholly separate API call, not tied to these handlers directly.  Then termination would be possible from any context, including from these handlers if you like.

-----

off-topic: I noticed that `panic_handler` claims to return ownership of its closure, resetting the configuration, but it really just clones the `Arc`.  I wonder what happened here?

rayon-core/src/lib.rs
froydnj: I...have no idea how that happened.  I'll fix that.

rayon-core/src/registry.rs
froydnj: e.g. if a job placed in the thread pool panicked?  I think the answer is "no".  We could certainly reach this code if some non-thread pool threads panicked, though.

froydnj:
> off-topic: I noticed that panic_handler claims to return ownership of its closure, resetting the configuration, but it really just clones the Arc. I wonder what happened here?

I had noticed that while trying to make some of my earlier changes.  I was going to submit a pull request to fix the documentation there.  Maybe it's a holdover from when `PanicHandler` wasn't an `Arc` and creating a register from the configuration really did steal the function value?

nikomatsakis:
> off-topic: I noticed that panic_handler claims to return ownership of its closure, resetting the configuration, but it really just clones the Arc. I wonder what happened here?

Oh, that's a doc fail. It's a holdover; I decided later on to use an `Arc`.

nikomatsakis:
> I was going to submit a pull request to fix the documentation there. 

Please do.

rayon-core/src/registry.rs
nikomatsakis: Actually, the answer here is (except in extreme conditions) yes. Generally, whenever user code panics, we catch the unwinding and then resume the unwinding over on one of the user's threads. Panics should never unwind up into this code we're looking at there.

The only reason that a user-induced panic could cause us not to reach this code is when it occurs due to `Drop` -- e.g., if dropping the user's closure triggers a panic, we won't catch and re-route that. In that case, we'll hit the `abort_guard` above and abort the entire process, basically. 

Similarly, in futures, there are some random cases where panics -- while caught -- cannot be clearly propagated anywhere, in which case we invoke the panic handler, which (typically) aborts. In the case where an abort occurred, obviously we won't execute this code.

nikomatsakis: (Do we want to add a test for this? Not sure it's needed, given that there are existing tests of panic propagation.)

froydnj: I think we can forego adding a test for this.

froydnj:
@nikomatsakis I don't think there are outstanding issues left...do I need to do anything else here, like squashing the series, etc.?

cuviper:
@froydnj Can you at least update the PR title and summary?  Those get written into the merge history, so we don't want to call that WIP anymore.  Then I'm inclined to go ahead and merge this.

froydnj:
@cuviper Done!  I am assuming that these commits will all be squashed when the pull request is merged; is that correct?

cuviper:
@froydnj I don't usually squash commits, but I will here if that's what you prefer.

froydnj:
@cuviper I was assuming that you'd want to preserve bisection and whatnot.  Even without that, I would prefer squashing the commits.  Thank you!

cuviper:
My *ideal* methodology is to have a series of smaller commits that are independently functional.  But this is definitely harder to achieve.  Anyway, here we go... and thanks again for the PR!

PR #272, opened 2017-03-23
Refactor slice extensions into a ParallelSlice trait

cuviper:
I started with a simple refactoring, and kept going...

- The former `ToParallelChunks`/`Mut` are now just part of `ParallelSlice`.
- A new `par_windows` is a parallel version of the standard `windows`.
- The slice iterator *types* are no longer part of the prelude.  We usually just want traits here.
- The slice iterator types are now renamed to mirror those in `std::slice`
  - `Iter`, `IterMut`, `Chunks`, `ChunksMut`, `Windows`
- ~~The distinct producer types weren't necessary - just implement `Producer` on the iterators.~~
- The private producers are now named `IterProducer`, `IterMutProducer`, etc.

### DIFF
--- a/src/iter/collections.rs
+++ b/src/iter/collections.rs
@@ -1,6 +1,6 @@
 use super::{IntoParallelIterator, ParallelIterator};
 use super::chain::Chain;
-use slice::{SliceIter, SliceIterMut};
+use slice::{Iter, IterMut};
 use vec::IntoIter as VecIter;
 use std::collections::{BinaryHeap, BTreeMap, BTreeSet, HashMap, HashSet, LinkedList, VecDeque};
 use std::iter::FromIterator;
@@ -85,7 +85,7 @@ vectorized!{ impl<T> IntoParallelIterator for VecDeque<T> }
 
 impl<'a, T: Sync> IntoParallelIterator for &'a VecDeque<T> {
     type Item = &'a T;
-    type Iter = Chain<SliceIter<'a, T>, SliceIter<'a, T>>;
+    type Iter = Chain<Iter<'a, T>, Iter<'a, T>>;
 
     fn into_par_iter(self) -> Self::Iter {
         let (a, b) = self.as_slices();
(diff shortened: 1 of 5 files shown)


cuviper:
> - The slice iterator types are now renamed to mirror those in std::slice
>     - Iter, IterMut, Chunks, ChunksMut, Windows

Now that I've tried this mirrored naming scheme, I really like it.  Having `std::foo::Iter`, then `rayon::foo::Iter` is the same thing in parallel.  I think we should do this elsewhere as much as possible.  So that would move `vec::VecIter` to `vec::IntoIter`, and all the string types would lose their `Par` and maybe rename that module to `rayon::str`.  `OptionIter` is a little weird in that we currently fold all three `Into`/`Iter`/`Mut` variations into this one.  Don't know what to about `RangeIter` either.

nikomatsakis:
> The distinct producer types weren't necessary - just implement Producer on the iterators.

This is the only that gives me pause. Since the producer traits are public, people can observe that the iterator implements it, and start doing "producer-like" things. That wasn't meant to be possible, there is supposed to be a "phase shift" when the producer *begins* (by calling `drive()`).

nikomatsakis:
> Having std::foo::Iter, then rayon::foo::Iter is the same thing in parallel. 

Big 👍 to this though. Great idea.

cuviper:
OK, I'll back out the producer change, and just give those simpler names than they had before.
(e.g. `Iter` and `IterProducer`, `Windows` and `WindowsProducer`, etc.)

We should probably make sure producers are consistently hidden throughout the code base.  If we trust rustdoc's list of implementors, the only other one exposed right now is `RangeIter`.  I'll touch that one when I look at using this new mirrored naming there.

cuviper:
Ditto `UnindexedProducer`, for which rustdoc lists `RangeIter` and `ParChars`.

nikomatsakis:
@cuviper maybe the name for `RangeIter` should just be `Range`? Unclear.

cuviper:
Right, the standard iterator is `std::ops::Range`.  Don't know that I like `rayon::ops::Range`, except that I like the idea of applying such symmetry everywhere.

nikomatsakis:
@cuviper it seems misleading. I think range is a special case, in that the iterator *is* the type, and we should just call it `RangeIter` or something.

cuviper:
I'm OK with treating ranges as a special case.  Perhaps `rayon::range::Iter` is slightly more consistent with where we're going?

Pushed the slice producer fixups.

cuviper:
Rebased to resolve merge conflicts.

Issue #205, opened 2017-01-12
make `Registry::new()` panic-safe 


nikomatsakis:
As discussed in #202, `Registry::new()` is not panic-safe. The call to `spawn()` for a worker thread might panic, which would leave the other worker threads hanging around. It'd be good to setup a guard or something to encourage those workers to shut themselves down.

One simple way to do this would be to have one-shot channel per thread, with `Registry::new()` sending a `()` message once everybody is up and going. That way, if any of the threads panic, the sender of the channel will be dropped and the other threads will receive an error.

cuviper:
The slight irony is that `Builder::spawn()` returns a `Result` instead of panicking, so #202 is actually improving that situation!  (But it doesn't check the result, hmm -- I'll note that.)

Do you see an actual safety issue here?  Or just leaking threads?

nikomatsakis:
@cuviper just leaking threads, I think

PR #273, opened 2017-03-23
Zero threads means use default

leodasvacas:
This simplifies the Configuration API and makes it more robust by removing the `NumberOfThreadsZero` error case and removing the `Option` from `num_threads`. The previous implementation was broken if the env var `RAYON_RS_NUM_CPUS == 0`. Now we check num_threads, then `RAYON_RS_NUM_CPUS` and finally use the default. The other commits are more agressive simplifications, killing off `validate` which is very worthwhile because it makes `ThreadPool::new` infallible and switching the `InitError` enum for a struct since it was left with a single variant. This is quite a breaking change.

### DIFF
--- a/rayon-core/src/configuration.rs
+++ b/rayon-core/src/configuration.rs
@@ -134,12 +134,6 @@ impl Configuration {
         self.stack_size = Some(stack_size);
         self
     }
-
-
-    /// Checks whether the configuration is valid.
-    pub fn validate(&self) -> Result<(), InitError> {
-        Ok(())
-    }
 }
 
 /// Initializes the global thread pool. This initialization is
@@ -160,7 +154,6 @@ impl Configuration {
 /// will return an error. An `Ok` result indicates that this
 /// is the first initialization of the thread pool.
 pub fn initialize(config: Configuration) -> Result<(), InitError> {
-    try!(config.validate());
     let registry = try!(registry::init_global_registry(config));
     registry.wait_until_primed();
     Ok(())
--- a/rayon-core/src/configuration.rs
+++ b/rayon-core/src/configuration.rs
@@ -6,33 +6,20 @@ use std::fmt;
 use std::sync::Arc;
 use registry;
 
-/// Custom error type for the rayon thread pool configuration.
+/// Error if the gloal thread pool is initialized multiple times.
 #[derive(Debug,PartialEq)]
-pub enum InitError {
-    /// Error if the gloal thread pool is initialized multiple times
-    /// and the configuration is not equal for all configurations.
-    GlobalPoolAlreadyInitialized,
-}
+pub struct GlobalPoolAlreadyInitialized;
 
-impl fmt::Display for InitError {
+impl fmt::Display for GlobalPoolAlreadyInitialized {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
-            InitError::GlobalPoolAlreadyInitialized => {
-                write!(f,
-                       "The gobal thread pool has already been initialized with a different \
+        write!(f, "The gobal thread pool has already been initialized with a different \
                         configuration. Only one valid configuration is allowed.")
-            }
-        }
     }
 }
 
-impl Error for InitError {
+impl Error for GlobalPoolAlreadyInitialized {
     fn description(&self) -> &str {
-        match *self {
-            InitError::GlobalPoolAlreadyInitialized => {
-                "global thread pool has already been initialized"
-            }
-        }
+        "global thread pool has already been initialized"
     }
 }
 
@@ -153,7 +140,7 @@ impl Configuration {
 /// changed. Therefore, if you call `initialize` a second time, it
 /// will return an error. An `Ok` result indicates that this
 /// is the first initialization of the thread pool.
-pub fn initialize(config: Configuration) -> Result<(), InitError> {
+pub fn initialize(config: Configuration) -> Result<(), GlobalPoolAlreadyInitialized> {
     let registry = try!(registry::init_global_registry(config));
     registry.wait_until_primed();
     Ok(())
--- a/rayon-core/src/configuration.rs
+++ b/rayon-core/src/configuration.rs
@@ -1,10 +1,13 @@
 #[allow(unused_imports)]
 use log::Event::*;
 use std::any::Any;
+use std::env;
 use std::error::Error;
+use std::str::FromStr;
 use std::fmt;
 use std::sync::Arc;
 use registry;
+use num_cpus;
 
 /// Error if the gloal thread pool is initialized multiple times.
 #[derive(Debug,PartialEq)]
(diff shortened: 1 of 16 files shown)


rayon-core/src/configuration.rs
```diff
@@ -6,48 +6,29 @@ use std::fmt;
 use std::sync::Arc;
 use registry;
 
-/// Custom error type for the rayon thread pool configuration.
+/// Error if the gloal thread pool is initialized multiple times.
 #[derive(Debug,PartialEq)]
-pub enum InitError {
-    /// Error if number of threads is set to zero.
-    NumberOfThreadsZero,
+pub struct GlobalPoolAlreadyInitialized;
```
cuviper: I'm happy with making this a struct, but I think it should still be an opaque `InitError`, with at least some private placeholder contents so it can't be created externally.  Or maybe even use `Box<Error>` where we need it, and let the actual error type be fully private.  We want rayon-core to be compatible ~forever.

rayon-core/src/configuration.rs
```diff
@@ -6,48 +6,29 @@ use std::fmt;
...
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
-            InitError::NumberOfThreadsZero => {
-                write!(f,
-                       "The number of threads was set to zero but must be greater than zero.")
-            }
-            InitError::GlobalPoolAlreadyInitialized => {
-                write!(f,
-                       "The gobal thread pool has already been initialized with a different \
+        write!(f, "The gobal thread pool has already been initialized with a different \
```
cuviper: Not *your* typo, but this should be "global". :)

rayon-core/src/configuration.rs
```diff
@@ -77,7 +58,7 @@ impl Configuration {
 
     /// Get the number of threads that will be used for the thread
     /// pool. See `set_num_threads` for more information.
-    pub fn num_threads(&self) -> Option<usize> {
+    pub fn num_threads(&self) -> usize {
         self.num_threads
```
cuviper: I think we should make this reflect the actual number that will be used.  So we should move the logic with `RAYON_RS_NUM_CPUS` and `num_cpus::get()` fallbacks here.

rayon-core/src/registry.rs
```diff
@@ -87,20 +87,21 @@ unsafe fn init_registry(config: Configuration) {
...
-            Some(value) => value,
-            None => {
-                match env::var("RAYON_RS_NUM_CPUS") {
-                    Ok(s) => usize::from_str(&s).expect("invalid value for RAYON_RS_NUM_CPUS"),
-                    Err(_) => num_cpus::get(),
-                }
+        // Determine number of threads to spawn. Use configuration value if not zero.
+        // If configuration is zero try the RAYON_RS_NUM_CPUS env var.
+        // If that fails use the default (num_cpus).
+        let config_num_threads = configuration.num_threads();
```
cuviper: With the logic moved, we can directly use this value.  Maybe just `assert` that we ended up non-zero.

rayon-core/src/thread_pool/mod.rs
```diff
@@ -20,9 +20,8 @@ impl ThreadPool {
     /// Constructs a new thread pool with the given configuration. If
     /// the configuration is not valid, returns a suitable `Err`
     /// result.  See `InitError` for more details.
-    pub fn new(configuration: Configuration) -> Result<ThreadPool, InitError> {
-        try!(configuration.validate());
-        Ok(ThreadPool { registry: Registry::new(configuration) })
+    pub fn new(configuration: Configuration) -> ThreadPool {
+        ThreadPool { registry: Registry::new(configuration) }
```
cuviper: I hesitate to make this infallible.  There's an open question in #205, what to do when thread creation fails.  We could just decide to make that panic "nicely", cleaning up as much as possible and then continuing the panic, so we'd just unwind though here.  But I think it's also reasonable to return a nice error.  Perhaps we should leave this as an opaque `Box<Error>` for future possibilities.

rayon-core/src/configuration.rs
leodasvacas: @cuviper What about setting `num_threads` eagerly by moving the logic to `new()` and then have `set_num_threads` override it (if argument not zero)?

leodasvacas: Actually, `set_num_threads` should redo the logic if called with `0`.

cuviper: That's possible, but I'm not sure there's much benefit to doing it eagerly.  This isn't something I expect to get called repeatedly, and lazy computation seems simpler.

leodasvacas:
@cuviper Thank you for the great review, all points have been addressed. I went with `Box<Error>` for maximum future-proofness and `GlobalPoolAlreadyInitialized` should no longer be accessible outside `rayon-core`.

rayon-core/src/configuration.rs
```diff
@@ -1,53 +1,36 @@
...
 use registry;
+use num_cpus;
 
-/// Custom error type for the rayon thread pool configuration.
+/// Error if the gloal thread pool is initialized multiple times.
 #[derive(Debug,PartialEq)]
-pub enum InitError {
-    /// Error if number of threads is set to zero.
-    NumberOfThreadsZero,
+pub struct GlobalPoolAlreadyInitialized;
```
nikomatsakis: maybe we should move this into the `registry.rs` module and make it private there (and just have the register already convert to `Box<Error>`). Having it public here makes me nervous, and it doesn't seem to be locally used anywhere anyhow.

zayenz:
Not sure if this is of interest in the context of the current pull request, but anyway. If a new semantics for the number of threads is to be used, the Gecode (http://www.gecode.org) system has a nice way of specifying the number of threads to use using a double value. Quoting from the documentation (http://www.gecode.org/doc-latest/MPG.pdf, section 9.3.1): 

Assume that your computer has m processing units and that the value for threads is n.
- If n = 0, then m threads are used (as many as available processing units).
- If n ≥ 1, then n threads are used (absolute number of threads to be used).
- If n ≤ −1, then m + n threads are used (absolute number of processing units not to be used). For example, when n = −6 and m = 8, then 2 threads are used.
- If 0 < n < 1, then n · m threads are used (relative number of processing units to be used). For example, when n = 0.5 and m = 8, then 4 threads are used.
- If −1 < n < 0, then (1 + n) · m threads are used (relative number of processing units not to be used). For example, when n = −0.25 and m = 8, then 6 threads are used.
Note that all values are of course rounded and that at least one thread will be used.

leodasvacas:
Review by @nikomatsakis addressed, thanks!

@zayenz Always good to know what other libraries are doing! Given how easy it is to get the number of CPUs in Rust I'd leave more complex logic up to the user. Taking a float might negate the value of this PR since `set_num_threads` would be fallible again (have to check for `NaN`).

PR #278, opened 2017-03-24
Move `FromParallelIterator` directly into `mod iter`

cuviper:
Before, when this trait was under the private `mod from_par_iter`,
rustdoc copied its documentation to the reexport paths in both `iter`
and `prelude`.  It was the only thing in `prelude` that didn't *look*
like a reexport, and its relative links were broken in this path.

Now that `iter::FromParallelIterator` is the actual definition path, it
shows up as a proper reexport under the `prelude`.


bluss:
You can control rustdoc a bit by using `#[doc(no_inline)]`, by the way.

PR #279, opened 2017-03-27
Rustfmt Reapply

ChristopherDavenport:
It seems like we have a few features of rustfmt that havn't stayed fully current. 

This was current on nightly if that is what you are looking for.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -248,12 +248,11 @@ pub trait ParallelIterator: Sized {
     fn reduce_with<OP>(self, op: OP) -> Option<Self::Item>
         where OP: Fn(Self::Item, Self::Item) -> Self::Item + Sync
     {
-        self.map(Some)
-            .reduce(|| None, |opt_a, opt_b| match (opt_a, opt_b) {
-                (Some(a), Some(b)) => Some(op(a, b)),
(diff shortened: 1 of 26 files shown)


src/iter/mod.rs
```diff
@@ -561,12 +561,14 @@ pub trait ParallelIterator: Sized {
...
     fn find_last<P>(self, predicate: P) -> Option<Self::Item>
-        where P: Fn(&Self::Item) -> bool + Sync {
+        where P: Fn(&Self::Item) -> bool + Sync
+    {
         find_first_last::find_last(self, predicate)
     }
 
     #[doc(hidden)]
-    #[deprecated(note = "parallel `find` does not search in order -- use `find_any`, `find_first`, or `find_last`")]
+    #[deprecated(note = "parallel `find` does not search in order -- use `find_any`, \\
```
ChristopherDavenport: I am uncertain if this is the appropriate way to handle splitting a Single string into a MultiLine String Split to lower 
line length which this exceeded.

nikomatsakis: rustfmt is annoying like that.

nikomatsakis:
Looks reasonably good to me, thanks @ChristopherDavenport.

cuviper:
Should we apply this to rayon-core and rayon-demo too?

PR #280, opened 2017-03-27
Rustfmt rayon-core and rayon-demo

ChristopherDavenport:
In response to missing core in the last PR, this applies rustfmt to rayon-core and rayon-demo

### DIFF
--- a/rayon-core/src/future/mod.rs
+++ b/rayon-core/src/future/mod.rs
@@ -5,7 +5,7 @@
 //!
 //! See `README.md` for details.
 
-use latch::{LatchProbe};
+use latch::LatchProbe;
 #[allow(warnings)]
 use log::Event::*;
 use futures::{Async, Poll};
@@ -62,7 +62,9 @@ pub unsafe trait FutureScope<'scope> {
 ///
 /// NB. Although this is public, it is not exposed to outside users.
 pub fn new_rayon_future<'scope, F, S>(future: F, scope: S) -> RayonFuture<F::Item, F::Error>
-    where F: Future + Send + 'scope, S: FutureScope<'scope>,
+where
+    F: Future + Send + 'scope,
+    S: FutureScope<'scope>,
 {
     let inner = ScopeFuture::spawn(future, scope);
 
@@ -76,8 +78,8 @@ pub fn new_rayon_future<'scope, F, S>(future: F, scope: S) -> RayonFuture<F::Ite
         return RayonFuture { inner: hide_lifetime(inner) };
     }
 
-    unsafe fn hide_lifetime<'l, T, E>(x: Arc<ScopeFutureTrait<T, E> + 'l>)
-                                      -> Arc<ScopeFutureTrait<T, E>> {
+    unsafe fn hide_lifetime<'l, T, E>(x: Arc<ScopeFutureTrait<T, E> + 'l>,)
+        -> Arc<ScopeFutureTrait<T, E>> {
         mem::transmute(x)
     }
 }
@@ -94,10 +96,13 @@ impl<T, E> RayonFuture<T, E> {
             unsafe {
                 (*worker_thread).wait_until(&*self.inner);
                 debug_assert!(self.inner.probe());
-                self.poll().map(|a_v| match a_v {
-                    Async::Ready(v) => v,
-                    Async::NotReady => panic!("probe() returned true but poll not ready")
-                })
+                self.poll()
+                    .map(
+                        |a_v| match a_v {
+                            Async::Ready(v) => v,
+                            Async::NotReady => panic!("probe() returned true but poll not ready"),
+                        },
+                    )
             }
         }
     }
@@ -134,7 +139,9 @@ impl<T, E> Drop for RayonFuture<T, E> {
 /// ////////////////////////////////////////////////////////////////////////
 
 struct ScopeFuture<'scope, F, S>
-    where F: Future + Send + 'scope, S: FutureScope<'scope>,
+where
+    F: Future + Send + 'scope,
+    S: FutureScope<'scope>,
 {
     state: AtomicUsize,
     registry: Arc<Registry>,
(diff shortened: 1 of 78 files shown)


rayon-core/src/configuration.rs
```diff
@@ -63,7 +63,8 @@ impl Configuration {
     /// Set a closure which takes a thread index and returns
     /// the thread's name.
     pub fn set_thread_name<F>(mut self, closure: F) -> Self
-    where F: FnMut(usize) -> String + 'static {
+        where F: FnMut(usize) -> String + 'static
```
stjepang: Isn't the new style to align `where` with `pub fn`?

rayon-core/src/future/mod.rs
```diff
@@ -184,18 +192,18 @@ impl<'scope, F, S> ScopeFuture<'scope, F, S>
...
+                                             contents: Mutex::new(ScopeFutureContents {
+                                                                      spawn: None,
+                                                                      unpark: None,
+                                                                      this: None,
+                                                                      scope: Some(scope),
+                                                                      waiting_task: None,
+                                                                      result: Ok(Async::NotReady),
+                                                                      canceled: false,
+                                                                  }),
+                                         });
```
stjepang: This looks pretty bad, IMO :(

rayon-core/src/configuration.rs
ChristopherDavenport: I was running a brand new update of rustfmt so I'm surprised if it didn't come through. Have the new changes made it to release?

rayon-core/src/future/mod.rs
nikomatsakis: My policy is generally to run rustfmt unless it produces errors or whatever; but feel free to file a bug in rustfmt if you don't like the output, I guess.

rayon-core/src/configuration.rs
nikomatsakis: Yeah, perhaps not...

rayon-core/src/future/mod.rs
nikomatsakis: (Actually, a better option would to search around for the appropriate conversation in the https://github.com/rust-lang-nursery/fmt-rfcs/ repository...)

nikomatsakis:
Can you squash these into one commit (or otherwise remove the merge commit, in any case)?

ChristopherDavenport:
Removed Merge Commit and Rebased Off Current Master to resolve merge conflict. 

After rebasing in I reapplied rustfmt on the upstream code. Which has some changes which are in the proper commit.

Sorry for the sloppy history, however it is not present in the final product.

rayon-core/src/configuration.rs
bluss: It's implemented (rustfmt 8.1) but not the default. where_style = "Rfc" should be it.

Conspicuous config file https://github.com/rust-lang-nursery/rustfmt/blob/master/rfc-rustfmt.toml

nikomatsakis:
@ChristopherDavenport see @nrc's [latest tweet](https://twitter.com/nick_r_cameron/status/846519755599101952): 

> Rustfmt 8.1 released - https://crates.io/crates/rustfmt  Lots of block-indent improvements. Try the emerging RFC-approved style using rfc-rustfmt.toml

Maybe you can add this file into the repo and use those settings?

ChristopherDavenport:
So the problem with the new RFC style is that is appears to break macro invocation code by applying trailing commas in places  which breaks compilation.

I removed them in the commit but reapplying cargo fmt seems to reapply these broken fixes. So I'm unsure if we would like to apply such radical fixes which are labeled as very WIP.

nrc:
That is https://github.com/rust-lang-nursery/rustfmt/issues/1419 I should fix it within the next week and issue a 0.8.2 release

nikomatsakis:
@ChristopherDavenport d'oh

nikomatsakis:
maybe we'll wait for the fix then

nrc:
I released Rustfmt 0.8.3, it should fix the problem here.

ChristopherDavenport:
On it.

ChristopherDavenport:
Sorry about all the mess on this PR. I also worked through to get a successful CI test with rustfmt if you would like that to be incorporated as well, meaning moving forward all code would stay current with the standards outlined in rustfmt.toml. I can make that as a separate PR later though.

It would however fix the problem of having to come back and deal with code after the initial commit.

cuviper:
In case you missed it, @nikomatsakis was contemplating a bot for ongoing maintenance:
https://internals.rust-lang.org/t/idea-rustfmtbot/5039

ChristopherDavenport:
Check, I won't worry about it. You should know that current rustfmt fails to build against 1.12.0. so whatever the solution is will need to insure it doesn't run against that. I'll see if I can try my hands at a bot that does that just for the fun of it. Even if in practice it is done by something else.

ChristopherDavenport:
So I have applied rusftmt again to whole project as a single commit. If there is still interest I hope this will be merged soon because rebasing this after every day or so is getting a little tiring.

nikomatsakis:
@ChristopherDavenport yeah sorry about that. Thanks! :)

nikomatsakis:
@ChristopherDavenport hmm some of the formatting within closures etc looks pretty messed up. I'm inclined to file some rustfmt bugs and maybe revisit this when they've made more progress with the block formatting. I would imagine that [closures are not meant to look like this](https://github.com/nikomatsakis/rayon/pull/280/files#diff-6a14094077d75dad989b9e8f9a440937R103) and [these kinds of if expressions don't seem to line up very well](https://github.com/nikomatsakis/rayon/pull/280/files#diff-2c4fd7e6fd8962b2edf0f53bc2b6dff0R344).

nikomatsakis:
Filed two rustfmt bugs. I guess I'll close this for now.

nikomatsakis:
Sorry for the confusion @ChristopherDavenport

stjepang:
I expressed negative feelings towards rustfmt a while back, but wanna say that it has improved *a lot*. 
These days rustfmt is absolutely fantastic and I use it all the time. Thanks and great job @nrc!

PR #281, opened 2017-03-27
Add unique iterator types for all options/results/collections

cuviper:
We now have all six `{option,result}::{IntoIter,Iter,IterMut}` as distinct types matching the iterators in the standard library.  Only `option::IntoIter` is actually implemented on its own, and the rest are newtype wrappers on that using a new `delegate_iterator!` macro.

For all of the `std::collections`, create distinct parallel iterators with the same path in rayon, e.g. `rayon::collections::hash_map::Iter`. Most of these are still implemented by vectorizing, using a sequential iterator to collect a vector, and then iterating *that* in parallel. But now this detail is not actually exposed in the API, so we can feel free to improve on this in the future without a breaking change.


cuviper:
One question is whether the vectorized iterators should still expose the fact that they can be indexed.  If we ever got "native" parallel iterators for them, this may not hold.  For example, my experiment with a [forked `HashMap`](https://github.com/cuviper/rayon/tree/hash/rayon-hash/src/par) only implements unindexed producers.  So we might not want to chain ourselves to having an indexed producer here.

src/delegate.rs
```diff
@@ -0,0 +1,97 @@
...
+///     impl<'a, T: Ord + Sync, U: Sync>
+/// }
+/// ```
+macro_rules! delegate_iterator_item {
+    ($( #[ $attr:meta ] )+
+     $iter:ident < $( $i:tt ),* > => $inner:ty : $item:ty,
+     impl $( $args:tt )*
+     ) => {
+        $( #[ $attr ] )+
+        pub struct $iter $( $args )* {
```
nikomatsakis: nice

src/delegate.rs
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+
+        impl $( $args )* ExactParallelIterator for $iter < $( $i ),* > {
+            fn len(&mut self) -> usize {
+                self.inner.len()
+            }
+        }
+
+        impl $( $args )* IndexedParallelIterator for $iter < $( $i ),* > {
```
nikomatsakis: ah; I see, we'd have to modify the macro to avoid implementing all of these add'l traits

cuviper: Right.  I'll probably split the macro into a base unindexed case and then an indexed one on top of it.  We still should allow indexed where it makes sense for a collection, like option, result, vecdeque, and maybe binary heap (debatable).  We don't want indexed where the type probably wouldn't support it, in the maps/sets and in linked lists.

cuviper:
OK, now the indexed iterators are explicit, and the rest have only the base `ParallelIterator`.

PR #282, opened 2017-03-27
Removed Weights From Rayon

### LINKED ISSUE #270: remove weight APIs entirely
In #111, we decided to deprecate the weight APIs and @cuviper replaced with "min/max len" APIs. We kept around the weight APIs as deprecated. Before 1.0, but after our next release (0.7), we should remove them.

This is a good starter task, but I'm not marking it as *help wanted* just yet, since we don't want to do it until after the next release.

ChristopherDavenport:
This a removal of the deprecated weight functions. I know this needs to wait to be merged until after release `0.7`

Fixes #270


nikomatsakis:
Thanks @ChristopherDavenport :)

nikomatsakis:
We can land this now =)

PR #283, opened 2017-03-30
Abuse `links` to make sure there's only one rayon-core

cuviper:
One of the main points of having `rayon-core` is so we can stabilize it
~forever, and make sure there's only one instance running, as everyone
should agree on some 1.x.y.  But that only works for caret requirements;
one could still specify distinct versions with tilde, wildcard, or
inequality requirements.

There is a `links` manifest key for making sure only one crate links to a
particular native library.  We can abuse that to achieve a similar effect
for rayon-core without even linking anything.

    error: native library `rayon-core` is being linked to by more than
    one package, and can only be linked to by one package

As long as every published rayon-core specifies `links` this way, they
won't be allowed to coexist.


cuviper:
cc @alexcrichton - Do you see any problem with doing this?  Is there a better way?

alexcrichton:
This'll certainly have the desired effect I believe, but I don't understand enough about the technical details to know whether this is desirable or not.

rayon-core/build.rs
```diff
@@ -0,0 +1,5 @@
+// We need a build script to use `link = "rayon-core"`.  But we're not
+// *actually* linking to anything, just making sure that we're the only
+// rayon-core in use.
+fn main() {
+}
```
bluss: The empty build script should still declare dependencies. The default is that the build script (and thus the whole crate) is recompiled if any file changes. Including for example a README in the same directory.

cuviper: Fair point, I'll update that.

bluss: Rayon is looking really good with all the changes after 0.6!

nikomatsakis:
@cuviper maybe add something into the comment on the main rust-core module explaining what it means if you get an error? I imagine the error you get from having multiple rust-core versions is fairly cryptic. Other than, r=me.

cuviper:
The exact message is that `error: ...` quoted in the PR here.  I don't think it's *that* cryptic, but still I agree it's worth spelling out the restriction in the crate docs.

PR #284, opened 2017-03-31
Hide map iterators behind standard names

cuviper:
This creates type aliases exported as `rayon::iter::{Cloned, Inspect, Map}`, matching `std::iter`.  They still refer to the implementation details of the shared base type, but that's in the private `mod map`.

It's not quite as clean as having truly distinct types, which would require separate parallel iterator implementations, but the effect on the API surface is similar.  They can only be named by these aliases, so we should be free to update that implementation later if we like.

### DIFF
--- a/src/iter/map.rs
+++ b/src/iter/map.rs
@@ -10,6 +10,9 @@ pub trait MapOp<In>: Sync {
     private_decl!{}
 }
 
+
+pub type Map<I, F> = BaseMap<I, MapFn<F>>;
+
 pub struct MapFn<F>(pub F);
 
 impl<F, In, Out> MapOp<In> for MapFn<F>
@@ -23,6 +26,9 @@ impl<F, In, Out> MapOp<In> for MapFn<F>
     private_impl!{}
 }
 
+
+pub type Cloned<I> = BaseMap<I, MapCloned>;
+
 pub struct MapCloned;
 
 impl<'a, T> MapOp<&'a T> for MapCloned
@@ -34,6 +40,9 @@ impl<'a, T> MapOp<&'a T> for MapCloned
     }
     private_impl!{}
 }
+
+
+pub type Inspect<I, F> = BaseMap<I, MapInspect<F>>;
 
 pub struct MapInspect<F>(pub F);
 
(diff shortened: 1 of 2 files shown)


cuviper:
I also have a branch which fully commits to separate types for these, if you'd like to see that.

nikomatsakis:
Does this play nicely with rustdoc? (Does this matter?)

cuviper:
It's _okay_, but not great.  The overview looks fine:

screen shot 2017-04-05 at 09 45 06

But the details:

screen shot 2017-04-05 at 09 46 37

You do get to see what the underlying type is, but it's not linked since they don't have documentation (coming from a private module).  And there's no information about what traits it implements.

cuviper:
Here's the branch comparison when I make them new types instead. https://github.com/nikomatsakis/rayon/compare/master...cuviper:new-map

(We could still keep the common producer/consumer/folder, but it seems less valuable now.)

nikomatsakis:
@cuviper it is a lot of code... I guess this is something we could change backwards compatibly, probably, given that user cannot name that type.

nikomatsakis:
We could probably just jot down a note in the type alias that you should only refer to the type by its alias?

cuviper:
> @cuviper it is a lot of code... I guess this is something we could change backwards compatibly, probably, given that user cannot name that type.

I'm not too worried about the amount of code there, but I hesitated that you might not like it, which is why I did the type aliases first. :smile:

But I don't like with the aliases that you can't see that these implement `ParallelIterator`, and `IndexedParallelIterator` if the base does, etc.  We could manually document that, but that feels like another hack upon a hack...

Yes, I think we can change this compatibly -- if not, I'd push harder toward the new types.

> We could probably just jot down a note in the type alias that you should only refer to the type by its alias?

Is it meaningful to recommend one _should_ do the only thing one _can_ do?

nikomatsakis:
@cuviper it's probably better to just use the newtypes. It feels right.

nikomatsakis:
@cuviper it'd be less code if we simplified the `Iterator` hierarchy, I suppose

cuviper:
Closed in favor of #304.

PR #288, opened 2017-04-02
Terminate new registries after panic or spawn errors

### LINKED ISSUE #205: make `Registry::new()` panic-safe

cuviper:
This adds a guard structure around the `Registry::new` thread startup,
which terminates the whole registry if we can't complete it.  It now
returns a boxed error if `Builder::spawn` fails, and should deal better
with any panics, like from the user's `thread_name` callback.

Fixes #205.

### DIFF
--- a/rayon-core/src/thread_pool/test.rs
+++ b/rayon-core/src/thread_pool/test.rs
@@ -62,15 +62,39 @@ fn sleeper_stop() {
     registry.wait_until_stopped();
 }
 
+/// Create a start/exit handler that increments an atomic counter.
 fn count_handler() -> (Arc<AtomicUsize>, ::StartHandler) {
     let count = Arc::new(AtomicUsize::new(0));
     (count.clone(), Arc::new(move |_| { count.fetch_add(1, Ordering::SeqCst); }))
 }
 
-#[test]
-fn failed_thread_stack() {
+/// Immediately unwrap a counter, asserting that it's no longer shared.
+fn unwrap_counter(counter: Arc<AtomicUsize>) -> usize {
+    Arc::try_unwrap(counter)
+        .expect("Counter is still shared!")
+        .into_inner()
+}
+
+/// Wait until a counter is no longer shared, then return its value.
+fn wait_for_counter(mut counter: Arc<AtomicUsize>) -> usize {
     use std::{thread, time};
 
+    for _ in 0..60 {
+        counter = match Arc::try_unwrap(counter) {
+            Ok(counter) => return counter.into_inner(),
+            Err(counter) => {
+                thread::sleep(time::Duration::from_secs(1));
+                counter
+            },
+        };
+    }
+
+    // That's too long -- do it now!
+    unwrap_counter(counter)
+}
+
+#[test]
+fn failed_thread_stack() {
     let (start_count, start_handler) = count_handler();
     let (exit_count, exit_handler) = count_handler();
     let config = Configuration::new()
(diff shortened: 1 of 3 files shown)


rayon-core/src/thread_pool/test.rs
```diff
@@ -57,3 +58,22 @@ fn sleeper_stop() {
...
+
+#[test]
+fn panic_thread_name() {
+    let config = Configuration::new().set_num_threads(10).set_thread_name(|i| {
+        if i == 5 { panic!(); }
+        format!("panic_thread_name#{}", i)
+    });
+    let pool = unwind::halt_unwinding(|| ThreadPool::new(config));
+    assert!(pool.is_err(), "thread-name panic should propagate!");
+    // TODO how to make sure threads actually stopped?
```
cuviper: Once #271 lands, I could use callbacks to count the actual thread starts and exits.

rayon-core/src/thread_pool/test.rs
```diff
@@ -57,3 +61,59 @@ fn sleeper_stop() {
...
+    let config = Configuration::new()
+        .set_stack_size(::std::usize::MAX)
+        .set_start_handler(start_handler)
+        .set_exit_handler(exit_handler);
+
+    let pool = ThreadPool::new(config);
+    assert!(pool.is_err(), "thread stack should have failed!");
+
+    // Give time for the internal cleanup to terminate.
+    thread::sleep(time::Duration::from_secs(1));
```
nikomatsakis: Is any delay needed in this particular test? I shouldn't think so.

rayon-core/src/thread_pool/test.rs
```diff
@@ -57,3 +61,59 @@ fn sleeper_stop() {
...
+                                 panic!();
+                             }
+                             format!("panic_thread_name#{}", i)
+                         });
+
+    let pool = unwind::halt_unwinding(|| ThreadPool::new(config));
+    assert!(pool.is_err(), "thread-name panic should propagate!");
+
+    // Give time for the internal cleanup to terminate.
+    thread::sleep(time::Duration::from_secs(1));
```
nikomatsakis: I think it'd be better to have the exit handlers send messages to a shared queue (i.e., a channel), so that we can just read from the channel and verify that we get the expected number of responses. This won't be sensitive to timing.

cuviper: Well, `mpsc::Sender` isn't `Sync`, which means the exit handler wouldn't be either.

I think maybe an easier method for both tests is to just use the `Arc::strong_count()` of the handler itself.  When that drops to one, we know that nobody else is around anymore.  This still requires polling, but it's not timing sensitive.  (It's only a test, so it doesn't need to be ideal, just reliable.)

cuviper: (even `mpsc::SyncSender` is `!Sync` -- ha!)

cuviper:
OK, tweaked the tests.

nikomatsakis:
👍

PR #290, opened 2017-04-05
fix the documentation for Configuration::panic_handler

froydnj:
The previous comment dates from a time when `Arc` was not used.


nikomatsakis:
Thanks =)

PR #292, opened 2017-04-06
Tighten up the public API in `Configuration`

### LINKED ISSUE #291: Consider hiding Arc from the Configuration API
In the `Configuration` setters for panic handler et al., it's awkward that one has to write `Arc::new(||...)` for the closure instead of writing it bare.  We probably should at least drop this part, and `Arc` it internally.  The setters would be generic over the actual closure type.

The getters for these handlers return `Arc` clones.  We could also hide this, and return a reference instead.  It's a little weird to imagine what users would do with these getters at all.  However, `Registry::new` is using that to effectively move them -- get a clone of the `Arc`, then drop the `Configuration`.  And that explains what the comment removed in #290 was thinking, as returning a reference won't work there.

We could make the getters steal the value again, but this can't be generic, so if not `Arc` then it could only be `Box`.  Or if we want these choices to simply be hidden from users, then we could change these getters to free functions in `mod configuration`.

nikomatsakis:
- This PR removes the `Arc<>` closure types and instead just takes a closure directly.
- I also tightened up the `Configuration` API by removing all the "getters".
- Renamed `ThreadPool::num_threads` to `current_num_threads` for consistency.
- Made the lack of guarantees when number of threads is not specified more clear.

Fixes #291


nikomatsakis:
@cuviper note the last commit -- just hacking around that odd test failure, not sure what's the background there.

rayon-core/src/lib.rs
```diff
@@ -58,6 +59,243 @@ pub use future::RayonFuture;
...
+    }
+
+    /// Get the thread name for the thread with the given index.
+    fn thread_name(&mut self, index: usize) -> Option<String> {
+        self.get_thread_name.as_mut().map(|c| c(index))
+    }
+
+    /// Set a closure which takes a thread index and returns
+    /// the thread's name.
+    pub fn set_thread_name<F>(mut self, closure: F) -> Self
```
cuviper: Now that the public methods are all setters, the `set_` prefix doesn't feel idiomatic.  Consider the standard library builders like [`Command`](https://doc.rust-lang.org/std/process/struct.Command.html) and [`thread::Builder`](https://doc.rust-lang.org/std/thread/struct.Builder.html) -- their methods are just named for the thing they're setting.

(But of course, we'd need to rename the private getters.)

rayon-core/src/lib.rs
```diff
@@ -58,6 +59,243 @@ pub use future::RayonFuture;
...
+    /// that number. To reproduce the current default behavior, you
+    /// may wish to use the [`num_cpus`
+    /// crate](https://crates.io/crates/num_cpus) to query the number
+    /// of CPUs dynamically.
+    pub fn set_num_threads(mut self, num_threads: usize) -> Configuration {
+        self.num_threads = num_threads;
+        self
+    }
+
+    /// Returns a copy of the current panic handler.
```
cuviper: nit: it's not a copy.  (ditto on the other take methods)

cuviper:
> note the last commit -- just hacking around that odd test failure, not sure what's the background there.

FWIW, I don't care about testing the stack in particular here.  I just wanted *something* that would make `Builder::spawn` return an error.  I just tried on Windows 10, which also fails to fail:

        thread 'thread_pool::test::failed_thread_stack' panicked at 'thread stack should have failed!', rayon-core\src\thread_pool\test.rs:106

Windows passes (fails properly) with `set_stack_size(::std::isize::MAX as usize)`.  Can you try that on your Mac?  Linux x86_64 still works too.  But then this fails on Linux i686, as it does actually allocate a 2GB stack for one thread and then fails the second, and we're expecting not to see any threads.

Anyway, I can tweak the test in a followup PR.  It's not a big deal to have it disabled for now.

cuviper:
Tight!

PR #294, opened 2017-04-06
add Josh Stone as an author

nikomatsakis:
@cuviper has been co-maintainer for some time, but I forgot to update `Cargo.toml`. Thanks @cuviper for being great.


cuviper:
Thanks!

PR #295, opened 2017-04-07
Make `Builder::spawn` fail more consistently

cuviper:
We want to test a `spawn` error, but macOS and Windows were somehow
running with the `usize::MAX` stack.  (Probably overflow due to rounding
the page size up or something, and then who knows.)  With `isize::MAX`
we get the error we want, with a little care for 32-bit platforms.


nikomatsakis:
👍

PR #298, opened 2017-04-07
Derive default for `Configuration`

leodasvacas:
And use it in `Configuration::new` to save some boilerplate.


cuviper:
This means we must  to commit to having `Default` forever, but that should be fine given that we already have an arg-free `new()`, and we must have some notion of default configuration for the global pool.

So, looks good, thanks!

PR #299, opened 2017-04-07
Add osx testing on Travis CI

cuviper:


cuviper:
@nikomatsakis We could also add AppVeyor for Windows, but you'll have to enable the project hooks.

nikomatsakis:
Hmm. Yes, the fast turn-around for Rayon CI is **so** nice. But let's leave it enabled I guess for a bit and see how it feels.

PR #300, opened 2017-04-08
Remove bounded and exact iterators

### LINKED ISSUE #229: simplify `ParallelIterator` hierarchy
Currently we have 4 traits in the parallel iterator hierarchy, we should simplify it before 1.0. The question is whether to have 2 or 3 traits. The current design is:

- `ParallellIterator`
- `BoundedParallelIterator`
- `ExactParallelIterator`
- `IndexedParallelIterator`

It seems clear that exact + indexed can be merged, since there isn't much use to knowing precisely how many items will be produced if you can't split it. Is there still a use for bounded?

After some discussion, @cuviper and I were thinking that we should remove both bounded and exact parallel iterator, since the primary use case is more efficient `collect()`, and it is likely that this will be achieved with specialization and without the needs for these traits. And we can always add them later if needed -- possibly as a breaking change, but possibly not.

### LINKED ISSUE #259: How is the BoundedParallelIterator supposed to be used?
I am currently [experimenting](https://github.com/slide-rs/specs/pull/120) implementing `ParallelIterator` for `specs`.
I have implemented it for the `BitSet` that it uses and and I am now looking at can I implement `BoundedParallelIterator` for it.
I can easily give upper bound how many bits are set in the set, but the trait seems to be designed for randomly indexable collections? Is that oversight or am I missing something?

cuviper:
This reduces our iterator traits to just `ParallelIterator` and
`IndexedParallelIterator`.  The bounded and exact iterators had some
theoretical use, but in practice the only distinction we use today is
whether something can be indexed.

The `upper_bound` method is totally removed.  The rest of the methods in
the removed traits are now merged into `IndexedParallelIterator`.

Fixes #229.
Closes #259.

### DIFF
--- a/rayon-demo/src/vec_collect.rs
+++ b/rayon-demo/src/vec_collect.rs
@@ -100,7 +100,7 @@ mod vec_i {
 
     const N: u32 = 4 * 1024 * 1024;
 
-    fn generate() -> impl ExactParallelIterator<Item=u32> {
+    fn generate() -> impl IndexedParallelIterator<Item=u32> {
         (0_u32..N)
             .into_par_iter()
     }
--- a/src/delegate.rs
+++ b/src/delegate.rs
@@ -98,25 +98,17 @@ macro_rules! delegate_indexed_iterator_item {
             impl $( $args )*
         }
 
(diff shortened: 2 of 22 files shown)


nikomatsakis:
Let's do it.

PR #304, opened 2017-04-12
Split into separate map/clone/inspect iterators

cuviper:
Before, the public API exposed the fact that these shared a common
`MapOp` implementation.  Now they are distinct iterator types with
fully custom implementations.


nikomatsakis:
Seems good.

PR #306, opened 2017-04-12
Simplify reduce and add min_by and max_by

cuviper:


src/iter/mod.rs
```diff
@@ -233,11 +233,15 @@ pub trait ParallelIterator: Sized {
     fn reduce_with<OP>(self, op: OP) -> Option<Self::Item>
         where OP: Fn(Self::Item, Self::Item) -> Self::Item + Sync
     {
-        self.map(Some).reduce(|| None, |opt_a, opt_b| match (opt_a, opt_b) {
-            (Some(a), Some(b)) => Some(op(a, b)),
-            (Some(v), None) | (None, Some(v)) => Some(v),
-            (None, None) => None,
-        })
+        self.fold(|| None, |opt_a, b| match opt_a {
```
nikomatsakis: Why change this?

cuviper: It *seems* like a waste to wrap values in `Some` only to immediately match that away.  It requires inlining and optimization to realize that the right side always has something while folding. In practice, I couldn't measure a difference, so I'll drop this change if you prefer.

nikomatsakis: I see. That's true. I don't have a strong opinion, just curious. I agree that it seems like this formulation *might* unlock some optimization somehow.

PR #308, opened 2017-04-12
Remove Deprecated Interfaces

### LINKED ISSUE #305: Remove all deprecated interfaces
Since we already removed `weight` which was only just deprecated in 0.7, let's go ahead and remove all of the older deprecated interfaces we're still carrying around.

ChristopherDavenport:
Removed reduce_with_identity, mul, find, and position.

Fixes #305

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -244,15 +244,6 @@ pub trait ParallelIterator: Sized {
             })
     }
 
-    /// Deprecated. Use `reduce()` instead.
-    #[deprecated(since = "v0.5.0", note = "call `reduce` instead")]
-    fn reduce_with_identity<OP>(self, identity: Self::Item, op: OP) -> Self::Item
-        where OP: Fn(Self::Item, Self::Item) -> Self::Item + Sync,
-              Self::Item: Clone + Sync
-    {
-        self.reduce(|| identity.clone(), op)
-    }
-
     /// Parallel fold is similar to sequential fold except that the
     /// sequence of items may be subdivided before it is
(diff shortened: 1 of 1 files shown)


src/iter/mod.rs
```diff
@@ -581,15 +563,6 @@ pub trait ParallelIterator: Sized {
         find_first_last::find_last(self, predicate)
     }
 
-    #[doc(hidden)]
-    #[deprecated(note = "parallel `find` does not search in order -- use `find_any`, \\
-    `find_first`, or `find_last`")]
-    fn find<P>(self, predicate: P) -> Option<Self::Item>
```
nikomatsakis: Ah, wait, shoot! `find` was meant to be kept indefinitely.

nikomatsakis: Would you mind reverting this part in another PR?

The idea was to help people who port over a sequential iterator and guide them to the right thing.

src/iter/mod.rs
```diff
@@ -850,15 +823,6 @@ pub trait IndexedParallelIterator: ParallelIterator {
             .map(|(i, _)| i)
     }
 
-    #[doc(hidden)]
-    #[deprecated(note = "parallel `position` does not search in order -- use `position_any`, \\
-    `position_first`, or `position_last`")]
-    fn position<P>(self, predicate: P) -> Option<usize>
```
nikomatsakis: Same with `position`

src/iter/mod.rs
ChristopherDavenport: I'm on it.

PR #309, opened 2017-04-13
Revert Subset of Remove Deprecated Interfaces

ChristopherDavenport:
Find and Position are meant to be kept indefinitely to guide them to the right thing.


ChristopherDavenport:
Rustup nightly seems broken. It failed on update for me locally as well so I don't believe that this one is on me.

nikomatsakis:
Thanks, I'm going to assume this PR is not at fault.

PR #310, opened 2017-04-13
Add `install_future` to `ThreadPool`

torkleyy:
Same as `install`, but non-blocking and returns a future

### DIFF
--- a/rayon-core/src/thread_pool/mod.rs
+++ b/rayon-core/src/thread_pool/mod.rs
@@ -1,6 +1,6 @@
 use Configuration;
 #[cfg(feature = "unstable")]
-use future::{Future, RayonFuture};
+use future::{Async, Future, Poll, RayonFuture};
 use latch::LockLatch;
 #[allow(unused_imports)]
 use log::Event::*;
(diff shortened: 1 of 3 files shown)


rayon-core/src/thread_pool/mod.rs
```diff
@@ -123,3 +152,47 @@ impl Drop for ThreadPool {
...
+{
+    type Item = R;
+    type Error = ();
+
+    fn poll(&mut self) -> Poll<R, ()> {
+        use latch::LatchProbe;
+
+        {
+            let job = self.job.as_ref().expect("Contract error: poll already returned a value");
+            if !job.latch.probe() {
```
torkleyy: Comment in `LockProbe`'s implementation says inefficient?

rayon-core/src/thread_pool/mod.rs
```diff
@@ -50,6 +50,35 @@ impl ThreadPool {
...
+    /// thread-local data from the current thread will not be
+    /// accessible.
+    ///
+    /// # Panics
+    ///
+    /// Polling the [`InstallFuture`] will propagate panics from
+    /// `op`.
+    ///
+    /// [`InstallFuture`]: struct.InstallFuture.html
+    #[cfg(feature = "unstable")]
```
torkleyy: Is it good to place it here, or would `feature != unstable` add these doc comments to `current_num_threads`?

torkleyy:
oh right `rustup` cannot install latest nightly atm...

torkleyy:
Not really sure if this is a necessary change, just close it if it's not

rayon-core/src/thread_pool/mod.rs
```diff
@@ -50,6 +50,35 @@ impl ThreadPool {
...
+    pub fn install_future<OP, R>(&self, op: OP) -> InstallFuture<OP, R>
+        where OP: FnOnce() -> R + Send
+    {
+        unsafe {
+            let job_a = StackJob::new(op, LockLatch::new());
+            self.registry.inject(&[job_a.as_job_ref()]);
+
+            InstallFuture { job: Some(job_a) }
```
cuviper: I don't think this is safe at all.  A `StackJob` must stay at the **exact** same memory location until completion, because you're handing its pointer to the registry.  When you move it to the `InstallFuture` and return, the registry is left with a dangling pointer!

torkleyy:
@cuviper Like I said, I actually thought of something that didn't quite make sense, and asked myself the same.

cuviper:
OK, thanks anyway!  If you do find the current interfaces lacking, let's figure that out together before we remove the "unstable" feature guard.

PR #312, opened 2017-04-14
[WIP] Sequential-to-parallel adapter

nikomatsakis:
This is more of a sketch than anything else, at least thus far, but it's kind of cute. It allows you to take more-or-less **any** sequential iterator and use it with Rayon -- though it works best with an iterator that is `Send`. It works by forking off, for each item, a parallel future to process it, and then pulling them back in. I really like this basic idea but I'm not sure about a number of details of how the implementation *actually* works. Anyway I thought I'd throw it up just so that people can take a look-see.

Some thoughts:

- The current mechanism of pulling back in tasks is sort of not ideal. It will invoke `rayon_wait()` on things at the bottom of the deque, which will cause us to start processing (locally) from the top of the deque; we'll stop when someone has stolen (and completed) enough things from the bottom of the deque. It just feels like we're not quite tracking what we probably **want** to be tracking.
- It feels like a lot of this logic exists in "streams" and maybe some more integration of streams into Rayon (not quite sure yet what that means) would pay off.
- It's not quite using the consumer protocol correctly, I don't think, in terms of which values it reduces with what. It'd be good to make a test case showing this. =)
- Tuning and benchmarking have not been done.
- I suspect that rewriting it into a more deeply futures-based version, rather than using a for loop combined with `rayon_wait()`, would be beneficial.
- I would also like to experiment with a version that uses [flat-combining](https://github.com/nikomatsakis/rayon/pull/141) to have the spawned tasks update a shared structure.

### DIFF
--- a/src/pipeline/mod.rs
+++ b/src/pipeline/mod.rs
@@ -1,0 +1,213 @@
+#![cfg(feature = "unstable")]
+
+use futures::lazy;
+use prelude::*;
+use iter::internal::*;
+use std::collections::VecDeque;
+use Scope;
+
+mod test;
+
+pub trait IntoParPipeline
+    where Self: Sized + IntoIterator, Self::Item: Send,
+{
+    /// Converts a sequential iterator into a parallel one. The
+    /// sequential iterator will be driven from a single thread, and a
+    /// Rayon task will be created for each item in the iterator.
+    ///
+    /// In general, it is preferable to use a native parallel iterator
+    /// if they are available. It may also be better to collect the
+    /// items into a vector and then use `into_par_iter()` on that
+    /// vector -- this can help us to create fewer parallel tasks and
+    /// reduce overall overhead.
+    ///
+    /// If the sequential iterator is not `Send`, see `into_par_pipeline_scoped()`.
+    fn into_par_pipeline(self) -> Pipeline<Self> where Self: Send;
+
+    /// Like `into_par_pipeline()`, but also works for iterators that
+    /// are not `Send`. Requires that your code is within a Rayon
+    /// scope, which ensures that the code is already running on the
+    /// Rayon threadpool.
+    ///
+    /// Example:
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    /// use std::rc::Rc; // `Rc` is not `Send`
+    /// rayon::scope(|s| {
+    ///     let v = vec![Rc::new(1), Rc::new(2)];
+    ///     v.into_iter()
+    ///      .map(|r| *r) // iterating over i32s
+    ///      .into_par_pipeline_scoped(s)
+    ///      .find_any(|&n| n == 1);
+    /// });
+    /// ```
+    fn into_par_pipeline_scoped<'s, 'scope>(self, scope: &'s Scope<'scope>) -> PipelineScoped<'s, 'scope, Self>;
+}
+
+impl<I> IntoParPipeline for I
+    where I: IntoIterator, I::Item: Send,
+{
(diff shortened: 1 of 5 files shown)


src/pipeline/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// Rayon task will be created for each item in the iterator.
+    ///
+    /// In general, it is preferable to use a native parallel iterator
+    /// if they are available. It may also be better to collect the
+    /// items into a vector and then use `into_par_iter()` on that
+    /// vector -- this can help us to create fewer parallel tasks and
+    /// reduce overall overhead.
+    ///
+    /// If the sequential iterator is not `Send`, see `into_par_pipeline_scoped()`.
+    fn into_par_pipeline(self) -> Pipeline<Self> where Self: Send;
```
nikomatsakis: Also, this needs a better name. But what?

src/pipeline/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// use std::rc::Rc; // `Rc` is not `Send`
+    /// rayon::scope(|s| {
+    ///     let v = vec![Rc::new(1), Rc::new(2)];
+    ///     v.into_iter()
+    ///      .map(|r| *r) // iterating over i32s
+    ///      .into_par_pipeline_scoped(s)
+    ///      .find_any(|&n| n == 1);
+    /// });
+    /// ```
+    fn into_par_pipeline_scoped<'s, 'scope>(self, scope: &'s Scope<'scope>) -> PipelineScoped<'s, 'scope, Self>;
```
nikomatsakis: Supporting non-Send iterators is kind of a pain and limits the implementation -- should we just drop them altogether?

src/pipeline/mod.rs
cuviper: Maybe `into_par_stream`?  I think "stream" is a little clearer than "pipeline" in general.

src/pipeline/mod.rs
cuviper: What's a realistic iterator that's non-Send itself, but still yields `Send` items?

We're already a bit limited that we don't have `ParallelIterator: Send`, but I think perhaps we should.  For instance, `Chain`'s drive functions would be better if they could `join` the left and right sides, but right now we can't.

cuviper: (On `ParallelIterator`, all of the methods which take closures currently require `Sync` but not `Send`, so that broadens the question considerably there.)

nikomatsakis: > What's a realistic iterator that's non-Send itself, but still yields Send items?

https://github.com/BurntSushi/walkdir/

nikomatsakis: > We're already a bit limited that we don't have ParallelIterator: Send, but I think perhaps we should

Yes, perhaps so.

> (On ParallelIterator, all of the methods which take closures currently require Sync but not Send, so that broadens the question considerably there.)

Probably they should be send + sync I guess.

cuviper:
> https://github.com/BurntSushi/walkdir/

Ah, and AFAICT that only lacks `Send` because its optional [`sort_by`](https://docs.rs/walkdir/1.0.7/walkdir/struct.WalkDir.html#method.sort_by) closure is not constrained.  That's boxed rather than monomorphized, so it can't even be conditionally `Send`.

nikomatsakis:
@cuviper yes, not a really good reason I suspect. No idea how representative this is.

nikomatsakis:
I'll close this for now and come back to it sometime.

PR #313, opened 2017-04-14
RFC: stop using the `_async` terminology

nikomatsakis:
Between the `_async` and `_future` suffixes, there was quite a zoo of
possibilities:

- `Scope::spawn`
- `Scope::spawn_future`
- `::spawn_async`
- `::spawn_async_future`

After this PR, there are now two orthogonal concepts. First, what
**scope** are you spawning the task into? If you use the free-standing
 functions, like `rayon::spawn()`, or methods on a thread-pool, like
 `pool.spawn_future()`, that is the "static" or "global" scope, and
 hence no stack references are allowed. Otherwise, if you create a scope
 using `scope()`, then when you do `scope.spawn()` it is spawned into
 that scope.

Second, are you spawning a task for side-effects (`spawn()`) or to
produce a result (`spawn_future()`).

**NB:** Everything but `Scope::spawn()` remains "unstable" in this PR.

### DIFF
--- a/rayon-core/src/thread_pool/mod.rs
+++ b/rayon-core/src/thread_pool/mod.rs
@@ -26,6 +26,26 @@ impl ThreadPool {
         Ok(ThreadPool { registry: registry })
     }
 
+    /// Returns a handle to the global thread pool. This is the pool
+    /// that Rayon will use by default when you perform a `join()` or
+    /// `scope()` operatioon, if no other thread-pool is installed. If
+    /// no global thread-pool has yet been started when this function
+    /// is called, then the global thread-pool will be created (with
+    /// the default configuration). If you wish to configure the
+    /// global thread-pool differently, then you can use [the
+    /// `rayon::initialize()` function][f] to do so.
+    ///
+    /// [f]: fn.initialize.html
+    #[cfg(feature = "unstable")]
+    pub fn global() -> &'static Arc<ThreadPool> {
+        lazy_static! {
+            static ref DEFAULT_THREAD_POOL: Arc<ThreadPool> =
+                Arc::new(ThreadPool { registry: Registry::global() });
+        }
+
+        &DEFAULT_THREAD_POOL
+    }
+
     /// Executes `op` within the threadpool. Any attempts to use
     /// `join`, `scope`, or parallel iterators will then operate
     /// within that threadpool.
@@ -123,3 +143,4 @@ impl Drop for ThreadPool {
         self.registry.terminate();
     }
 }
+
--- a/rayon-core/src/thread_pool/mod.rs
+++ b/rayon-core/src/thread_pool/mod.rs
@@ -6,7 +6,7 @@ use latch::LockLatch;
 use log::Event::*;
 use job::StackJob;
 #[cfg(feature = "unstable")]
-use spawn_async;
+use spawn;
 use std::sync::Arc;
 use std::error::Error;
 use registry::{Registry, WorkerThread};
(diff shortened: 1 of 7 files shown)


nikomatsakis:
I also considered adding a synthetic "global scope", so that you would do something like `rayon::GLOBAL_SCOPE.spawn()`, but it just looks ugly.

nikomatsakis:
I suppose `rayon::global_scope().spawn(...)` is not so ugly. And, to be clear, this idea of a global scope is actually something that exists **in the implementation**: the `futures` code for example uses the `Registry` as a kind of "global scope" (i.e., it implements [the `FutureScope` trait](https://github.com/nikomatsakis/rayon/blob/master/rayon-core/src/future/mod.rs#L54-L58)). Hmm, I sort of like that.

One could also removing the free functions and giving a way to get a handle on the "global thread-pool" (probably an `Arc<ThreadPool>`). Then you would do `rayon::global_pool().spawn()` and hence there would be two spawn methods, one on thread-pools and one on scopes.

rayon-core/src/thread_pool/mod.rs
```diff
@@ -97,24 +97,24 @@ impl ThreadPool {
         }
     }
 
-    /// Spawns an asynchronous task in this thread-pool. See
-    /// `spawn_async()` for more details.
+    /// Spawns a task in this thread-pool contained within the static
+    /// scope. See `spawn()` for more details.
```
cuviper: The suggested details should be an actual link.

rayon-core/src/thread_pool/mod.rs
```diff
@@ -97,24 +97,24 @@ impl ThreadPool {
...
     {
         // We assert that `self.registry` has not terminated.
-        unsafe { spawn_async::spawn_async_in(op, &self.registry) }
+        unsafe { spawn::spawn_in(op, &self.registry) }
     }
 
-    /// Spawns an asynchronous task in this thread-pool. See
-    /// `spawn_future_async()` for more details.
+    /// Spawns an future in this thread-pool within the static
+    /// scope. See `spawn_future()` for more details.
```
cuviper: Ditto, please link this.

cuviper:
> One could also removing the free functions and giving a way to get a handle on the "global thread-pool" (probably an Arc<ThreadPool>). Then you would do rayon::global_pool().spawn() and hence there would be two spawn methods, one on thread-pools and one on scopes.

I like the idea of a `global_pool()`, but note that this is *not* exactly the same as the free functions.  Those operate on the *current* pool (`Registry::current()`) which is still valuable in itself.  I think that's a good generalized model -- use free functions for the current pool, `ThreadPool` methods for a targeted pool.

nikomatsakis:
@cuviper 

> I think that's a good generalized model -- use free functions for the current pool, ThreadPool methods for a targeted pool.

Yes. I realized this at some point. =)

nikomatsakis:
> The only thing not clear to me -- already underdocumented -- is the lifetime of Scope::spawn_future.

Yeah, we need docs here.

> I guess the returned future must either be completed or dropped before the scope returns, but it would be nice to see this spelled out.

Probably this is accurate, I'm not 100% sure what you mean. The answer is that the future must have finished executing before the scope is dropped; but the returned future *can* escape the scope, if its types don't include any references.

nikomatsakis:
@cuviper so I'm inclined to merge this PR, and leave the docs of future for another PR. In particular, I want to open a PR that moves the future support out of core (which is blocked on this one), and we can start thinking about docs then. Seem ok?

rayon-core/src/thread_pool/mod.rs
```diff
@@ -97,24 +119,46 @@ impl ThreadPool {
...
+              RA: Send,
+              RB: Send
+    {
+        self.install(|| join(oper_a, oper_b))
+    }
+
+    /// Execute `oper_a` and `oper_b` in the thread-pool and return
+    /// the results. Equivalent to `self.install(|| scope(...))`.
+    #[cfg(feature = "unstable")]
+    pub fn scope<'scope, OP, R>(&self, op: OP) -> R
```
cuviper: These docs look like copypasta from `join` above.

rayon-core/src/thread_pool/mod.rs
```diff
@@ -97,24 +119,46 @@ impl ThreadPool {
...
+        where OP: for<'s> FnOnce(&'s Scope<'scope>) -> R + 'scope + Send, R: Send
+    {
+        self.install(|| scope(op))
+    }
+
+    /// Spawns an asynchronous task in this thread-pool. See `spawn()`
+    /// for more details.
+    #[cfg(feature = "unstable")]
+    pub fn spawn<OP>(&self, op: OP)
```
cuviper: These docs should be more clear what `spawn()` it's referencing - presumably the global one and not a recursive reference.  Maybe even put a link on that.

nikomatsakis:
@cuviper ok, fixed the docs I think (I tested and the links seem to resolve too...)

cuviper:
The current nightly is broken for CI, but locally my slightly older nightly tested fine.

👍

PR #315, opened 2017-04-15
Add while_some; impl FromParallelIterator for Option/Result

cuviper:
The new `while_some` filters for `Some` values, and halts if it encounters any
`None`.  This is inspired by `Itertools::while_some`, although being parallel
means it doesn't consider the order.

Using this, we can implement a short-circuiting collect for `Option` and
`Result` like the standard library, returning just `None` or `Err` respectively
if any such item is found in the iterator.  Since this operates in parallel,
it's not deterministic which items will have already been visited, nor which
`Err` will be returned if there are multiple.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -67,6 +67,8 @@ mod cloned;
 pub use self::cloned::Cloned;
 mod inspect;
 pub use self::inspect::Inspect;
+mod while_some;
+pub use self::while_some::WhileSome;
 
 #[cfg(test)]
 mod test;
@@ -594,6 +596,15 @@ pub trait ParallelIterator: Sized {
         where P: Fn(Self::Item) -> bool + Sync
     {
         self.map(predicate).find_any(|&p| !p).is_none()
+    }
+
+    /// Creates an iterator over the `Some` items of this iterator, halting
+    /// as soon as any `None` is found.
+    fn while_some<T>(self) -> WhileSome<Self>
+        where Self: ParallelIterator<Item = Option<T>>,
+              T: Send
+    {
+        while_some::new(self)
     }
 
     /// Create a fresh collection containing all the element produced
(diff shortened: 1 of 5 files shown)


cuviper:
> This is inspired by Itertools::while_some, although being parallel
> means it doesn't consider the order.

For reference: [`Itertools::while_sum()`](https://docs.rs/itertools/0.6.0/itertools/trait.Itertools.html#method.while_some)

The difference is similar to `Itertools::find` vs. `ParallelIterator::find_any`, which we deliberately chose to name differently to avoid any ordering assumptions by the user.  I'm not sure of a good rename in this case -- `while_all_some`? `until_any_none`?  meh...

nikomatsakis:
Interesting. I'm not too worried about the naming here.

cuviper:
I also thought of a similar idea for a plain `while`, something like an unordered `take_while`, but it's not as useful for the collects here since the predicate is separate from the value.  I'd have to do `while(|x| x.is_some())` and then unwrap it in a separate step.  But maybe `while` could be useful in general.

nikomatsakis:
Isn't `while` a keyword?

cuviper:
Ah, indeed.  Phooey.

nikomatsakis:
The name `while_some` definitely only makes limited sense in the parallel context. I'm struggling to come up with a better one, though. It's basically "keep processing until a some is seen *anywhere*", so "while" is not inappropriate I guess.

nikomatsakis:
I say we merge it =) I was debating if we wanted to call it "unstable" till we find a name we like better, but I think the analogy to itertools alone makes this name a good choice, and anyway we have the option of breaking changes to rayon for a while yet.

cuviper:
> It's basically "keep processing until a some is seen anywhere"

The opposite - while some, until none.

cuviper:
> anyway we have the option of breaking changes to rayon for a while yet.

Indeed!  We've already made a few since 0.7 too. :)

nikomatsakis:
> The opposite - while some, until none.

Right yes

PR #316, opened 2017-04-15
Remove FromParallelIterator from the prelude

cuviper:
The standard prelude does not include `FromIterator`, expecting that
folks will usually interact with this only through `collect()`.  We can
say the same about `FromParallelIterator`.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -909,7 +909,8 @@ pub trait IndexedParallelIterator: ParallelIterator {
 /// By implementing `FromParallelIterator` for a type, you define how it will be
 /// created from an iterator.
 ///
-/// `FromParallelIterator` is used through [`ParallelIterator`]'s [`collect()`] method.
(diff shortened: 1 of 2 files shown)


bluss:
It would be great if FromIterator was in the prelude though, an interface like `Vec::from_iter` is usually easier to work with and more intuitive. That said, I know you value consistency with std high in rayon.

cuviper:
I can go either way - I also value real user feedback highly. 😀

nikomatsakis:
I also can go either way. In principle, I too prefer `Vec::from_iter`, but in practice I always use `collect()`. I think partly because I wind up with a setup like:

```rust
let x: Vec<_> =
    foo
    .bar
    .collect();
```

and that feels relatively nice (I almost *always* write the type annotation).

cuviper:
OK, let's just leave it in.  It's not that harmful, since the method name is pretty unique.

PR #317, opened 2017-04-15
Fold should pass the full check to its base

cuviper:
`FoldConsumer` and `FoldFolder` never got their `fn full`, probably
because these features were developed near the same time.  They just
need to pass the call down to their base consumer/folder.

### DIFF
--- a/src/iter/fold.rs
+++ b/src/iter/fold.rs
@@ -73,6 +73,10 @@ impl<'r, U, T, C, ID, F> Consumer<T> for FoldConsumer<'r, C, ID, F>
             fold_op: self.fold_op,
         }
     }
+
+    fn full(&self) -> bool {
+        self.base.full()
+    }
 }
 
 impl<'r, U, T, C, ID, F> UnindexedConsumer<T> for FoldConsumer<'r, C, ID, F>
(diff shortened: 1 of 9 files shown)


src/iter/fold.rs
```diff
@@ -114,4 +118,8 @@ impl<'r, C, ID, F, T> Folder<T> for FoldFolder<'r, C, ID, F>
     fn complete(self) -> C::Result {
         self.base.consume(self.item).complete()
     }
+
+    fn full(&self) -> bool {
+        self.base.full()
+    }
```
nikomatsakis: Should we make `full` required instead of optional, to prevent us from forgetting it?

cuviper: I guess that depends on whether we view this as merely an optimization to quit early, or if it carries some kind of guarantee of timeliness.  Probably a soft guarantee if any, being parallel and all, since we don't want a ton of synchronization.

Additional consumers/folders that don't have `full` since they're fine with the default: `collect`, `for_each`, `reduce`, `sum`, `product`, and the internal no-op.

nikomatsakis: I view it as somewhere in between. For example, if you have an infinite iterator and you are using `find`, then if `full` doesn't work properly, your program might never terminate.

cuviper: Fair point, and those that don't get full are logically infinite given infinite input.  I guess we should make that required rather than implicit.

cuviper:
OK, `full` is now required.

nikomatsakis:
👍

PR #320, opened 2017-04-17
Clarify the ordering of find_first etc.

cuviper:
It's not fully clear that "first" in `find_first` and `position_first`
is referring to the first in the sequential order of items, rather than
the temporally first match found anywhere like `find_any`.  This update
emphasizes that we mean sequentially first.

There shouldn't be the same issue in `find_last` and `position_last`, as
there's no reason to look for a "temporally last" match, but those docs
are also updated just to keep similarity to the "first" methods.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -537,30 +537,33 @@ pub trait ParallelIterator: Sized {
         find::find(self, predicate)
     }
 
-    /// Searches for the **first** item in the parallel iterator that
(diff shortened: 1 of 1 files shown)


cuviper:
This is trying to address the misinterpretation reported [here](https://www.reddit.com/r/rust/comments/65ukhv/i_reimplemented_a_friends_python_tool_in_rust/dgdufjn/?context=10000).

PR #321, opened 2017-04-19
Add Windows CI(Appveyor) and enable unstable features in Travis on stable Rust

lilianmoraru:
I enabled the unstable feature on stable because it's unstable only for `Rayon`(doesn't contain Rust nightly features). At least in Appveyor, `1.12.0` with the unstable feature enabled fails to build - that's why there is no matrix for `1.12.0` + `FEATURES=unstable`.

You can see the status of these builds:
- Windows: https://ci.appveyor.com/project/lilianmoraru/rayon
- Linux/macOS: https://travis-ci.org/lilianmoraru/rayon

On Windows, nightly fails and I don't know why(I don't have a Windows machine to test locally).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,12 @@ env:
 
 matrix:
   include:
+  - rust: stable
+    env: FEATURES=unstable
+    os: linux
+  - rust: stable
+    env: FEATURES=unstable
+    os: osx
   - rust: nightly
     env: FEATURES=unstable
     os: linux
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,61 @@
+environment:
+  RUST_MIN_STACK: 16777216
+  matrix:
+    - TARGET: x86_64-pc-windows-gnu
+      CHANNEL: 1.12.0
+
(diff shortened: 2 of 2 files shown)


cuviper:
> On Windows, nightly fails and I don't know why(I don't have a Windows machine to test locally).

Strange!  When I run `cargo test` on Windows 10, I get three popup dialogs:

image
image
image

It also happens on beta and stable if I cheat the feature usage with `set RUSTC_BOOTSTRAP=1`.

But the tests for rayon-core run ok.  I suspect it's something to do with `compiletest-rs`.

We were saying in #277 that it would be nice to run most tests on non-nightly rust too, so one part of that is we'd have to mask `compiletest-rs` usage to just nightly.  Under that idea, we could also avoid it on Windows for now.

lilianmoraru:
I amended the commit and removed the first `cargo test` - now it seems to work on nightly.
I still think that should probably be a separate issue - many people use nightly.

nikomatsakis:
OK, well, I agree, that is surprising and should be investigated -- but it seems like we ought to enable these builds at least. Thanks @lilianmoraru !

PR #322, opened 2017-04-21
Add Conway's Game of Life benchmark

crazymykl:


rayon-demo/src/life/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    pub fn next_generation(&self) -> Board {
+        let new_brd = (0..self.len()).map(|cell| self.successor_cell(cell)).collect();
+
+        self.next_board(new_brd)
+    }
+
+    pub fn parallel_next_generation(&self) -> Board {
+        let new_brd = (0..self.len())
+            .collect::<Vec<usize>>()
```
nikomatsakis: why the call to `collect()` here, rather than just using `into_par_iter` on the range directly?

rayon-demo/src/life/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    fn len(&self) -> usize {
+        self.rows * self.cols
+    }
+
+    fn next_board(&self, new_board: Vec<bool>) -> Board {
+        assert!(new_board.len() == self.len());
+
+        Board { board  : new_board,
+                born   : self.born.clone(),
```
nikomatsakis: Seems like these are immutable, right? Maybe make them `Arc<Vec<usize>>` for cheaper cloning?

rayon-demo/src/life/mod.rs
crazymykl:
```
   Compiling rayon-demo v0.1.0 (file:///Users/mmacdonald/workspace/rayon/rayon-demo)
error: no method named `par_iter` found for type `std::ops::Range<usize>` in the current scope
  --> src/life/mod.rs:88:14
   |
88 |             .par_iter()
   |              ^^^^^^^^
   |
   = note: the method `par_iter` exists but the following trait bounds were not satisfied: `&std::ops::Range<usize> : rayon::iter::IntoParallelIterator`

error: Could not compile `rayon-demo`.
```

rayon-demo/src/life/mod.rs
crazymykl: Yeah, I even think they can be Rc, since they aren't touched by the parallel bit.

cuviper: It needs to be `Arc` for threading.

rayon-demo/src/life/mod.rs
cuviper: Specifically `into_par_iter()` for iterating by value.

rayon-demo/src/life/mod.rs
crazymykl: Yeah, I noticed in my edit. Fixed now.

Issue #235, opened 2017-03-03
[docs] the `ThreadPool` struct in `rayon-core` needs documentation


nikomatsakis:
The `ThreadPool` struct represents a user created thread-pool (as opposed to the default, global threadpool). It needs `/// foo` docs.

Some things to say:

- Create a new thread-pool with `ThreadPool::new` (!). You can use the configuration to specify the number or names of its threads, how to handle uncaptured panics, and so forth.
- There are two ways to use the thread-pool:
    - The first is that you can execute `ThreadPool::install()`. This will execute a closure in one of the threadpool's threads. Any other Rayon operations -- like parallel iterators, `join()`, or `scope()` -- will then execute in the context of this thread-pool.
    - The second is that you invoke `ThreadPool::spawn_async()` or `ThreadPool::spawn_future_async()`. These methods will inject jobs into the thread-pool that are not tied to your current stack frame; however, those jobs cannot access data owned by the stack.
- When the thread-pool is dropped, that is a signal for its threads to terminate. Once they have completed executing any remaining work that you have spawned, they will automatically terminate.

bobbo:
Can this be closed given #325 was merged?

cuviper:
Sounds good, thanks!

Enamex:
Are there any non-doc [help wanted] issues remaining or coming?

cuviper:
@Enamex I would say docs are where we're most lacking, but I'll try to remember to file any other ideas as I have them.  I just filed #465, and maybe you could check if someone is still working on #330.

There are also open questions, like how to create parallel iterators from arbitrary iterators (#46) or vice versa (#210).  I guess those could be considered [help wanted] in the sense of "help, we don't even know how to do this!"  We tend to only tag [help wanted] on mentor-able issues, but anyone is welcome to chime in on other issues too.

PR #325, opened 2017-04-26
Threadpool documentation

leshow:
This addresses #235 

Let me know if this is along the lines of what you were looking for. I made an assumption that `rayon::join` called inside of ThreadPool's `spawn` methods will also execute inside of the ThreadPool context, like `install`. If this isn't the case, I'll alter the second example and amend this commit. The example for `spawn_future_async` is based on code from your test cases.

### DIFF
--- a/rayon-core/src/thread_pool/mod.rs
+++ b/rayon-core/src/thread_pool/mod.rs
@@ -98,10 +98,8 @@ impl ThreadPool {
     ///    # use rayon_core as rayon;
     ///    fn main() {
     ///         let pool = rayon::ThreadPool::new(rayon::Configuration::new().num_threads(8)).unwrap();
-    ///         pool.install(|| {
-    ///                          let n = fib(20);
-    ///                          println!("{}", n);
-    ///                      });
+    ///         let n = pool.install(|| fib(20)); 
+    ///         println!("{}", n);
     ///    }
     ///
     ///    fn fib(n: usize) -> usize {
@@ -201,7 +199,7 @@ impl ThreadPool {
     /// the current stack frame -- therefore, it cannot capture any references
     /// onto the stack (you will likely need a `move` closure).
     ///
-    /// See the [`spawn()` method defined on scopes][spawn] for more details.
+    /// See also: [the `spawn()` function defined on scopes][spawn].
     ///
     /// [spawn]: struct.Scope.html#method.spawn
     #[cfg(feature = "unstable")]
@@ -225,28 +223,22 @@ impl ThreadPool {
     /// ```rust
     ///    # extern crate rayon_core as rayon;
     ///    extern crate futures;
-    ///    
     ///    use futures::{future, Future};
-    ///    use std::sync::{Arc, Mutex};
     ///    # fn main() {
-    ///    let data = Arc::new(Mutex::new(String::from("Hello,")));
+    ///
     ///    let pool = rayon::ThreadPool::new(rayon::Configuration::new().num_threads(8)).unwrap();
     ///
-    ///    let a = pool.spawn_future(future::lazy({
-    ///                                               let data = data.clone();
-    ///                                               move || Ok::<_, ()>(data)
-    ///                                           }));
-    ///    let b = pool.spawn_future(a.map(|data| {
-    ///                                               let mut v = data.lock().unwrap();
-    ///                                               v.push_str(" world!");
-    ///                                           }));
-    ///    let () = b.wait().unwrap();
-    ///    println!("{:?}", &data.lock().unwrap()[..]);
-    ///    // prints: "Hello, world!"
+    ///    let a = pool.spawn_future(future::lazy(move || Ok::<_, ()>(format!("Hello, "))));
+    ///    let b = pool.spawn_future(a.map(|mut data| {
+    ///                                        data.push_str("world");
+    ///                                        data
+    ///                                    }));
+    ///    let result = b.wait().unwrap(); // `Err` is impossible, so use `unwrap()` here
+    ///    println!("{:?}", result); // prints: "Hello, world!"
     ///    # }
     /// ```
     ///
-    /// See the [`spawn_future()` method defined on scopes][spawn_future] for more details.
+    /// See also: [the `spawn_future()` function defined on scopes][spawn_future].
     ///
     /// [spawn_future]: struct.Scope.html#method.spawn_future
     #[cfg(feature = "unstable")]
--- a/rayon-core/src/thread_pool/mod.rs
+++ b/rayon-core/src/thread_pool/mod.rs
@@ -14,101 +14,37 @@ use std::error::Error;
 use registry::{Registry, WorkerThread};
 
 mod test;
-/// ## ThreadPool
+/// # ThreadPool
 ///
-/// `ThreadPool` represents a user crated [threadpool]. [`ThreadPool::new()`]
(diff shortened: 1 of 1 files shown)


leshow:
I'm not quite sure why the docs are failing to compile, if someone could point out how to fix that I can amend the commit.

```
---- src/thread_pool/mod.rs - thread_pool::ThreadPool (line 28) stdout ----
	error[E0463]: can't find crate for `rayon`
 --> <anon>:2:4
  |
2 |    extern crate rayon;
  |    ^^^^^^^^^^^^^^^^^^^ can't find crate
```

cuviper:
It's a little funny because of the way we expose rayon-core.  You can follow [this example](https://github.com/nikomatsakis/rayon/blob/ad980b45c956375e2905aa6f479c48846d1d78b3/rayon-core/src/scope/mod.rs#L57).

leshow:
Lovely. I have an example of `spawn_future_async`, which of course uses futures. Is there a way to get that to work also?

cuviper:
Ah - not easily since that's behind an unstable feature flag.  Probably the best thing is to move that conditional example onto `spawn_future_async` itself.

leshow:
I don't believe the `# use rayon_core as rayon;` is working properly either. In particular, in the example I
`use rayon::{ThreadPool, Configuration};` and it doesn't appear to like that.

nikomatsakis:
@leshow this looks like great work, thanks. I have one question: I have this PR that renames the various "async" methods (#313) that I had hoped to land soon. Would you mind updating the docs to reflect the new naming conventions?

rayon-core/src/thread_pool/mod.rs
```diff
@@ -12,7 +12,49 @@ use std::error::Error;
...
+///    use rayon::{ThreadPool, Configuration};
+///
+///    let pool = ThreadPool::new(Configuration::new().num_threads(8)).unwrap();
+/// ```
+///
+/// [`install()`] executes a closure in one of the `ThreadPool`'s threads. In addition, 
+/// any other rayon operations called inside of `install()` (such as `rayon::join()`)
+/// will also execute in the context of the `ThreadPool`.
+///
+/// The other way to execute a closure in the `ThreadPool` is to invoke [`ThreadPool::spawn_async()`]
```
nikomatsakis: I think these methods are currently "feature-gated", and https://github.com/nikomatsakis/rayon/pull/313 renames them in any case. You might want to mention the unstable feature, or move the comments onto the method itself for now. 

In general, #313 adds a number of other methods here -- e.g., `join`, `spawn`, that seem relevant though. We should think about the best way to present them. The general idea is that all of the top-level functions that rayon offers will be mirrored in the `ThreadPool` API, which will then execute them in a specific thread-pool. (In contrast, the top-level functions execute (implicitly) in the *current* thread-pool.)

leshow:
@nikomatsakis Absolutely, I'll have a look at #313  and rename them appropriately. 

I added the initial preamble because I thought it would be nice to have a simple example at the top of `ThreadPool` so users of rayon would know how things fit together. If it makes more sense to move everything to their corresponding methods I will do that.

leshow:
I'm having trouble with the imports for example code still.

```rust
# use rayon_core as rayon;
use rayon::{Configuration, ThreadPool};
```
does not work, I have to `use rayon_core:{Configuration, ThreadPool}` so that it can find `ThreadPool`.

Additionally, I can't get `extern crate futures` recognized.

nikomatsakis:
@leshow 

> I'm having trouble with the imports for example code still.

Huh, curious. My guess is that it has to do with exactly how rustdoc generates the code. I'm actually a bit surprised that my initial hack worked, now that I think about it, but it must be that the `use` winds up generated (in my code) at the root module, but not in your examples.

nikomatsakis:
Ah, well, I guess I'm using things like `rayon::scope` in an expression context, where the paths are *relative*, but you are trying to do `use rayon::`, where the paths are *absolute*. Hence the difference. It seems like we should either give up and just do `use rayon_core`, or else hide the `use` statements altogether.

cuviper:
Or use explicit `rayon::ThreadPool` and `rayon::Configuration` in expressions here too.

leshow:
After amending my commit about 20 different times trying different combinations of imports, I can across this:

https://github.com/rust-lang/cargo/issues/960

To use something like `extern crate futures` I need to include a `fn main()` declaration and either hide it with `#` or have it part of the example. So that's the futures issue taken care of.

Unfortunately on the nightly+unstable build, `# use rayon_core as rayon` says that rayon_core can't be found.

nikomatsakis:
@leshow 

> To use something like extern crate futures I need to include a fn main() declaration and either hide it with # or have it part of the example. So that's the futures issue taken care of.

Ah yes good catch.

> Unfortunately on the nightly+unstable build, # use rayon_core as rayon says that rayon_core can't be found.

Grr. Try adding `extern crate rayon_core as rayon` maybe, instead of the `use`?

leshow:
Ha! That did it! thanks @nikomatsakis

rayon-core/src/thread_pool/mod.rs
```diff
@@ -61,6 +91,27 @@ impl ThreadPool {
...
     ///
     /// If `op` should panic, that panic will be propagated.
+    ///
+    /// ## Using `install()`
+    ///  
+    /// ```rust
+    ///    # use rayon_core as rayon;
+    ///    fn main() {
+    ///         let pool = rayon::ThreadPool::new(rayon::Configuration::new().num_threads(8)).unwrap();
+    ///         pool.install(|| {
```
nikomatsakis: This would look mildly simpler if we did

```rust
let n = pool.install(|| fib(20));
println!("{}", n);
```

rayon-core/src/thread_pool/mod.rs
```diff
@@ -161,8 +212,43 @@ impl ThreadPool {
...
+    ///    use std::sync::{Arc, Mutex};
+    ///    # fn main() {
+    ///    let data = Arc::new(Mutex::new(String::from("Hello,")));
+    ///    let pool = rayon::ThreadPool::new(rayon::Configuration::new().num_threads(8)).unwrap();
+    ///
+    ///    let a = pool.spawn_future(future::lazy({
+    ///                                               let data = data.clone();
+    ///                                               move || Ok::<_, ()>(data)
+    ///                                           }));
+    ///    let b = pool.spawn_future(a.map(|data| {
```
nikomatsakis: This is a bit of a funny example -- in particular, futures are best used when they **produce a result that you want to consume**, but this example works via side-effects. I would recommend that code like this be written with plain spawns, rather than futures (in particular, the cancellation semantics of futures could lead to surprising interactions).

Maybe rewrite this to the following:

```rust
let a = pool.spawn_future(future::lazy(move || Ok::<_, ()>(format!("Hello, world!")));
let b = pool.spawn_future(a.map(|mut data| { data.push_str("world"); data }));
let result = b.wait().unwrap(); // `Err` is impossible, so use `unwrap()` here
println!("{:?}", result); // prints: "Hello, world!"
```

leshow: I took this example from one of the test cases for `spawn_future`. (https://github.com/nikomatsakis/rayon/blob/master/rayon-core/src/spawn/test.rs#L53) I should have given more thought in hindsight though. Thank you for spending the time on this.

leshow:
Thanks, I've made those changes. I also amended the text for links to use the same wording that's ubiquitous throughout the docs already.

nikomatsakis:
Nice :)

PR #326, opened 2017-04-26
Add `par_extend`, `partition`, `unzip`, and `unzip_into`

cuviper:
- `ParallelExtend` mirrors `std::iter::Extend`, with one method `par_extend`.  It works basically like `FromParallelIterator`, except it adds to an existing container instead of creating a new one.  In fact, most of the `FromParallelIterator`s we implement now just create an empty container and then `par_extend` it.
- `ParallelIterator::partition` mirrors `Iterator::partition`, separating the iterator into two `ParallelExtend` containers based on the user's `predicate` function.
- `ParallelIterator::unzip` mirrors `Iterator::unzip`, separating an iterator over `(A, B)` into two `ParallelExtend` containers, one with just `A` and the other with just `B`.
- `IndexedParallelIterator::unzip_into` is analogous to `collect_into`, writing directly into the target vectors.


cuviper:
@nikomatsakis In the [wiki](https://github.com/nikomatsakis/rayon/wiki/Iterator-Trait-Parity) you said that `Iterator::partition` really ought to allow two distinct output types.  It wouldn't be very hard for me to do that here if you want.

cuviper:
The nightly failures are rust-lang/rust#41479.  It does work with `nightly-2017-04-21`.

src/iter/unzip.rs
```diff
@@ -0,0 +1,236 @@
...
+                left_result: &mut result,
+            };
+            self.b.par_extend(iter);
+        }
+        result.unwrap()
+    }
+
+    fn opt_len(&mut self) -> Option<usize> {
+        self.base.opt_len()
+    }
```
cuviper: This one got me thinking.  It's fine to leak the source-iterator's index-ness through `opt_len()` here, and it should result in the optimization we want.  However, with an eye on moving that mechanism to specialization, I can't actually implement a true `IndexedParallelIterator` for this!

I could implement `IndexedParallelIterator::drive()`, but not `with_producer()` because of the iterator requirement in `Producer`.  We need to stay in "push" mode here to feed values into multiple consumers, but once we flip to an iterator we're left pulling.  And that means we can't support `zip` et al. on this -- which I wouldn't expect in a `par_extend()` implementation anyway, but with specialization one *could* encounter that.

Maybe this is a distinction which deserves to split up the traits again.  An `IndexedParallelIterator` that can `drive()`, `enumerate()`, etc., and a subset `ProducableParallelIterator` that supports `with_producer()`, `zip()`, etc.

Ugh.

(Incidentally, this "push vs. pull" is basically why `Iterator::unzip` requires `Extend` instead of `FromIterator`.)

cuviper:
There's also [`Itertools::partition_map`](https://docs.rs/itertools/0.6.0/itertools/trait.Itertools.html#method.partition_map) we could emulate, which can map to different item types and allows different collection types.  That would be yet another full implementation, where `unzip.rs` and `partition.rs` are already very similar... maybe they can be merged and abstracted so that only the `Folder::consume` is really different.  Kind of like we used to with `MapOp`, but this code is trickier and also hidden from the API, so it's more favorable to sharing.

cuviper:
I merged that common functionality, generalized `partition`, and added `partition_map`.

cuviper:
Added `FromParallelIterator<&char> for String` since Rust 1.17 now has it -- rust-lang/rust#40028.
(not directly related to this PR, but it touches the same code as extend, so consider it a bonus...)

src/iter/unzip.rs
```diff
@@ -0,0 +1,110 @@
+use super::internal::*;
+use super::*;
+
+
+/// Unzip an `IndexedParallelIterator` into two arbitrary `Consumer`s.
+pub fn unzip_indexed<I, A, B, CA, CB>(pi: I, left: CA, right: CB)
```
nikomatsakis: Are these intended to be Rayon public APIs? (I'd appreciate a comment if not.)

src/iter/collect/mod.rs
```diff
@@ -39,6 +40,26 @@ fn special_extend<I, T>(pi: I, len: usize, v: &mut Vec<T>)
     collect.complete();
 }
 
+/// Unzips the results of the exact iterator into the specified vectors.
+pub fn unzip_into<I, A, B>(mut pi: I, left: &mut Vec<A>, right: &mut Vec<B>)
```
nikomatsakis: Are these intended to be Rayon public APIs? (I'd appreciate a comment if not.)

src/iter/unzip.rs
nikomatsakis: Interesting. It's also worth considering whether using specialization is worth it versus the current `opt_len()` trick.

src/iter/partition.rs
```diff
@@ -0,0 +1,217 @@
...
+          P: Fn(&I::Item) -> bool + Sync
+{
+    type Item = I::Item;
+
+    fn drive_unindexed<CM>(self, matches: CM) -> CM::Result
+        where CM: UnindexedConsumer<Self::Item>
+    {
+        let mut result = None;
+        {
+            // Now it's time to find the consumer for non-matching items
```
nikomatsakis: Clever trick, I will say. I wonder if we can use some internal traits + generics to "package up" this pattern rather than repeating it. It would probably make things less clear though, although it would provide a good place to document what is going on.

src/iter/unzip.rs
```diff
@@ -1,15 +1,24 @@
 use super::internal::*;
 use super::*;
 
+trait UnzipOp<T>: Sync {
```
nikomatsakis: I think perhaps you were a step ahead of me. Good place for comments, though!

src/iter/mod.rs
```diff
@@ -649,8 +649,9 @@ pub trait ParallelIterator: Sized {
     /// Partitions the items of a parallel iterator into a pair of arbitrary
     /// `ParallelExtend` containers.  Items for which the `predicate` returns
     /// true go into the first container, and the rest go into the second.
-    fn partition<C, P>(self, predicate: P) -> (C, C)
-        where C: Default + ParallelExtend<Self::Item>,
+    fn partition<A, B, P>(self, predicate: P) -> (A, B)
+        where A: Default + ParallelExtend<Self::Item>,
+              B: Default + ParallelExtend<Self::Item>,
```
nikomatsakis: Yay! My only hesitation is that it means that transition to rayon *may* require more type annotations. e.g., with seq iterators, people can do:

```rust
let (x, y): (Vec<_>, _) = iter.partition();
```

maybe we should add another method for this case? (We could then propose the same method for seq iterators, I suppose). But I don't know what would be a good name.

src/iter/mod.rs
```diff
@@ -650,6 +656,19 @@ pub trait ParallelIterator: Sized {
         unzip::partition(self, predicate)
     }
 
+    /// Partitions and maps the items of a parallel iterator into a pair of
+    /// arbitrary `ParallelExtend` containers.  `Either::Left` items go into
+    /// the first container, and `Either::Right` items go into the second.
+    fn partition_map<A, B, P, L, R>(self, predicate: P) -> (A, B)
```
nikomatsakis: This is meant for parity with itertools, I guess? Makes sense. I was wondering if we should link directly to itertools, so we can use the same `Either` type, but it's problematic that it is at only 0.6 I suppose, and it seems a *bit* silly to add an itertools dependency just for `Either`.

(cc @bluss -- as an aside, any reason itertools can't be 1.0?)

nikomatsakis:
So the only real question that came up in my review is whether to have `partition()` return two collections or not. It depends how much we care about compatibility. We could include a note in the docs that our signature is more flexible that the iterator one and hence may require an additional type annotation. I think I'm inclined to do that for now, and just leave it the way you have it.

(I have definitely needed two distinct collection types in the past and been annoyed that I can't have it.)

cuviper:
Yeah, I'm ok with the possible ambiguity, since it would only bite at compile time, not a runtime surprise, and it's simple to correct.

I'll add more comments where requested.

nikomatsakis:
r=me, after rebase

bluss:
Sorry for the very late reply :smile:. Itertools is working towards 1.0, it doesn't feel very far off.

bluss:
That said we don't have a "tracking issue" for it. One reason might be that if std adopts anything from itertools using the same method name, we have a problem.

PR #332, opened 2017-05-06
Use the either crate for `Either`

cuviper:
We added a bare `Either` type for `partition_map`, but that's not really a type that belongs in Rayon.  The `either` crate has a stable implementation of `Either` for everyone to share.

While we're at it, `Either` implements pass-throughs for `Iterator` and `Extend`, so we can do the same for `ParallelIterator` and `ParallelExtend`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ exclude = ["ci"]
 
 [dependencies]
 rayon-core = { version = "1.2", path = "rayon-core" }
+either = "1.1"
 
 [dev-dependencies]
 compiletest_rs = "0.2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,11 @@ exclude = ["ci"]
 
 [dependencies]
 rayon-core = { version = "1.2", path = "rayon-core" }
-either = "1.1"
+
+# This is a public dependency!
+[dependencies.either]
+version = "1.0"
+default-features = false
 
 [dev-dependencies]
 compiletest_rs = "0.2.1"
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -9,7 +9,7 @@
 //! the code itself, the `internal` module and `README.md` file are a
 //! good place to start.
 
-use either::Either;
+pub use either::Either;
 use std::cmp::{self, Ordering};
 use std::iter::{Sum, Product};
 use std::ops::Fn;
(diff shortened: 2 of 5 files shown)


cuviper:
FWIW, `itertools` has a `pub use either::Either`, so we'll be using a compatible type.  I'm not sure if we should follow suit and reexport it too.  It becomes part of our public API anyway for `partition_map`, so I guess the question is whether we want folks to be forced to get `crate either` themselves.
(cc @bluss)

cuviper:
We could also make this optional, probably enabled by default, and gate `partition_map` entirely.

nikomatsakis:
I am ok with requiring people to compile `Either`.

nikomatsakis:
Features are annoying. =)

bluss:
Seems fine to use IMO.

Design choices for which methods to add to Either are really hard, it's already “gone wrong” in the sense of choosing to implement some special case methods which excludes some competing alternatives. It has niched itself into being a stock `Iterator` of two cases, so it makes sense to continue that with `ParallelIterator` if you want.

cuviper:
If we want to take advantage of the chance to design rayon differently, and not be directly analogous, I think I would rather have `IntoParallelIterator` implementations into distinct `ParallelIterator` types, `rayon::either::Iter` etc.  The `Extend` and `ParallelExtend` seem fine though.

@bluss If you regret that design, maybe it's worth making `either 2.0`?  Better to make such a change sooner, or not at all...

bluss:
It's a design that's possible to regret since it picks one road other another, but it's fine.

nikomatsakis:
So -- should we do this? I think so.

cuviper:
> So -- should we do this? I think so.

Which "this"?  The PR as-is, or with custom `Iter` etc.?

Also, do you want `Either` re-exported?  Then this isn't a breaking change.

nikomatsakis:
> Which "this"? The PR as-is, or with custom Iter etc.?

Good question =)

I think I meant mostly "use `Either` at all".

cuviper:
Rebased and made it a `pub use`.

nikomatsakis:
1.12.0 fails:

```
error[E0252]: a type named `Either` has already been imported in this module
 --> src/iter/unzip.rs:3:5
  |
1 | use either::Either;
  |     -------------- previous import of `Either` here
2 | use super::internal::*;
3 | use super::*;
  |     ^^^^^^^^^ already imported
```

cuviper:
Doh, should be fixed now.

PR #334, opened 2017-05-10
Introduce RAYON_NUM_THREADS and RAYON_LOG env vars

SuperFluffy:
I hope this PR doesn't appear like bikeshedding, but I believe `RAYON_RS_NUM_THREADS` is a better, more mnemonic name for the environment variable compared with `RAYON_RS_NUM_CPUS`.

Reasons:

1. It corresponds to what the environment variable is supposed to be controlling, namely the *number of threads*; from `rayon-core/src/lib.rs`:

```rust
pub struct Configuration {
    /// The number of threads in the rayon thread pool.
    /// If zero will use the RAYON_RS_NUM_CPUS environment variable.
    /// If RAYON_RS_NUM_CPUS is invalid or zero will use the default.
num_threads: usize,
...
}
```

2. The variable name would correspond to what OpenMP is using, which is `OMP_NUM_THREADS`. Since rayon is very likely going to be used for the same purposes that OpenMP is used in other languages, this would be a more natural name to remember (ok, `RAYON_NUM_THREADS` would be even better. :-) ).

### DIFF
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -97,8 +97,8 @@ pub fn current_num_threads() -> usize {
 #[derive(Default)]
 pub struct Configuration {
     /// The number of threads in the rayon thread pool.
-    /// If zero will use the RAYON_RS_NUM_CPUS environment variable.
-    /// If RAYON_RS_NUM_CPUS is invalid or zero will use the default.
+    /// If zero will use the RAYON_NUM_THREADS environment variable.
+    /// If RAYON_NUM_THREADS is invalid or zero will use the default.
     num_threads: usize,
 
     /// Custom closure, if any, to handle a panic that we cannot propagate
@@ -144,6 +144,13 @@ impl Configuration {
         if self.num_threads > 0 {
             self.num_threads
         } else {
+            match env::var("RAYON_NUM_THREADS").ok().and_then(|s| usize::from_str(&s).ok()) {
+                Some(x) if x > 0 => return x,
+                Some(x) if x == 0 => return num_cpus::get(),
+                _ => {},
+            }
+
+            // Support for deprecated `RAYON_RS_NUM_CPUS`.
             match env::var("RAYON_RS_NUM_CPUS").ok().and_then(|s| usize::from_str(&s).ok()) {
                 Some(x) if x > 0 => x,
                 _ => num_cpus::get(),
@@ -173,7 +180,7 @@ impl Configuration {
     /// If `num_threads` is 0, or you do not call this function, then
     /// the Rayon runtime will select the number of threads
     /// automatically. At present, this is based on the
-    /// `RAYON_RS_NUM_CPUS` environment variable (if set), 
+    /// `RAYON_NUM_THREADS` environment variable (if set), 
     /// or the number of logical CPUs (otherwise). 
     /// In the future, however, the default behavior may
     /// change to dynamically add or remove threads as needed.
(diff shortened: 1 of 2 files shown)


rayon-core/src/lib.rs
```diff
@@ -144,7 +144,7 @@ impl Configuration {
         if self.num_threads > 0 {
             self.num_threads
         } else {
-            match env::var("RAYON_RS_NUM_CPUS").ok().and_then(|s| usize::from_str(&s).ok()) {
+            match env::var("RAYON_RS_NUM_THREADS").ok().and_then(|s| usize::from_str(&s).ok()) {
```
cuviper: Maybe we should keep checking the old name too for compatibility?

cuviper:
> (ok, RAYON_NUM_THREADS would be even better. :-) )

I would be happy with that too.  @nikomatsakis ?

rayon-core/src/lib.rs
SuperFluffy: Sure, I can check for both. Should one be treated preferrentiably over the other?

Also, do you think it's a big concern? a) rayon is pre-1.0 :), b) how many people are blindly going to keep using the same env var after rayon hits 1.0 (which I heard is planned for soonish)? I think advertising the change in a prominent position should be fine?

cuviper: Certainly give priority to the new and documented name.

This is in rayon-core, which *is* 1.0 already.  Environment variables are a gray area in an API, but since it was documented, and it's easy to keep the compatible fallback, I think we should.

rayon-core/src/lib.rs
```diff
@@ -144,6 +144,13 @@ impl Configuration {
...
             self.num_threads
         } else {
+            match env::var("RAYON_RS_NUM_THREADS").ok().and_then(|s| usize::from_str(&s).ok()) {
+                Some(x) if x > 0 => return x,
+                Some(x) if x == 0 => return num_cpus::get(),
+                _ => {},
+            }
+
+            // Support for deprecated `RAYON_RS_NUM_CPUS`.
             match env::var("RAYON_RS_NUM_CPUS").ok().and_then(|s| usize::from_str(&s).ok()) {
```
SuperFluffy: I am not sure if a message along the lines of “`RAYON_RS_NUM_CPUS` is deprecated and will be removed in the future; please use `RAYON_RS_NUM_THREADS`” would be appropriate? At first I intended to move this match into a function, but then I realize that the `#[deprecated]` attribute is a compile time feature.

nikomatsakis: Meh, I'd probably just leave it. However, to be strictly backwards compatible, shouldn't we prefer `RAYON_RS_NUM_CPUS` if both are given?

nikomatsakis:
Sure, new name seems better.

SuperFluffy:
@nikomatsakis @Cuviper [pinged](https://github.com/nikomatsakis/rayon/pull/334#issuecomment-300561531) about `RAYON_NUM_THREADS`. Do you want to keep the `RAYON_RS_NUM_THREADS` form or remove `_RS` suffix? I guess for consistency one should then also change `RAYON_RS_LOG` to `RAYON_LOG`.

Given that there doesn't seem to be any other software even remotely occupying the same namespace, maybe dropping `_RS` makes sense? After all, I always think about using *rayon*, not *rayon.rs*.

nikomatsakis:
Ah, I missed that. I guess that's fine. I mostly wanted something that was **unambiguously talking to this particular Rayon**, hence the `_RS_`. I get a bit nervous about environment variables affecting our behavior in major ways (but it's so dang handy sometimes...). But I can get behind `RAYON_XXX` as a consistent set of variables.

nikomatsakis:
Are we going to change to `RAYON` rather than `RAYON_RS`, or should I merge this? I'm fine either way, really.

cuviper:
I say we stake a claim on `RAYON`. :)

SuperFluffy:
Before I start spamming you with notifications about fixed commits, what's the decision on @nikomatsakis backwards compatibility, https://github.com/nikomatsakis/rayon/pull/334#discussion_r115860390?

> Meh, I'd probably just leave it. However, to be strictly backwards compatible, shouldn't we prefer RAYON_RS_NUM_CPUS if both are given?

nikomatsakis:
@SuperFluffy I'm fine either way.

SuperFluffy:
@nikomatsakis Changes pushed and rebased on top of recent master.

nikomatsakis:
@SuperFluffy thanks! :)

PR #336, opened 2017-05-11
Duplicate part of the {Start,Exit}Handler doc to Configuration::{start,exit}_handler

### LINKED ISSUE #333: [doc] Configuration::*_handler() doc doesn't specify what the usize argument to the callback is for
I guess it's a thread id, but it should be spelled out.

glandium:
Fixes #333


cuviper:
Looks good, thanks!

Issue #331, opened 2017-05-01
Add an std::iter::repeat() equivalent


cuviper:
```rust
pub fn repeat<T>(elt: T) -> Repeat<T>
    where T: Clone
```

"Creates a new iterator that endlessly repeats a single element."

It should be straightforward in rayon with `T: Clone + Send`.  I think we could have both an indexed `Producer` (len `usize::MAX`) for use with `zip` etc. and an infinite `UnindexedProducer`, depending on whether it goes through `drive`/`with_producer` or `drive_unindexed`.

The novelty `std::iter::once()` and `std::iter::empty()` could also be implemented, probably just wrapping `Option::Some` and `None`.

ChristopherDavenport:
I'll take a run at this one.

Enamex:
Doesn't #397 close this?

cuviper:
Yep, thanks!

PR #337, opened 2017-05-13
Add Repeat

ChristopherDavenport:
Creates a Parallel Iterator that endlessly repeats a single element

I am sure this needs more testing, however I don't know the best way. I believe I have a working implementation.

Refs #331

### DIFF
--- a/src/iter/repeat.rs
+++ b/src/iter/repeat.rs
@@ -1,11 +1,6 @@
-use super::internal::*;
 use super::*;
-use iter::*;
-use iter::internal::*;
-use std;
-use std::iter;
+use super::internal::*;
 use std::usize;
-
 
 pub struct Repeat<T> {
     element: T,
--- a/src/iter/repeat.rs
+++ b/src/iter/repeat.rs
@@ -1,0 +1,113 @@
+use super::internal::*;
+use super::*;
+use iter::*;
+use iter::internal::*;
+use std;
+use std::iter;
+use std::usize;
+
+
+pub struct Repeat<T> {
+    element: T,
+}
+
+pub fn repeat<T: Clone>(elt: T) -> Repeat<T> {
+    Repeat{element: elt}
+}
+
+impl<T: Clone> Clone for Repeat<T>{
+    fn clone(&self) -> Repeat<T> {
+        Repeat{ element: self.element.clone()}
+    }
+}
+
+impl<T> ParallelIterator for Repeat<T>
+    where T: Clone + Send
+{
+    type Item = T;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+       where C: UnindexedConsumer<Self::Item>
+   {
+       bridge(self, consumer)
+   }
+}
+
+impl<T> IndexedParallelIterator for Repeat<T>
+    where T: Clone + Send
+{
+    fn drive<C>(self, consumer: C) -> C::Result
+        where C: Consumer<Self::Item>
+    {
+            bridge(self, consumer)
+    }
+
+    fn with_producer<CB>(self, callback: CB) -> CB :: Output
+        where CB: ProducerCallback<Self::Item>
+        {
+            callback.callback(RepeatProducer{ repeat: self})
+        }
+
+    fn len(&mut self) -> usize { usize::MAX }
+}
+
+
+/// Producer
+
+struct RepeatProducer<P> {
+    repeat: Repeat<P>,
+}
+
+impl<T: Clone + Send> Producer for RepeatProducer<T> {
+    type Item = T;
+    type IntoIter = RepeatIter<T>;
+
+    fn into_iter(self) -> Self::IntoIter{
+        RepeatIter{repeat: self.repeat}
+    }
+
+    fn split_at(self, index: usize) -> (Self, Self) {
+        (
+            RepeatProducer{repeat: self.repeat.clone()},
+            RepeatProducer{repeat: self.repeat.clone()}
+        )
+    }
+}
+
+impl<T: Clone + Send> UnindexedProducer for Repeat<T> {
+    type Item = T;
+
+    fn split(self) -> (Self, Option<Self>) {
+        (Repeat{element: self.element.clone()}, Some(Repeat{element: self.element.clone()}))
+    }
+
+    fn fold_with<F>(self, folder: F) -> F where F: Folder<T> {
+        folder.consume_iter(RepeatIter{repeat: Repeat{element: self.element}})
+    }
+}
+
+/// Repeat Iter Create As Repeat Does Not Have ExactSizeIterator
+
+struct RepeatIter<T> {
+    repeat: Repeat<T>,
+}
+
+impl<T: Clone> Iterator for RepeatIter<T> {
+    type Item = T;
+
+    #[inline]
+    fn next(&mut self) -> Option<T> { Some(self.repeat.element.clone()) }
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) { (usize::MAX, None) }
+}
+
+impl<A: Clone> DoubleEndedIterator for RepeatIter<A> {
+    #[inline]
+    fn next_back(&mut self) -> Option<A> { Some(self.repeat.element.clone()) }
+}
+
+impl<T: Clone> ExactSizeIterator for RepeatIter<T> {
+    fn len(&self) -> usize {
+        usize::MAX
+    }
+}
--- a/src/iter/repeat.rs
+++ b/src/iter/repeat.rs
@@ -2,11 +2,25 @@ use super::*;
 use super::internal::*;
 use std::usize;
 
-pub struct Repeat<T> {
+/// Iterator adapter for [the `repeat()` function](fn.split.html).
+pub struct Repeat<T: Clone + Send> {
     element: T,
 }
 
-pub fn repeat<T: Clone>(elt: T) -> Repeat<T> {
+/// Creates a parallel iterator that endlessly repeats `elt` (by
+/// cloning it). Note that this iterator has "infinite" length, so
+/// typically you would want to use `zip` or `take` or some other
+/// means to shorten it.
+///
+/// Example:
+///
+/// ```
+/// use rayon::prelude::*;
+/// use rayon::iter::repeat;
+/// let x: Vec<(i32, i32)> = repeat(22).zip(0..3).collect();
+/// assert_eq!(x, vec![(22, 0), (22, 1), (22, 2)]);
+/// ```
+pub fn repeat<T: Clone + Send>(elt: T) -> Repeat<T> {
     Repeat { element: elt }
 }
 
@@ -45,8 +59,8 @@ impl<T> IndexedParallelIterator for Repeat<T>
 
 /// Producer
 
-struct RepeatProducer<P> {
-    repeat: P,
+struct RepeatProducer<T: Clone + Send> {
+    repeat: T,
 }
 
 impl<T: Clone + Send> Producer for RepeatProducer<T> {
@@ -104,6 +118,9 @@ impl<A: Clone> DoubleEndedIterator for RepeatIter<A> {
 
 impl<T: Clone> ExactSizeIterator for RepeatIter<T> {
     fn len(&self) -> usize {
+        // FIXME this is .. sort of wrong. After all, we produce
+        // *more* than `usize::MAX`, potentially. Is there a better
+        // alternative?
         usize::MAX
     }
 }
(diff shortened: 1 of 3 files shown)


src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
+use super::*;
+use super::internal::*;
+use std::usize;
+
+pub struct Repeat<T> {
+    element: T,
+}
+
+pub fn repeat<T: Clone>(elt: T) -> Repeat<T> {
+    Repeat{element: elt}
```
nikomatsakis: Nit: I'd format this as `Repeat { element: elt }`

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl<T: Clone> Clone for Repeat<T>{
+    fn clone(&self) -> Repeat<T> {
+        Repeat{ element: self.element.clone()}
```
nikomatsakis: Nit: same here

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+   }
+}
+
+impl<T> IndexedParallelIterator for Repeat<T>
+    where T: Clone + Send
+{
+    fn drive<C>(self, consumer: C) -> C::Result
+        where C: Consumer<Self::Item>
+    {
+            bridge(self, consumer)
```
nikomatsakis: Nit: indentation

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    fn with_producer<CB>(self, callback: CB) -> CB :: Output
+        where CB: ProducerCallback<Self::Item>
+        {
```
nikomatsakis: Nit: the `{` should align with `fn`

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+impl<T> IndexedParallelIterator for Repeat<T>
+    where T: Clone + Send
+{
+    fn drive<C>(self, consumer: C) -> C::Result
+        where C: Consumer<Self::Item>
+    {
+            bridge(self, consumer)
+    }
+
+    fn with_producer<CB>(self, callback: CB) -> CB :: Output
```
nikomatsakis: Nit: `CB::Output`

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+            RepeatProducer{repeat: self.repeat.clone()}
+        )
+    }
+}
+
+impl<T: Clone + Send> UnindexedProducer for Repeat<T> {
+    type Item = T;
+
+    fn split(self) -> (Self, Option<Self>) {
+        (Repeat{element: self.element.clone()}, Some(Repeat{element: self.element.clone()}))
```
nikomatsakis: Nit: only one of these needs to be a `clone`

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+    type IntoIter = RepeatIter<T>;
+
+    fn into_iter(self) -> Self::IntoIter{
+        RepeatIter{repeat: self.repeat}
+    }
+
+    fn split_at(self, index: usize) -> (Self, Self) {
+        (
+            RepeatProducer{repeat: self.repeat.clone()},
+            RepeatProducer{repeat: self.repeat.clone()}
```
nikomatsakis: Nit: only one of these needs to be a clone; the other can just be `repeat: self.repeat` and take ownership

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+
+struct RepeatProducer<P> {
+    repeat: Repeat<P>,
+}
+
+impl<T: Clone + Send> Producer for RepeatProducer<T> {
+    type Item = T;
+    type IntoIter = RepeatIter<T>;
+
+    fn into_iter(self) -> Self::IntoIter{
```
nikomatsakis: Nit: `Self::IntoIter {`

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) { (usize::MAX, None) }
+}
+
+impl<A: Clone> DoubleEndedIterator for RepeatIter<A> {
+    #[inline]
+    fn next_back(&mut self) -> Option<A> { Some(self.repeat.element.clone()) }
+}
+
+impl<T: Clone> ExactSizeIterator for RepeatIter<T> {
```
nikomatsakis: Argh. The need to have `ExactSizeIterator` implemented is frustrating; I'm not sure if this impl is *technically* obeying the intentions of this trait (although I guess it works fine in practice). I'm trying to remember why we need `ExactSizeIterator`; something to do with the reverse operation iirc.

cuviper: Because `std::iter::Zip` is only double-ended if its inputs are double-ended and exact-size.

src/iter/repeat.rs
cuviper: I'd suggest just running `rustfmt repeat.rs` to probably address all of the nits at once.

src/iter/test.rs
```diff
@@ -1608,3 +1608,14 @@ fn check_partition_map() {
...
     assert_eq!(b, "abcxyz");
 }
+
+#[test]
+fn check_repeat_eq() {
+    let v = vec!(4,4,4,4);
+    let mut fours: Vec<_> = repeat(4)
+    .take(4)
+    .zip(v)
+    .collect();
```
cuviper: Since your producer's `split_at` doesn't actually use the index, I don't think this `take(4)` actually works -- but the result is still limited by the `zip` length instead.  Try another test with just `repeat(4).take(4).collect()`, and it panics "too many values pushed to consumer".

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl<T> ParallelIterator for Repeat<T>
+    where T: Clone + Send
+{
+    type Item = T;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+       where C: UnindexedConsumer<Self::Item>
+   {
+       bridge(self, consumer)
```
cuviper: This should use `bridge_unindexed` instead, otherwise you're never using the `UnindexedProducer` implementation at all.

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+    fn len(&mut self) -> usize { usize::MAX }
+}
+
+
+/// Producer
+
+struct RepeatProducer<P> {
+    repeat: Repeat<P>,
```
cuviper: I wonder, why does this store the whole `Repeat` type, and not just the element?  It's a minor difference, but then we don't need to worry about `Clone` for `Repeat` itself.

src/iter/repeat.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    fn fold_with<F>(self, folder: F) -> F where F: Folder<T> {
+        folder.consume_iter(RepeatIter{repeat: Repeat{element: self.element}})
+    }
+}
+
+/// Repeat Iter Create As Repeat Does Not Have ExactSizeIterator
+
+struct RepeatIter<T> {
+    repeat: Repeat<T>,
```
cuviper: Same here, seems like this could just store the element `T`.

src/iter/repeat.rs
nikomatsakis: Yeah, I remembered it had something to do with `Zip`

ChristopherDavenport:
I have not forgotten about this, sorry about the long delay. Hopefully can get around to it by tomorrow evening or at latest this weekend.

src/iter/test.rs
ChristopherDavenport: Is there something I should do in order to get take to work effectively with this implementation. It seems to play nicely with others, but no so nicely by itself.

cuviper: The producer will need to track its effective length.  It should initialize to `usize::MAX` since that's what `len()` reported.  Then each time it splits, use the given index to determine the resulting length of each side.  And finally the iterator will also need to limit its length.

ChristopherDavenport: So basically despite this being infinite in this formulation, when managing the `IndexedParralellIterator` actually limit the size to usize::MAX and reduce as we consume. So we can split where necessary.

cuviper: Yes, exactly.  We can't use this as an `IndexedParallelIterator` unless we actually map it to indexes, which puts an upper bound on the length.

Well... actually if you want to be clever, you could track the length as `Option<usize>` and leave the rightmost part of the splits open-ended.  So a bounded segment with `Some(len)` splits into two bounded segments, and an unbounded `None` splits into `Some` on the left and `None` on the right.  This will still work with `take` because it splits and discards the (infinite) remainder.

src/iter/repeat.rs
```diff
@@ -0,0 +1,109 @@
+use super::*;
+use super::internal::*;
+use std::usize;
+
+pub struct Repeat<T> {
+    element: T,
+}
+
+pub fn repeat<T: Clone>(elt: T) -> Repeat<T> {
```
nikomatsakis: Can we add some docs to this?

src/iter/repeat.rs
```diff
@@ -0,0 +1,109 @@
+use super::*;
+use super::internal::*;
+use std::usize;
+
+pub struct Repeat<T> {
```
nikomatsakis: I think this should type should require `T: Clone + Send`

nikomatsakis: Also, it should probably have some documentation, since it is user-facing.

src/iter/repeat.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    fn len(&mut self) -> usize {
+        usize::MAX
+    }
+}
+
+
+/// Producer
+
+struct RepeatProducer<P> {
```
nikomatsakis: Nit: Can we make this `<T>` instead of `<P>`, so it matches the others? This confused me for a while, since I thought `P` must imply it is a producer or something. I'd also rather see `T: Clone + Send`.

src/iter/repeat.rs
```diff
@@ -0,0 +1,109 @@
...
+impl<A: Clone> DoubleEndedIterator for RepeatIter<A> {
+    #[inline]
+    fn next_back(&mut self) -> Option<A> {
+        Some(self.repeat.clone())
+    }
+}
+
+impl<T: Clone> ExactSizeIterator for RepeatIter<T> {
+    fn len(&self) -> usize {
+        usize::MAX
```
nikomatsakis: I'm still mildly nervous about this but I don't have a better suggestion. =) I feel like maybe we should revisit the way we handle `rev()` to see if we can lift this requirement; but I guess it's ok for now.

nikomatsakis:
@ChristopherDavenport do you still have time to make further changes here? (If not, that's fine, just want to know -- I can try to make the final tweaks myself.)

ChristopherDavenport:
I may need you to finish the final tweaks. 

I will not bore you with in depth details(sisters graduation, switching positions, 2 emergency applications needed asap for work). My experience with rust being so weak I haven't been able to yank myself out of my work, primary OSS long enough to finish this.

I apologize and if you don't finish it up I will fix it myself eventually, but the "eventually" is a problem I can't nail down at the moment.

nikomatsakis:
@ChristopherDavenport no worries at all! I very much appreciate the work you did so far.

src/iter/repeat.rs
```diff
@@ -0,0 +1,126 @@
...
+impl<T: Clone + Send> Producer for RepeatProducer<T> {
+    type Item = T;
+    type IntoIter = RepeatIter<T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        RepeatIter { repeat: self.repeat }
+    }
+
+    fn split_at(self, index: usize) -> (Self, Self) {
+        (RepeatProducer { repeat: self.repeat.clone() }, RepeatProducer { repeat: self.repeat })
```
cuviper: I think this is still wrong.  If I `repeat(x).take(4)`, then we'll get a `split_at(4)`, and this needs to remember to only produce 4 items on the left side.  The right side is debatable, whether we leave that as infinite/`usize::MAX`, or now reduce it to `MAX - 4` remaining.

(noted here before: https://github.com/nikomatsakis/rayon/pull/337#discussion_r116654223)

nikomatsakis: Ah yes, good point. On this note, I'm softening on `repeatn`. Seems like a useful building block.

nikomatsakis: Maybe we should just not offer `repeat` at all; you can use `repeatn(usize::MAX)` for all practical purposes.

nikomatsakis:
Closing in favor of #397

nikomatsakis:
Thanks very much @ChristopherDavenport for laying the foundation here!

Issue #335, opened 2017-05-11
replace the thread-id hack with `thread-id` crate


nikomatsakis:
Turns out there is a crate for getting the id of the current thread:

https://crates.io/crates/thread-id

We currently use a kind of [nasty hack instead](https://github.com/nikomatsakis/rayon/blob/master/src/iter/internal.rs#L146-L152), where we take the address of a thread-local variable. We should replace that with `thread-id`. 

I'm marking this as Help Wanted since it seems like a good starter issue.

cuviper:
Well... *maybe*.  As that hack notes, it really doesn't care about the actual value, just that it's unique per thread.  I fear that a call to get an OS thread identifier might hurt performance.  I guess that depends on how fast TLS is in comparision, but I also hope that Rust will someday use static TLS, which makes that something like just a register and offset.

Alternatively, our own `ThreadPool::current_thread_index()` might also suffice, although we'd also need a way to get the current thread pool, and use a default like `usize::MAX` if outside any pool.

nikomatsakis:
> Well... maybe. As that hack notes, it really doesn't care about the actual value, just that it's unique per thread. I fear that a call to get an OS thread identifier might hurt performance. I guess that depends on how fast TLS is in comparision, but I also hope that Rust will someday use static TLS, which makes that something like just a register and offset.

Rust already does use static TLS, at least on linux. I also believe that this is how `pthread_self()` works, but I agree it's worth benchmarking.

cuviper:
When I say static TLS, I'm referring to `-ftls-model=initial-exec`.  You can see this in `readelf -d` that the flags will include `STATIC_TLS`.  I've never seen this on a Rust executable, but even with GCC it's only the default for non-PIC compilation.  (The space for static TLS could get exhausted by dynamic library loads.)  Dynamic TLS is still native ELF, but IIRC it requires libc calls to get the thread area every time.

In glibc, `pthread_self()` returns the arch-specific `THREAD_SELF`, which [on x86_64](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/nptl/tls.h;h=9b8ad82550f9ed719a5a18cf16c5b9fc346cdd28;hb=HEAD#l172) is just an `fs` offset, much like static TLS.  Compared to dynamic TLS which must call glibc's `__tls_get_addr()`... hmm, ok `pthread_self()` might be better for us after all.

[...]

PR #347, opened 2017-05-24
Use `thread-id` in thief_id call

hodgesds:
Hi, I'm new and wanted to help out a little if I can. I might not have followed all the conversation in the related issue but happy to help out if I can.

Resolves: #335

### DIFF
--- a/src/iter/internal.rs
+++ b/src/iter/internal.rs
@@ -5,6 +5,7 @@
 
 use rayon_core::join;
 use super::IndexedParallelIterator;
+use thread_id;
 
 use std::cmp;
 use std::usize;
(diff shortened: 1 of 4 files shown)


src/iter/internal.rs
```diff
@@ -147,8 +148,7 @@ impl Splitter {
     fn thief_id() -> usize {
         // The actual `ID` value is irrelevant.  We're just using its TLS
         // address as a unique thread key, faster than a real thread-id call.
-        thread_local!{ static ID: bool = false }
-        ID.with(|id| id as *const bool as usize)
+        thread_id::get()
```
cuviper: The comment should be updated too, like:

    // The actual `ID` value is irrelevant, as long as it's unique.

hodgesds:
Thanks for the comments, I'll try to take a look at doing some performance testing tonight.

nikomatsakis:
As it happens, I just made [a wiki page](https://github.com/nikomatsakis/rayon/wiki/Benchmarking-Rayon) about benchmarking that you may find useful. I can try to do some measurements too.

nikomatsakis:
This is superceded by #354, right? I think I prefer that approach.

nikomatsakis:
(To be clear, I appreciate the PR @hodgesds -- just seems better if we can avoid the dependency and make things a bit cleaner in the bargain. =)

Sorry for the long delay btw.

PR #348, opened 2017-05-30
Fix #343: Rayon 0.7.0: unbounded spins in Sleep::no_work_found

### LINKED ISSUE #343: Rayon 0.7.0: unbounded spins in Sleep::no_work_found
Here's the scenario:

Two worker threads, single (timesliced) core, or for whatever reason, the OS is
in a situation where it only runs one of our threads at once

T1 is in sleepy state

T2 is looking for work and iterates through no_work_found.
Its |yields| value rises to ROUNDS_UNTIL_SLEEPY

At this point, it loops:
   calls get_sleepy, which returns false, because T1 is sleepy
   so no_work_found returns |yields| unchanged

No other thread can run, so the global state of the system
remains unchanged, and T2 spins around this loop until the end of
its timeslice.

julian-seward1:
Fix #343: Rayon 0.7.0: unbounded spins in Sleep::no_work_found

### DIFF
--- a/rayon-core/src/sleep/mod.rs
+++ b/rayon-core/src/sleep/mod.rs
@@ -70,6 +70,7 @@ impl Sleep {
             thread::yield_now();
             yields + 1
         } else if yields == ROUNDS_UNTIL_SLEEPY {
+            thread::yield_now();
             if self.get_sleepy(worker_index) {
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Thanks!

Issue #93, opened 2016-09-07
Use an user-provided "backend" (threadpool, event loop, ...)


gnzlbg:
Is it possible to tell rayon to use an user provided thread-pool (e.g. in my crate's thread function)? 

For example if I'd had a crate using tokio where I also want to use rayon, I would like to have a single thread-pool/event loop/task manager... that is used as the backend for both, instead of two competing ones.

nikomatsakis:
It is not currently possible, but one of my big To Do items for rayon is to factor out the backend into a distinct crate.

iduartgomez:
From looking at the source, I wonder if as an alternative (or in addition/ temporal workaround), it would be possible to have a finer control of the available number of threads for execution at a given time (ie. setting it up before running a parallel iterator chain) during runtime though a function call (that would in turn modify the Registry).

This would be really helpful in the context of load balancing in an application where you have several live threads competing over resources with different pools (other than rayon), for example. Is a bit hackish, but still something.

(The ideal solution would be IMO to have the same functionality provided by the global thread pool provided by instantiated ThreadPool types so we would have total control over how we split resources, a good steep would be to have something more akin https://github.com/frewsxcv/rust-threadpool incorporated to rayon).

cuviper:
> The ideal solution would be IMO to have the same functionality provided by the global thread pool provided by instantiated ThreadPool types so we would have total control over how we split resources

What do you mean by this?  Rayon's global thread pool is literally just a global `ThreadPool` instance.  You can `ThreadPool::install` your parallel iterator into whatever pool you like.

iduartgomez:
@cuviper Didn't see how to do it in the documentation (maybe is just that, a lack of documentation/example?), but for example:
```
hash_map
            .par_iter()
            .for_each(|(k, v)| // do stuff ... );
```
works but when if I use the `install` method I can't (`install` closure signature is FnOnce() -> R + Send so does not expect arguments), that's what i was thinking when saying 'have the same functionality'.

Probably missing something obvious.

cuviper:
The `install` closure has no arguments, but you can capture anything you like.  For instance:

```rust
let hash_map = get_map();
my_pool.install(|| {
    // this implicitly captures a reference to `hash_map`
    hash_map.par_iter().for_each(|(k, v)| {
        // do stuff ...
    })
});
```

nikomatsakis:
[...]

PR #353, opened 2017-05-31
introduce custom schedulers for parallel iterators

nikomatsakis:
This PR introduces the notion of "custom schedulers". The idea is that you can invoke, for any parallel iterator `iter`, `iter.in_scheduler(S)`, and it will run with the **scheduler** `S`. The scheduler basically takes the place of the (currently hard-coded) "bridge" functions -- it connects a parallel producer and a parallel consumer.

We introduce a `Scheduler` associated type into the `ParallelIterator` trait. This is used to ensure that the user cannot invoke `in_scheduler()` twice on any given parallel iterator. For implicity, we also forbid customized schedulers from the "inputs" where mutiple parallel iterators are "joined" (like `zip()` or `chain()`). Hence `a.in_scheduler().zip(b)` is illegal, but `a.zip(b).in_scheduler()` is fine. We could alter these rules to permit both (but we want to avoid `a.in_scheduler().zip(b.in_scheduler())`).

In this PR, Rayon exposes three schedulers: 

- `DefaultScheduler` -- what you get if you don't say anything; also encodes the fact that `in_scheduler` has never been called. Currently, this delegates to `AdaptiveScheduler`.
- `AdaptiveScheduler` -- the default strategy of "split to depth N, and more if stolen"
- `FixedLenScheduler` -- like above, but where we adjust N to try and respect the user's min/max requests

Originally, the idea here was to allow you to swap out the Rayon runtime completely. In fact, you can still see this in the `Runtime` trait introduced by the first commit, though I've removed it by the end -- basically because I don't feel we're ready to commit to this trait yet (and in principle no flexibility is lost). I imagine we'll probably add this back at some point in some form.

Some unknowns:

- [ ] Module structure. I feel like we should move most of these things to `iter::scheduler`.
- [ ] What should be done with the `Scheduler` associated type in `ParallelIterator`?
    - It's kind of weird right now. It *sounds* like it specifies the scheduler, but really it's always one of two values: `DefaultScheduler` or `CustomScheduler`. This is just used to prevent multiple `in_scheduler()` calls. I think I'd rather just give a semantics to multiple calls in some useful way. This interacts with the next question.
- [ ] Can we support "layering" schedulers?
    - Right now, we prohibit multiple calls to `in_scheduler`, but I was thinking that it'd also be possible to permit "combining" schedulers (e.g., `with_min` and `with_max`). I have to play around with this a bit more.
- [ ] What to do with `chain()` and other ad-hoc calls to `join()`?
    - @cuviper recently modified the `drive()` impl of `chain()` so that it will drive in parallel, using `join()`. This kind of goes against the grain of this PR, so we probably want to either extend the scheduler interface with a `join` method *or* reframe that as a parallel producer consumer, or something like that.

cc #93 -- this permits a user-provided "backend", though I think ultimately more flexibility might be desired. (This only works for parallel iterators, for one thing.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ docopt = "0.7"
 futures = "0.1.7"
 rand = "0.3"
 rustc-serialize = "0.3"
+thread-id = "3.1.0"
 
 [features]
 # Unstable APIs that have not yet
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -299,13 +299,6 @@ pub fn initialize(config: Configuration) -> Result<(), Box<Error>> {
     Ok(())
 }
 
-/// This is a debugging API not really intended for end users. It will
-/// dump some performance statistics out using `println`.
-#[cfg(feature = "unstable")]
-pub fn dump_stats() {
-    dump_stats!();
-}
-
 impl fmt::Debug for Configuration {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let Configuration { ref num_threads, ref get_thread_name, ref panic_handler, ref stack_size,
--- a/rayon-core/src/log.rs
+++ b/rayon-core/src/log.rs
@@ -9,7 +9,6 @@
 //! variable, which is still supported for backwards compatibility.
 
 use std::env;
-use std::sync::atomic::{AtomicUsize, ATOMIC_USIZE_INIT};
 
 #[derive(Debug)]
 pub enum Event {
@@ -57,29 +56,3 @@ macro_rules! log {
     }
 }
 
-pub static STOLEN_JOB: AtomicUsize = ATOMIC_USIZE_INIT;
-
-macro_rules! stat_stolen {
-    () => {
-        ::log::STOLEN_JOB.fetch_add(1, ::std::sync::atomic::Ordering::SeqCst);
-    }
-}
-
-pub static POPPED_JOB: AtomicUsize = ATOMIC_USIZE_INIT;
-
-macro_rules! stat_popped {
-    () => {
-        ::log::POPPED_JOB.fetch_add(1, ::std::sync::atomic::Ordering::SeqCst);
-    }
-}
-
-macro_rules! dump_stats {
-    () => {
-        {
-            let stolen = ::log::STOLEN_JOB.load(::std::sync::atomic::Ordering::SeqCst);
-            println!("Jobs stolen: {:?}", stolen);
-            let popped = ::log::POPPED_JOB.load(::std::sync::atomic::Ordering::SeqCst);
-            println!("Jobs popped: {:?}", popped);
-        }
-    }
-}
--- a/src/collections/binary_heap.rs
+++ b/src/collections/binary_heap.rs
@@ -12,6 +12,7 @@ use vec;
 impl<T: Ord + Send> IntoParallelIterator for BinaryHeap<T> {
     type Item = T;
     type Iter = IntoIter<T>;
+    type Scheduler = DefaultScheduler;
 
     fn into_par_iter(self) -> Self::Iter {
         IntoIter { inner: Vec::from(self).into_par_iter() }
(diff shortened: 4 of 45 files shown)


nikomatsakis:
OK, I see I was not running all the tests and some broke. Bah. Well, I've got to run now, but I'll get them working.

nikomatsakis:
Rebased and fixed the broken tests. (Well, let's see what travis has to say.)

xiaoniu-578fa6bff964d005:
Yet another custom scheduler ([code](https://github.com/xiaoniu-578fa6bff964d005/rayon/tree/custom_scheduler) and [diff](https://github.com/rayon-rs/rayon/compare/master...xiaoniu-578fa6bff964d005:eeb63cea4464505b95a30aafa1933f7a5f766660)). The design is not finalized and we wish to get feedback early. 

We tried another design which diverges in following ways:
- `Scheduler` is not an associated type in the ParallelIterator trait, but only belongs to `WithScheduler`, which is created by `iter.with_scheduler(S)`.
- `WithScheduler` forces the `bridge` of producer and consumer to happen in this part. That require the iterator on the left to work in producer mode and iterator on the right to work in consumer mode.
- Distinguish between `Scheduler` and `UnindexedScheduler`, since their behaviour are so different.

Some consequences of the design:
- Non breaking change (at least for indexed parallel iterator). No need to change the implementation of other existing tools like `Map`, `Zip` and `MinLen`. 
- User-provided "backend" is supported. 
- Scheduler can't be applied before `zip`. So `a.in_scheduler().zip(b)` and `a.zip(b.in_scheduler())` is illegal. That is because after `in_scheduler()`, iter can only work in consumer mode, but `Zip` works in producer mode, and iter on it's left can only work in producer mode. 
- Scheduler can't be applied twice, so there is no "layering" schedulers. The reason is similar to above.

Some minor designs that could change:
- Use empty link [hack](https://github.com/xiaoniu-578fa6bff964d005/rayon/blob/custom_scheduler/src/iter/scheduler/mod.rs#L55) to forbid using `iter.with_scheduler(S)` further in producer mode. When users do that, the program simply won't compile and output an ugly link error. 
Some alternatives:
 1. Use `panic!()`. Program will compile but panic when bridge happens.
 2. Use more rigours type system to seperate `with_producer()` and `drive()` into two different trait. So we can tell whether a ParallelIterator support working in producer mode or consumer mode. This would be a breaking change.

Some obstacle we are facing:
- Current `ParallelIterator` doesen't have `with_unindexed_producer()` method, so we can't explicitly get the producer. We believe breaking change is needed for support of unindexed scheduler.
- Benchmark is hard...
We find that merely using scheduler in a program will affect the performance of other part of the program where just default bridge function is used. We have no idea where this gloabl effect comes from. 

Features we wish to implement in the future:
-  UnindexedScheduler
We believe breaking change is needed to obtain the producer from ParallelIterator.
- Thread Affinity
We wish to allow users to assign which core they want to use to for a task and a slice of ParallelIterator.

cuviper:
@xiaoniu-578fa6bff964d005 
> * Benchmark is hard...
>   We find that merely using scheduler in a program will affect the performance of other part of the program where just default bridge function is used. We have no idea where this gloabl effect comes from.

A lot of performance depends on sufficient inlining -- it could just be that your additional code changed some thresholds in the module that caused LLVM to make different inlining decisions.

Apart from raw performance, you could also try to evaluate your changes at a higher level, like @wagnerf42 has been doing with his [rayon-adaptive](http://datamove.imag.fr/frederic.wagner/category/rayon-adaptive.html). He has logging infrastructure that can show exactly how the jobs were divided, so differences can be compared and justified.
(Those API changes are more invasive than what you seem to be suggesting.)

cuviper:
@xiaoniu-578fa6bff964d005 
> * Current `ParallelIterator` doesen't have `with_unindexed_producer()` method, so we can't explicitly get the producer. We believe breaking change is needed for support of unindexed scheduler.

Perhaps a breaking change can be avoided by adding a new trait for scheduling purposes.

Anyway, I'd suggest opening your own issue or pull request to discuss your design further.

nikomatsakis:
I'm going to close this PR as being obviously stalled and outdated.

PR #354, opened 2017-06-01
Introduce `join_context` which tells whether jobs were stolen

cuviper:
This removes the need for tracking thread-id or hacky TLS variables in
the adaptive splitter.  Instead, with `join_context` the parallel
iterator internals are directly informed whether a job is executing on
its original thread or stolen elsewhere.

Note that this is insta-stable since we want to use it ourselves.  We
could hide docs and/or mark it deprecated to approach stability a little
differently than the usual "unstable" feature.

### DIFF
--- a/rayon-core/src/join/mod.rs
+++ b/rayon-core/src/join/mod.rs
@@ -54,7 +54,21 @@ pub fn join<A, B, RA, RB>(oper_a: A, oper_b: B) -> (RA, RB)
           RA: Send,
           RB: Send
 {
-    registry::in_worker(|worker_thread| unsafe {
+    join_notify(|_| oper_a(), |_| oper_b())
+}
+
+/// The `join_notify` function is identical to `join`, except the closures
+/// have a parameter that indicates whether they're executing on a different
+/// thread than where `join_notify` was called.  This will occur if the second
+/// job is stolen by a different thread, or if `join_notify` was called from
+/// outside the thread pool to begin with.
+pub fn join_notify<A, B, RA, RB>(oper_a: A, oper_b: B) -> (RA, RB)
+    where A: FnOnce(bool) -> RA + Send,
+          B: FnOnce(bool) -> RB + Send,
+          RA: Send,
+          RB: Send
+{
+    registry::in_worker(|worker_thread, injected| unsafe {
         log!(Join { worker: worker_thread.index() });
 
         // Create virtual wrapper for task b; this all has to be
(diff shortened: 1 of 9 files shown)


leodasvacas:
Should the name be changed considering that `unpark` got renamed to `notify` in futures?

cuviper:
I think the places you'd use this are distinct enough from futures, but I'm open to other names.

nikomatsakis:
I wonder if we should design this API in a more open-ended way. That is, right now it supplies a simple `bool`. But maybe it should supply a struct with methods, so that we can add additional information in the future.

nikomatsakis:
As an example, I think it'd potentially be useful to tell closure B whether it was *stolen* or not (irrespective of whether we wound up jumping into the Rayon thread-pool). Right now, closure B can't tell the difference between the two.

nikomatsakis:
Apart from that nit, though, I think this looks pretty nice, and I'm inclined to say I like it better than the current hack of testing which thread you are on. It feels cleaner to me to be told, particularly since it doesn't require a lot of work on our part to track that information.

cuviper:
Rebased and renamed to `join_context` with an opaque `FnContext` struct, still just a single bool for now.  I didn't call it just `JoinContext` because I thought it might be a useful addition to stuff like `spawn` too.  Alternate name suggestions are welcome.

join_context-points

join_context-points-medians

I think that looks ok?  Some gains, some losses -- seem like noise.

rayon-core/src/join/mod.rs
```diff
@@ -43,18 +45,35 @@ pub fn join<A, B, RA, RB>(oper_a: A, oper_b: B) -> (RA, RB)
...
+
+/// The `join_context` function is identical to `join`, except the closures
+/// have a parameter that provides context for the way the closure has been
+/// called, especially indicating whether they're executing on a different
+/// thread than where `join_context` was called.  This will occur if the second
+/// job is stolen by a different thread, or if `join_context` was called from
+/// outside the thread pool to begin with.
+pub fn join_context<A, B, RA, RB>(oper_a: A, oper_b: B) -> (RA, RB)
+    where A: FnOnce(FnContext) -> RA + Send,
+          B: FnOnce(FnContext) -> RB + Send,
```
cuviper: Then its `Default` seems weird to me, which is used by the bridge functions in their starting state.  Should we allow arbitrary `FnContext<'a>` in that case, or just `FnContext<'static>`?

Perhaps we should also force this `!Send` and `!Sync` if we think it might reference local state.

nikomatsakis: Hmm I think I didn't notice that it supports `Default`. Maybe it's overkill to worry about this stuff anyway.

cuviper:
Rebased.

rayon-core/src/lib.rs
```diff
@@ -359,3 +360,41 @@ impl fmt::Debug for Configuration {
...
+/// Provides the calling context to a closure called by `join_context`.
+#[derive(Debug)]
+pub struct FnContext {
+    migrated: bool,
+
+    /// disable `Send` and `Sync`, just for a little future-proofing.
+    _marker: PhantomData<*mut ()>,
+}
+
+impl Default for FnContext {
```
nikomatsakis: Do we really want this to support `default`? That means anybody can make one...?

cuviper: I thought I needed a way for the iterator bridge functions to create one -- which is as good as "anybody" as far as crate privacy goes -- but I just modified those to extract the `migrated` bool to pass around instead.

PR #357, opened 2017-06-05
Bump futures to 0.1.16, move to `Notify` API

### LINKED ISSUE #356: Figure out upgrade to futures `Notify` API
See [this thread](https://users.rust-lang.org/t/how-to-pretend-something-lives-for-static/) for previous discussion. The new `Notify` API in futures 0.1.14, which replaces the old `Unpark`, no longer takes `Arc<Unpark>` but instead takes an generic `T: Into<UnparkHandle>`, which includes `Arc<T: Notify + Sized>`.

I naively thought this meant we could get rid of the `Unpark` trait object in [make_unpark](https://github.com/nikomatsakis/rayon/blob/194945535241f3ccafd5144aa9472d50692ae664/rayon-core/src/future/mod.rs#L223-L245), but trying to make the change I was blocked because I couldn't somehow cast a `ScopeFutureContents` to live for `'static`, and @cuviper notes that using a trait object may be necessary. The new futures API doesn't support `Arc<TraitObject>` so are going to have to hack around this somehow.

leodasvacas:
Fixes #356. See that issue for previous discussion. Ping @cuviper.

The `Notify` API is not fond of `Arc<Notify>` as the old API was with `Arc<Unpark>`. Instead it prefers to take a generic `T: Into<UnsafeNotify>`. Because of the `'static` restriction in the implementation provided by futures-rs we need to roll our own implementation of `UnsafeNotify` and jump through extra hoops to hide the lifetimes in `ScopeFuture`.

The implementation is easier to follow if you consider that it is analogous to [the implementation for `Arc` in futures-rs](https://github.com/alexcrichton/futures-rs/blob/master/src/task_impl/std/mod.rs#L610).
 
As a bright side, the old `make_unpark` is gone along with the `unpark` field, we now simply reuse the `this` self reference.

There are probably many places where the unpark terminology should be switched to notify, I should also update the great README.md.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,5 @@ default-features = false
 [dev-dependencies]
 compiletest_rs = "0.2.1"
 docopt = "0.7"
-futures = "0.1.14"
 rand = "0.3"
 rustc-serialize = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,6 @@ default-features = false
 [dev-dependencies]
 compiletest_rs = "0.2.1"
 docopt = "0.7"
-futures = "0.1.7"
+futures = "0.1.14"
 rand = "0.3"
 rustc-serialize = "0.3"
--- a/rayon-futures/src/lib.rs
+++ b/rayon-futures/src/lib.rs
@@ -9,19 +9,21 @@ extern crate futures;
 extern crate rayon_core;
 
 use futures::{Async, Future, Poll};
-use futures::executor;
 use futures::future::CatchUnwind;
-use futures::task::{self, Spawn, Task, Unpark};
+use futures::task::{self, Spawn, Task};
 use rayon_core::internal::task::{Task as RayonTask, ScopeHandle, ToScopeHandle};
 use rayon_core::internal::worker;
 use std::any::Any;
 use std::fmt;
 use std::panic::{self, AssertUnwindSafe};
+use futures::executor::{self, Notify, NotifyHandle, UnsafeNotify};
+use std::marker::PhantomData;
 use std::mem;
 use std::sync::Arc;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::*;
 use std::sync::Mutex;
+use std::ptr;
 
 const STATE_PARKED: usize = 0;
 const STATE_UNPARKED: usize = 1;
@@ -139,11 +141,10 @@ struct ScopeFutureContents<'scope, F, S>
     where F: Future + Send + 'scope, S: ScopeHandle<'scope>,
 {
     spawn: Option<Spawn<CU<F>>>,
-    unpark: Option<Arc<Unpark>>,
 
     // Pointer to ourselves. We `None` this out when we are finished
     // executing, but it's convenient to keep around normally.
-    this: Option<Arc<ScopeFuture<'scope, F, S>>>,
+    this: Option<ArcScopeFuture<'scope, F, S>>,
 
     // the counter in the scope; since the scope doesn't terminate until
     // counter reaches zero, and we hold a ref in this counter, we are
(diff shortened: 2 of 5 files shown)


lilianmoraru:
On AppVeyor seems like a Rust MinGW build issue(nightly regression?): https://ci.appveyor.com/project/lilianmoraru/rayon/build/1.0.4

In that build, this is being tested: https://github.com/nikomatsakis/rayon/compare/master...lilianmoraru:master

Pinging @nikomatsakis to raise awareness of a probable nightly regression.

cuviper:
Hmm, similar gcc issues were reported in the users forum:
https://users.rust-lang.org/t/rustup-beta-nightly-linker-error-on-windows-10/11236

lilianmoraru:
The issue is fixed on the latest nightly(not sure about beta but it is not relevant to this pull request since it already passed that test): https://ci.appveyor.com/project/lilianmoraru/rayon/build/1.0.5

nikomatsakis:
I'd rather make these changes atop of https://github.com/nikomatsakis/rayon/pull/344

nikomatsakis:
I'll have to review this `Notify` API. This feels like a lot of boilerplate, but maybe it's necessary.

leodasvacas:
Sure I can rebase this on #344

cuviper:
We now have separated rayon-futures -- please do rebase on that!

leodasvacas:
@cuviper Thanks for the heads up, rebased!

Cargo.toml
```diff
@@ -23,6 +23,6 @@ default-features = false
 [dev-dependencies]
 compiletest_rs = "0.2.1"
 docopt = "0.7"
-futures = "0.1.7"
+futures = "0.1.14"
```
cuviper: I missed this dependency in the shuffle, but I think it can just go away, as rayon-futures now tests itself.

nikomatsakis:
OK so this *looks* pretty reasonable! I'll be honest, I'm having a bit of a hard time wrapping my head around it, in part because I'm just a bit tired right now. =) @leodasvacas did you want to update the README.md etc?

leodasvacas:
@nikomatsakis Nice, I was just looking for an OK before updating the README, will do.

leodasvacas:
@cuviper Fixed it thanks!

rayon-futures/src/lib.rs
```diff
@@ -156,6 +157,142 @@ struct ScopeFutureContents<'scope, F, S>
...
+impl<'scope, F, S> Notify for ArcScopeFuture<'scope, F, S>
+where
+    F: Future + Send + 'scope,
+    S: ScopeHandle<'scope>,
+{
+    fn notify(&self, id: usize) {
+        self.0.notify(id)
+    }
+
+    fn clone_id(&self, id: usize) -> usize {
```
tmccombs: I don't think you really need to implement clone_id and drop_id, since you don't ever use a meaningful id that needs to be cloned or dropped specially.

rayon-futures/src/lib.rs
```diff
@@ -156,6 +157,142 @@ struct ScopeFutureContents<'scope, F, S>
...
+            let me: *const ScopeFutureWrapped<'scope, F, S> = self;
+            ArcScopeFuture::notify(
+                &*(&me as *const *const ScopeFutureWrapped<'scope, F, S> as
+                       *const ArcScopeFuture<'scope, F, S>),
+                id,
+            )
+        }
+    }
+
+    fn clone_id(&self, id: usize) -> usize {
```
tmccombs: Again I don't think you need to implement clone_id and drop_id here.

rayon-futures/src/lib.rs
```diff
@@ -156,6 +157,142 @@ struct ScopeFutureContents<'scope, F, S>
...
+}
+
+// Newtype so we can implement Into<UnsafeNotify> even though the contents are not 'static.
+#[derive(Debug)]
+struct ArcScopeFuture<'scope, F, S>(Arc<ScopeFuture<'scope, F, S>>)
+where
+    F: Future + Send + 'scope,
+    S: ScopeHandle<'scope>;
+
+impl<'scope, F, S> Clone for ArcScopeFuture<'scope, F, S>
```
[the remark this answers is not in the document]
leodasvacas: Nope, derive isn't smart enough I think it derives only when `F: Clone` and `S: Clone`. See rust-lang/rust#26925.

leodasvacas:
@tmccombs Thank you very much for the review! I've clarified the README. Regarding the `clone_id` and `drop_id` impls for `ArcScopeFuture` and `ScopeFutureWrapped`, while they will they might be useless for us and using the default would happen to be equivalent to the `Notify` impl for `ScopeFuture`, I believe the correct thing is to make them explicitly call that impl.

nikomatsakis:
bors r+

I've got to dig more into the notify APIs and stuff myself, but this looks really good. We can continue iterating here regardless. Thanks @leodasvacas very much and sorry about the delays.

nikomatsakis:
PS thanks to @tmccombs for the review as well. =) It's good to have more eyes on this.

cuviper:
bors retry

cuviper:
bors r=nikomatsakis

cuviper:
Looks like a corrupted rustc install on OSX stable.

Third time's the charm, or I'll merge it manually:

@bors r=nikomatsakis

notriddle:
Drop the `@`-sign.

cuviper:
Bah, thanks.

bors r=nikomatsakis

PR #360, opened 2017-06-10
mock up `breadth_first` and `spawn_tail`

nikomatsakis:
These introduce two new core APIs that were suggested by @bholley for use in Servo (Stylo, in particular). They are insta-stable because I would want to make a public release so that Servo can make use of them.

The APIs are:

- Introduce a new "breadth-first" scheduling mode that can be enabled per thread-pool. In breadth-first mode, both the current thread *and* stealer threads draw from the bottom of the deque. Depending on your workload, this may be desirable (in the case of stylo, it leads to better cache reuse). Users can opt into breadth-first mode by adjusting the configuration.
- Introduce a new `spawn_tail()` API that is intended to be used for the *last* task that will be spawned. It is intended as a slight optimization on calling `spawn` and then immediately returning from the scope, particularly in breadth-first mode. Specifically, it will invoke the closure directly if that closure would have been the next task that the local thread executes anyway.

These changes introduce an `if` into one of the main hot-paths of Rayon, so I was a bit nervous about performance. However, keep in mind that the `if` is 100% predictable in normal usage (i.e., if you don't mix breadth- and depth-first modes). Both @bholley and I did measurements that suggest basically zero impact on depth-first code. Also, using the newer modes leads to a 20% improvement in performance for Stylo. [The stylo measurements are documented here.](https://bugzilla.mozilla.org/show_bug.cgi?id=1365692#c20) My own measurements, gathered on my 8x2-core linux desktop, are visible here:

- [All measurements.](https://people-mozilla.org/~nmatsakis/rayon-bfs-points.svg)
- [Medians, normalized to 100.](https://people-mozilla.org/~nmatsakis/rayon-bfs-points-medians.svg)

In the median view, you can see *some* effect, but it is only a few percent, and I am not that confident in my measurement setup. =) Because of my inability to convince GNUplot to use colors etc in median mode (have to tinker with that more), it's a bit hard to tell which were affected; but if you hover the cursor over, you can see that `join_recursively`, `increment_all_max`, and `increment_all` are the ones that took more time, whereas `fibonacci` and `increment_all_min` took less time. All of these are micro-benchmarks.

I was considering putting these behind a cargo feature (perhaps one enabled by default), so that users could "opt-out" from having the `if`. This seems like a plausible path, but the measurements suggest to me that it's not necessary, and it seems better to avoid having "options" if we can (so as to reduce the need to measure an exponentially increasing number of combinations).

### DIFF
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -336,13 +336,6 @@ pub fn initialize(config: Configuration) -> Result<(), Box<Error>> {
     Ok(())
 }
 
-/// This is a debugging API not really intended for end users. It will
-/// dump some performance statistics out using `println`.
-#[cfg(feature = "unstable")]
-pub fn dump_stats() {
-    dump_stats!();
-}
-
 impl fmt::Debug for Configuration {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let Configuration { ref num_threads, ref get_thread_name,
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -116,6 +116,11 @@ pub struct Configuration {
 
     /// Closure invoked on worker thread exit.
     exit_handler: Option<Box<ExitHandler>>,
+
+    /// If false, worker threads will execute spawned jobs in a
+    /// "depth-first" fashion. If true, they will do a "breadth-first"
+    /// fashion. Depth-first is the default.
+    breadth_first: bool,
 }
 
 /// The type for a panic handling closure. Note that this same closure
@@ -239,6 +244,26 @@ impl Configuration {
         self
     }
 
+    /// Configure worker threads to execute spawned jobs in a
+    /// "breadth-first" fashion.  Typically, when a worker thread is
+    /// idle or blocked, it will attempt to execute the job from the
+    /// *top* of its local deque of work (i.e., the job most recently
+    /// spawned).  If this flag is set to true, however, workers will
+    /// prefer to execute in a *breadth-first* fashion -- that is,
+    /// they will search for jobs at the *bottom* of their local
+    /// deque.
+    ///
+    /// At present, workers *always* steal from the bottom of other
+    /// worker's deques, regardless of the setting of this flag.
+    pub fn breadth_first(mut self) -> Self {
+        self.breadth_first = true;
+        self
+    }
+
+    fn get_breadth_first(&self) -> bool {
+        self.breadth_first
+    }
+
     /// Takes the current thread start callback, leaving `None`.
     fn take_start_handler(&mut self) -> Option<Box<StartHandler>> {
         self.start_handler.take()
@@ -308,8 +333,10 @@ pub fn dump_stats() {
 
 impl fmt::Debug for Configuration {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let Configuration { ref num_threads, ref get_thread_name, ref panic_handler, ref stack_size,
-                            ref start_handler, ref exit_handler } = *self;
+        let Configuration { ref num_threads, ref get_thread_name,
+                            ref panic_handler, ref stack_size,
+                            ref start_handler, ref exit_handler,
+                            ref breadth_first } = *self;
 
         // Just print `Some("<closure>")` or `None` to the debug
         // output.
@@ -328,6 +355,7 @@ impl fmt::Debug for Configuration {
          .field("stack_size", &stack_size)
          .field("start_handler", &start_handler)
          .field("exit_handler", &exit_handler)
+         .field("breadth_first", &breadth_first)
          .finish()
     }
 }
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -247,17 +247,26 @@ impl Configuration {
         self
     }
 
-    /// Configure worker threads to execute spawned jobs in a
-    /// "breadth-first" fashion.  Typically, when a worker thread is
+    /// Suggest to worker threads that they execute spawned jobs in a
+    /// "breadth-first" fashion. Typically, when a worker thread is
     /// idle or blocked, it will attempt to execute the job from the
     /// *top* of its local deque of work (i.e., the job most recently
-    /// spawned).  If this flag is set to true, however, workers will
+    /// spawned). If this flag is set to true, however, workers will
     /// prefer to execute in a *breadth-first* fashion -- that is,
     /// they will search for jobs at the *bottom* of their local
-    /// deque.
+    /// deque. (At present, workers *always* steal from the bottom of
+    /// other worker's deques, regardless of the setting of this
+    /// flag.)
     ///
-    /// At present, workers *always* steal from the bottom of other
-    /// worker's deques, regardless of the setting of this flag.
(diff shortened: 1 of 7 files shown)


nikomatsakis:
I think now that I am re-considering the `spawn_tail()` API in particular. Perhaps it would be better to expose a method on `Scope` that simply tells you whether the local deque is empty or not. In that case, code using BFS mode that used to call `scope.spawn_tail(closure)` can instead do:

```rust
if scope.local_thread_has_pending_tasks() {
    scope.spawn(closure)
} else {
    closure()
}
```

There are two reasons for this:

- `spawn_tail` doesn't make sense in depth-first mode at all, since it is equivalent to just calling the closure.
- even in BF-mode, you probably would prefer to use a loop or something to avoid making your stack grow deeply.

For example:

```rust
let mut parent = some_node();
loop {
    let work_items = parent.children();

    // avoid spawning a task if (a) no other work is pending and (b) only one thing
    if work_items.len() == 1 && !scope.local_thread_has_pending_tasks() {
        parent = work_items[0];
        continue;
    }

    for child in work_items { scope.spawn(move || process(child)); }
}
```

@bholley -- thoughts?

rayon-core/src/registry.rs
```diff
@@ -403,11 +412,27 @@ impl WorkerThread {
...
+    /// Attempts to obtain a "local" job -- typically this means
+    /// popping from the top of the stack, though if we are configured
+    /// for breadth-first execution, it would mean dequeuing from the
+    /// bottom.
+    #[inline]
+    pub unsafe fn take_local_job(&self) -> Option<JobRef> {
+        if !self.breadth_first {
+            self.worker.pop()
+        } else {
+            self.stealer.steal()
```
stjepang: I've just published a new version of Coco that introduces `Worker::steal` method. Now you can call `self.worker.steal()` here. This change will give some performance wins in breadth-first mode. I'd be curious to know how much :)

nikomatsakis: Ah, nice!

nikomatsakis:
@stjepang I updated to use the new APIs. @bholley will have to re-run the measurements, though since we actually don't have any benchmarks employing BF mode (I've been meaning to make a pared down version of stylo for that purpose, but haven't gotten around to it.)

nikomatsakis:
Regarding the "local thread has pending tasks" idea, I guess that the best thing would be to expose it as a free function, rather like `current_thread_index` (well, that's presently only a method on `ThreadPool`, but I think we should rectify that). And it would probably be called `pending_local_tasks()` or something, with a comment explaining that this is an approximation that indicates whether the current thread has published unstolen tasks. Because of the inherent raciness, a return of `false` is meaningful here, but `true` maybe not.

(I'm a bit wary of 'locking in' on work-stealing here, but this seems like an API that we could continue to support even if we changed to something else, even if it meant always returning false or true. And anyway I can't really imagine changing to something else.)

bholley:
Yes, I would prefer the accessor over spawn_tail, which would be more ergonomic for our use-case.

nikomatsakis:
@bholley I pushed a change to `current_thread_has_pending_tasks()`, which is offered as a method on the `ThreadPool` (much like `current_thread_index()`).

I tend to think we should mirror all of these methods at the top-level as well. The main reason that I've been hesitating (particularly with respect to `current_thread_index()`) is that I was wary that if we ever changed our thread-pool strategy to dynamically change the number of threads, this would invalidate a lot of existing code. As long as the methods are restricted to the thread-pool, this seems harmless enough, since we could make it an "opt-in" bit of configuration (but make it the default for the **global** thread-pool).

But I am beginning to wonder if we will *ever* do this change -- or at least if we would ever do it without some form of opt-in. There doesn't seem to be much reason to do it, except to try to scale gracefully with blocking I/O tasks. And based on conversations I've had with others, that is best done via some opt-in APIs that wrap the blocking I/O, which would serve the purpose.

Therefore, I think I'd like to ensure that -- to the extent it makes sense -- the API surface of `ThreadPool` is mirrored with top-level functions that act on whatever the currently executing thread-pool is. That is, if you do `thread_pool.foo()`, then it is local to he given thread-pool, but `foo()` just works in your current context, whatever that is.

(@cuviper, what do you think of this reasoning?)

bholley:
SGTM. FWIW I think that making the thread pools dynamic would be a pretty surprising change, and we'd probably want it to be opt-in.

nikomatsakis:
@bholley 

> FWIW I think that making the thread pools dynamic would be a pretty surprising change, and we'd probably want it to be opt-in.

Yes, and also I'm not keen on the "default" configuration (i.e., `Configuration::new()`) and the configuration that the **default scheduler uses** diverging from one another.

nikomatsakis:
I've mirrored the `current_foo()` APIs at the global level now.

rayon-core/src/thread_pool/mod.rs
```diff
@@ -256,3 +293,83 @@ impl Drop for ThreadPool {
...
+}
+
+/// If called from a Rayon worker thread, indicates whether that
+/// thread's local deque still has pending tasks. Otherwise, returns
+/// `None`. For more information, see [the
+/// `ThreadPool::current_thread_has_pending_tasks()` method][m].
+///
+/// [m]: struct.ThreadPool.html#method.current_thread_has_pending_tasks
+#[inline]
+pub fn current_thread_has_pending_tasks(&self) -> Option<bool> {
```
cuviper: There's no `&self` in free functions...

rayon-core/src/lib.rs
```diff
@@ -63,6 +63,9 @@ mod unwind;
 mod util;
 
 pub use thread_pool::ThreadPool;
+pub use thread_pool::current_num_threads;
```
cuviper: There's already a `current_num_threads` below, and it's a little different because it reports on the global pool if you're not in a pool already.

rayon-core/src/thread_pool/mod.rs
```diff
@@ -256,3 +293,83 @@ impl Drop for ThreadPool {
...
+///
+/// [snt]: struct.Configuration.html#method.num_threads
+#[inline]
+pub fn current_num_threads() -> Option<usize> {
+    unsafe {
+        let curr = WorkerThread::current();
+        if curr.is_null() {
+            None
+        } else {
+            Some((*curr).registry.num_threads())
```
cuviper: The field `registry` is private.

rayon-core/src/registry.rs
```diff
@@ -403,11 +409,27 @@ impl WorkerThread {
...
-    /// stolen.
     #[inline]
-    pub unsafe fn pop(&self) -> Option<JobRef> {
-        self.worker.pop()
+    pub fn local_deque_is_empty(&self) -> bool {
+        self.worker.len() == 0
+    }
+
+    #[inline]
+    pub fn breadth_first(&self) -> bool {
```
cuviper: Is this method used anywhere?

nikomatsakis:
@cuviper um yeah obviously I didn't really build that. I pushed a new revision that addresses your various points. I removed the new `current_num_threads` in favor of the existing one. I had forgotten we made that available! I guess it's fine that it returns `usize` instead of `Option<usize>` -- that's probably what you want anyway, I imagine (i.e., you can measure the number of threads before you start your parallel operation, and this tells you how many threads you will have, unless you switch thread-pools in the meantime).

PR #362, opened 2017-06-12
Add `find_first` and `find_last` tests on a giant range

cuviper:
It's possible to generate giant iterators in Rayon, and these tests show
that they can still be short-circuited properly in find operations.

### DIFF
--- a/src/iter/find_first_last/test.rs
+++ b/src/iter/find_first_last/test.rs
@@ -100,3 +100,50 @@ fn find_last_folder_yields_last_match() {
     let f = f.consume(0_i32).consume(1_i32).consume(2_i32);
     assert_eq!(f.complete(), Some(2_i32));
 }
+
+
+/// Produce a parallel iterator for 0u128..10²⁷
+fn octillion() -> impl ParallelIterator<Item = u128> {
+    (0u32..1_000_000_000)
+        .into_par_iter()
+        .with_max_len(1_000)
+        .map(|i| i as u64 * 1_000_000_000)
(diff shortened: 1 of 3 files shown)


src/iter/find_first_last/test.rs
```diff
@@ -68,3 +68,82 @@ fn same_range_last_consumers_return_correct_answer() {
...
+}
+
+#[test]
+fn find_last_octillion() {
+    // FIXME: If we don't use at least two threads, then we end up walking
+    // through the entire iterator sequentially, without the benefit of any
+    // short-circuiting.  We probably don't want testing to wait that long. ;)
+    // It would be nice if `find_last` could prioritize the later splits,
+    // basically flipping the `join` args, without needing indexed `rev`.
+    // (or could we have an unindexed `rev`?)
```
cuviper: To some extent, this PR is really an RFC about this comment.

nikomatsakis: About which part exactly? :)

Like, is there a way to make `find_last` prioritize the later splits?

cuviper: Right.  I'd like this sort of thing to be usable even in a single-threaded "pool", and walking from start to end is a poor way to find the last match.  I doubt it would come up often with absurd iterators like this, but it should be a good optimization to search roughly backwards in general.

nikomatsakis: I see. That makes sense. I don't off hand know how best to do that though. =)

Issue #365, opened 2017-06-14
Idiomatic use of `map_with` for side-effects?


robinst:
Rayon 0.8.0 added the `map_with` and `fold_with` methods, which makes working with `Sender` simpler. Basically what had to be written like this before:

```rust
let pairs: Vec<_> = inputs.into_iter()
    .map(|input| (input, sender.clone()))
    .collect();
pairs.into_par_iter()
    .for_each(|(input, sender)| {
        let result = calculate(input);
        sender.send(result).unwrap();
    });
```

Can now be written like this:

```rust
inputs.into_par_iter()
    .map_with(sender, |sender, input| {
        let result = calculate(input);
        sender.send(result).unwrap();
    })
    .count();
```

The only problem is: If you forget the `count()`, everything will compile fine, except your program will not work at all. So, a couple of questions:

1. Is using `.count()` to force the work to happen considered idiomatic? If I change our code to that, I would have to add a comment explaining why the `.count()` is important, because it's a gotcha.
2. Have you considered adding a `for_each_with` for this use case, which does not have the above problem?
3. If that's not an option, is it possible to at least have a compiler warning for when the result of `map_with` is not used?

cuviper:
Yeah, you can run into the same gotcha even with normal iterators -- if you just call `map()` without doing anything else, then you've only created a wrapper on the iterator without actually running anything.  The Book has a section titled ["Iterators are Lazy"](https://doc.rust-lang.org/stable/book/second-edition/ch13-02-iterators.html#iterators-are-lazy) which discusses this, which just taught me that they *do* put a warning on those resulting types like so:

```rust
#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
```

So, yeah, we could add the same.

It's pretty common to workaround this with `count()`, or an empty `for_each(|_| {})` should work too.  You could also `collect()` into a `Result<Vec<()>, SendError>` instead of using your `unwrap()`, so you can deal with the error nicely instead of panicking.

I did consider `for_each_with` -- even had it in my initial implementation locally -- but I decided to try to limit the spread of `_with` suffixes to what seemed essential.  Just `fold_with` and `map_with` seemed to be the two unique primitives that couldn't be composed otherwise.  But maybe `for_each_with` is useful enough we should add it, given that the `Sender` motivation doesn't really need to propagate values.

cuviper:
0.8.1 now has the `#[must_use]` attributes as well as `for_each_with`.

[...]

PR #366, opened 2017-06-14
Add `#[must_use]` attributes to iterator adaptors

cuviper:
This follows the example of the standard library, even so lazy as to
steal their exact message about laziness.

cc #365


nikomatsakis:
👍

PR #368, opened 2017-06-14
adopt stricter semver policy around unstable features

### LINKED ISSUE #369: change from unstable cargo feature to RUSTFLAGS
So we encountered some problems that people are relying on things gated by "feature=unstable" (e.g., #364). This isn't really surprising. I was thinking that a better approach would be use a `cfg` flag that is not tied to a feature. Therefore, in order to build Rayon with experimental features, you would have to do:

    RUSTFLAGS='--cfg rayon_unstable' cargo build

This flag will propagate to the rayon create, which will then include unstable features. This has the same "infectious" property that Rust nightly builds do: you can't use an unstable feature in your crate without the top-level crate being forced to acknowledge it.

nikomatsakis:
This branch documents our semver policy and also makes it less convenient to use unstable features. They are no longer exposed as a cargo feature, but instead require `RUSTFLAGS=--cfg unstable`. 

Fixes #369.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ as:
 
 ```rust
 [dependencies]
-rayon = XXX # <-- insert latest version of Rayon from crates.io here
+rayon = 0.8.0
 ```
 
 and then add the following to to your `lib.rs`:
@@ -51,7 +51,7 @@ extern crate rayon;
 
 To use the Parallel Iterator APIs, a number of traits have to be in
 scope. The easiest way to bring those things into scope is to use the
-[Rayon prelude](https://docs.rs/rayon/0.8.0/rayon/prelude/index.html).
+[Rayon prelude](https://docs.rs/rayon/*/rayon/prelude/index.html).
 In each module where you would like to use the parallel iterator APIs,
 just add:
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -32,6 +32,33 @@ integers, please see the [notes on atomicity](#atomicity).
...
+
+and then add the following to to your `lib.rs`:
+
+```rust
+extern crate rayon;
+```
+
+To use the Parallel Iterator APIs, a number of traits have to be in
+scope. The easiest way to bring those things into scope is to use the
+[Rayon prelude](https://docs.rs/rayon/0.8.0/rayon/prelude/index.html).
```
cuviper: We can lazily use `*` instead of the explicit version number here.

README.md
```diff
@@ -32,6 +32,33 @@ integers, please see the [notes on atomicity](#atomicity).
...
 
+### Using Rayon
+
+[Rayon is available on crates.io](https://crates.io/crates/rayon). The
+recommended way to use it is to add a line into your Cargo.toml such
+as:
+
+```rust
+[dependencies]
+rayon = XXX # <-- insert latest version of Rayon from crates.io here
```
cuviper: Here I think we should fill in the current `"0.8"` semver, so users don't have to figure that out.  It's a relatively small thing for us to update when we do bump semver.

README.md
```diff
@@ -395,7 +422,35 @@ fn search(path: &Path, cost_so_far: usize, best_cost: &Arc<AtomicUsize>) {
...
+
+```
+RUSTFLAGS='--cfg rayon_unstable' cargo build
+```
+
+Note that this must not only be done for your crate, but for any crate
+that depends on your crate. This infectious nature is intentional, as
+it serves as a reminder that you are outside of the normal semver
+guarantees. **If you see unstable APIs that you would like to use,
+please open an issue about stabilizing them!**
```
cuviper: I think we should just make it standard policy to open tracking issues for all unstable features.

README.md
```diff
@@ -395,7 +422,35 @@ fn search(path: &Path, cost_so_far: usize, best_cost: &Arc<AtomicUsize>) {
...
+intended to be the main, user-facing crate, and hence all the
+documentation refers to `rayon`. This crate is still evolving and
+regularly goes through (minor) breaking changes. The `rayon-core`
+crate contains the global thread-pool and defines the core APIs: we no
+longer permit breaking changes in this crate (except to unstable
+features). The intention is that multiple semver-incompatible versions
+of the rayon crate can peacefully coexist; they will all share one
+global thread-pool through the `rayon-core` crate.
+
+## Unstable features## License
```
cuviper: Misplaced header?

README.md
cuviper: It occurs to me that the infectious nature is actually a double-edged sword.  Suppose one of my dependencies is happily using stable rayon-0.8 behind the scenes, but then I want to enable unstable rayon-0.9 / rayon-core-1.2 for my own purposes.  Broadly enabling `--cfg rayon_unstable` will break that stable 0.8-using crate too.

nikomatsakis: Hmm, interesting point. Still seems like an improvement relative to where we are now.

nikomatsakis:
OK, addressed your points.

cuviper:
(the CI queue is totally stalled, so I'm assuming you did sanity-check the build...)

PR #370, opened 2017-06-14
use <> instead of () for url blocks :)

amosbird:
well this just breaks my emacs since `()` is valid url characters.


jdanford:
The parentheses are actually necessary because that's the [Markdown syntax](https://guides.github.com/features/mastering-markdown/) for a link.

amosbird:
well... why not use asciidoc format in source code?

jdanford:
From [The Rust Programming Language](https://doc.rust-lang.org/1.4.0/book/documentation.html):
> Documentation comments are written in Markdown.

cuviper:
Alternatives have been discussed, but right now we can only use markdown.
https://internals.rust-lang.org/search?q=asciidoc

amosbird:
well then.

PR #371, opened 2017-06-14
Stabilize spawning an asynchronous task and convenience methods on ThreadPool

nikomatsakis:
Building on #368, this branch stabilizes the `spawn` API and a bunch of convenience methods in `ThreadPool`:

- `rayon_core::ThreadPool::join`
- `rayon_core::ThreadPool::scope`
- `rayon_core::ThreadPool::spawn`
- `rayon_core::spawn` -- runs async task in current (or global) thread-pool

The only one that enables something fundamentally new that wasn't available in a stable API is `rayon_core::spawn`. The `ThreadPool` APIs are convenience wrappers around other APIs.

The following APIs remain **unstable**:

- `ThreadPool::global` -- I don't know that we want to expose th `Arc`
  here, although it's no real commitment
- everything related to futures -- that is all planned to change

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,24 +15,24 @@ env:
 matrix:
   include:
   - rust: stable
-    env: FEATURES=unstable
+    env: RUSTFLAGS='--cfg rayon_unstable'
     os: linux
   - rust: stable
-    env: FEATURES=unstable
+    env: RUSTFLAGS='--cfg rayon_unstable'
     os: osx
   - rust: nightly
-    env: FEATURES=unstable
+    env: RUSTFLAGS='--cfg rayon_unstable'
     os: linux
   - rust: nightly
-    env: FEATURES=unstable
+    env: RUSTFLAGS='--cfg rayon_unstable'
     os: osx
 
 script:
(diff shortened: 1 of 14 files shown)


nikomatsakis:
> I've forgotten the reason why this is even Arc at all...

I'm not sure the reason is super good. I think it was because I found that some code I was using wanted to have an `Arc<ThreadPool>` so that it owned the thread-pool, and this way that thread-pool could be *either* locally created *or* the global one. I predicted that using `Arc` would be a common choice because it's handy to share the thread-pool widely.

But maybe just returning `&'static ThreadPool` is better, and people can use `Cow` or something to accommodate the "maybe shared" case (but `Cow` doesn't play nice with `Arc`...).

seanmonstar:
I haven't tested yet, but wouldn't a crate be able to automatically add this cfg in using a build.rs, without consumers knowing?

cuviper:
@seanmonstar I believe that can only enable a cfg for that crate itself, not its build of rayon.

nikomatsakis:
@seanmonstar I don't know, maybe. If so, I mean, what ya' gonna do...

PR #375, opened 2017-06-15
add LICENSE files to rayon-core

### LINKED ISSUE #374: LICENSE is not included in rayon-core
As usual, Apache-2.0 requires license to be presented along the source code. Even we have license in upstream, but crates uploaded on crates.io do not have them... So please add license files to rayon-core.

nikomatsakis:
Fixes #374


cuviper:
👍

PR #377, opened 2017-06-17
Fix dependencies block in readme

stjepang:
The version must be enclosed with double quotes, otherwise you get something like:

```
error: failed to parse manifest at `/home/stjepan/work/par-sort/Cargo.toml`

Caused by:
  could not parse input as TOML

Caused by:
  expected newline, found a period at line 10
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ as:
 
 ```rust
 [dependencies]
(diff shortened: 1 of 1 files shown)


cuviper:
Thanks!

PR #379, opened 2017-06-22
Add parallel sorts

### LINKED ISSUE #376: quicksort example stack overflows on large input
I was hoping I could be super lazy and copy paste the parallel quicksort, but alas, it seems that for very large inputs this can stack overflow, as it recurses.

On that note are there any plans on introducing an efficient parallel sort into the lib or is that outside the scope of rayon ? (I frequently want this, but I don't have the time to mess around with it and was hoping there was a solution already)

stjepang:
This PR introduces 6 new methods on mutable slices:

1. `par_sort`
2. `par_sort_by`
3. `par_sort_by_key`
4. `par_sort_unstable`
5. `par_sort_unstable_by`
6. `par_sort_unstable_by_key`

The first three methods perform stable sorting (it allocates a temporary buffer), and the last three perform unstable sorting (it is in-place). As a rule of thumb, unstable sorting is generally faster.

cc @cuviper @m4b
Fixes #376 

### Stable sort (mergesort)

The implementation is based on the one in libstd - it is a mixture of plain mergesort and timsort. An important characteristic of the algorithm is that it identifies pre-sorted intervals in advance and only then merges them.

The parallel variant divides the slice into multiple chunks and sorts them in parallel. After that it concatenates adjacent chunks if they're parts of greater pre-sorted runs. The remaining chunks are merged recursively in parallel. The final procedure that merges two sorted chunks is parallelized as well.

### Unstable sort (quicksort)

This is pretty much vanilla `sort_unstable` from libcore with `rayon::join` in place of recursive quicksort calls. The algorithm is based on pattern-defeating quicksort, which is today the state-of-the-art of general unstable sorting.

This implementation could be improved by parallelizing partitioning, but that would introduce a new bound on types - `T: Sync`. Instead of forcing the bound on users we could use specialization (when it gets into stable Rust) and choose parallel partitioning only for types that satisfy the bound. However, the implementation already performs quite well so I didn't bother with that for now.

### Tests

There are unit tests that test sorting on corner cases, random slices and partially sorted slices.

An important aspect of sort implementations is panic safety. The comparison function may panic at any time during the algorithm. If a panic occurs, we must ensure that all elements from the original slice are still present and have the most recent value (this is important if the elements have interior mutability - we don't want to hold stale versions of elements).

Panic safety is tested through a special run-pass test that panics at a bunch of stages during sort procedures and verifies that all elements are still present and up-to-date.

### Benchmarks

*Note*: The following numbers were obtained on my dual-core machine with hyperthreading.

Sorting 100 million random unsigned 64-bit integers:

| Implementation | Language | Parallel? | Stable? | Seconds | Code                                              |
|----------------|----------|-----------|---------|--------:|---------------------------------------------------|
| OpenJDK        | Java     | No        | No      | 11.08   | `Arrays.sort(v)`                                  |
| OpenJDK        | Java     | Yes       | No      | 5.51    | `Arrays.parallelSort(v)`                          |
| libstdc++      | C++      | No        | Yes     | 11.83   | `std::stable_sort(v.begin(), v.end())`            |
| libstdc++      | C++      | No        | No      | 9.98    | `std::sort(v.begin(), v.end())`                   |
| GCC / OpenMP   | C++      | Yes       | Yes     | 5.03    | `__gnu_parallel::stable_sort(v.begin(), v.end())` |
| GCC / OpenMP   | C++      | Yes       | No      | 4.28    | `__gnu_parallel::sort(v.begin(), v.end())`        |
| Intel TBB      | C++      | Yes       | No      | 4.37    | `tbb::parallel_sort(v.begin(), v.end())`          |
| libstd         | Rust     | No        | Yes     | 8.26    | `v.sort()`                                        |
| libcore        | Rust     | No        | No      | 4.55    | `v.sort_unstable()`                               |
| Rayon          | Rust     | Yes       | Yes     | 3.46    | `v.par_sort()`                                    |
| Rayon          | Rust     | Yes       | No      | 2.45    | `v.par_sort_unstable()`                           |

Benchmarks in Rayon:

```
test sort::par_sort_ascending                  ... bench:     57,595 ns/iter (+/- 33,612) = 6945 MB/s
test sort::par_sort_big                        ... bench: 14,346,411 ns/iter (+/- 5,629,095) = 446 MB/s
test sort::par_sort_descending                 ... bench:    100,633 ns/iter (+/- 73,757) = 3974 MB/s
test sort::par_sort_expensive                  ... bench: 62,525,848 ns/iter (+/- 18,243,649) = 6 MB/s
test sort::par_sort_mostly_ascending           ... bench:    445,101 ns/iter (+/- 387,678) = 898 MB/s
test sort::par_sort_mostly_descending          ... bench:    465,715 ns/iter (+/- 359,726) = 858 MB/s
test sort::par_sort_random                     ... bench:    911,327 ns/iter (+/- 679,526) = 438 MB/s
test sort::par_sort_strings                    ... bench:  5,342,137 ns/iter (+/- 3,964,614) = 149 MB/s
test sort::par_sort_unstable_ascending         ... bench:     49,189 ns/iter (+/- 20,092) = 8131 MB/s
test sort::par_sort_unstable_big               ... bench:  7,372,289 ns/iter (+/- 3,964,774) = 868 MB/s
test sort::par_sort_unstable_descending        ... bench:     73,053 ns/iter (+/- 5,045) = 5475 MB/s
test sort::par_sort_unstable_expensive         ... bench: 77,953,278 ns/iter (+/- 26,770,855) = 5 MB/s
test sort::par_sort_unstable_mostly_ascending  ... bench:    350,948 ns/iter (+/- 183,640) = 1139 MB/s
test sort::par_sort_unstable_mostly_descending ... bench:    351,790 ns/iter (+/- 229,890) = 1137 MB/s
test sort::par_sort_unstable_random            ... bench:    812,712 ns/iter (+/- 443,710) = 492 MB/s
test sort::par_sort_unstable_strings           ... bench:  5,740,113 ns/iter (+/- 5,193,645) = 139 MB/s
```

Note that sorting is much faster on pre-sorted (ascending/descending) and partially sorted (mostly ascending/descending) than random inputs.

### DIFF
--- a/rayon-demo/src/sort.rs
+++ b/rayon-demo/src/sort.rs
@@ -1,7 +1,9 @@
-use rand::{thread_rng, Rng};
+use rand::{Rng, SeedableRng, XorShiftRng};
 use rayon::prelude::*;
 use std::mem;
-use test::Bencher;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering::SeqCst;
+use test::{Bencher, black_box};
 
 fn gen_ascending(len: usize) -> Vec<u64> {
     (0..len as u64).collect()
@@ -12,12 +14,12 @@ fn gen_descending(len: usize) -> Vec<u64> {
 }
 
 fn gen_random(len: usize) -> Vec<u64> {
-    let mut rng = thread_rng();
+    let mut rng = XorShiftRng::from_seed([0, 1, 2, 3]);
     rng.gen_iter::<u64>().take(len).collect()
 }
 
 fn gen_mostly_ascending(len: usize) -> Vec<u64> {
-    let mut rng = thread_rng();
+    let mut rng = XorShiftRng::from_seed([0, 1, 2, 3]);
     let mut v = gen_ascending(len);
     for _ in (0usize..).take_while(|x| x * x <= len) {
         let x = rng.gen::<usize>() % len;
@@ -28,7 +30,7 @@ fn gen_mostly_ascending(len: usize) -> Vec<u64> {
 }
 
 fn gen_mostly_descending(len: usize) -> Vec<u64> {
-    let mut rng = thread_rng();
+    let mut rng = XorShiftRng::from_seed([0, 1, 2, 3]);
     let mut v = gen_descending(len);
     for _ in (0usize..).take_while(|x| x * x <= len) {
         let x = rng.gen::<usize>() % len;
@@ -39,7 +41,7 @@ fn gen_mostly_descending(len: usize) -> Vec<u64> {
 }
 
 fn gen_strings(len: usize) -> Vec<String> {
-    let mut rng = thread_rng();
+    let mut rng = XorShiftRng::from_seed([0, 1, 2, 3]);
     let mut v = vec![];
     for _ in 0..len {
         let n = rng.gen::<usize>() % 20 + 1;
@@ -49,7 +51,7 @@ fn gen_strings(len: usize) -> Vec<String> {
 }
 
 fn gen_big_random(len: usize) -> Vec<[u64; 16]> {
-    let mut rng = thread_rng();
+    let mut rng = XorShiftRng::from_seed([0, 1, 2, 3]);
     rng.gen_iter().map(|x| [x; 16]).take(len).collect()
 }
 
@@ -57,24 +59,61 @@ macro_rules! sort {
     ($f:ident, $name:ident, $gen:expr, $len:expr) => {
         #[bench]
         fn $name(b: &mut Bencher) {
-            b.iter(|| $gen($len).$f());
+            let v = $gen($len);
+            b.iter(|| v.clone().$f());
             b.bytes = $len * mem::size_of_val(&$gen(1)[0]) as u64;
         }
     }
 }
 
-sort!(par_sort, par_sort_large_ascending, gen_ascending, 50_000);
-sort!(par_sort, par_sort_large_descending, gen_descending, 50_000);
-sort!(par_sort, par_sort_large_mostly_ascending, gen_mostly_ascending, 50_000);
-sort!(par_sort, par_sort_large_mostly_descending, gen_mostly_descending, 50_000);
-sort!(par_sort, par_sort_large_random, gen_random, 50_000);
-sort!(par_sort, par_sort_large_big_random, gen_big_random, 50_000);
-sort!(par_sort, par_sort_large_strings, gen_strings, 50_000);
+macro_rules! sort_strings {
+    ($f:ident, $name:ident, $gen:expr, $len:expr) => {
+        #[bench]
+        fn $name(b: &mut Bencher) {
+            let v = $gen($len);
+            let v = v.iter().map(|s| &**s).collect::<Vec<&str>>();
+            b.iter(|| v.clone().$f());
+            b.bytes = $len * mem::size_of::<&str>() as u64;
+        }
+    }
+}
 
-sort!(par_sort_unstable, par_sort_unstable_large_ascending, gen_ascending, 50_000);
-sort!(par_sort_unstable, par_sort_unstable_large_descending, gen_descending, 50_000);
-sort!(par_sort_unstable, par_sort_unstable_large_mostly_ascending, gen_mostly_ascending, 50_000);
-sort!(par_sort_unstable, par_sort_unstable_large_mostly_descending, gen_mostly_descending, 50_000);
-sort!(par_sort_unstable, par_sort_unstable_large_random, gen_random, 50_000);
-sort!(par_sort_unstable, par_sort_unstable_large_big_random, gen_big_random, 50_000);
-sort!(par_sort_unstable, par_sort_unstable_large_strings, gen_strings, 50_000);
+macro_rules! sort_expensive {
+    ($f:ident, $name:ident, $gen:expr, $len:expr) => {
+        #[bench]
+        fn $name(b: &mut Bencher) {
+            let v = $gen($len);
+            b.iter(|| {
+                let count = AtomicUsize::new(0);
+                let mut v = v.clone();
+                v.$f(|a: &u64, b: &u64| {
+                    count.fetch_add(1, SeqCst);
+                    if count.load(SeqCst) % 1_000_000_000 == 0 {
+                        panic!("should not happen");
+                    }
+                    (*a as f64).cos().partial_cmp(&(*b as f64).cos()).unwrap()
+                });
+                black_box(count);
+            });
+            b.bytes = $len * mem::size_of_val(&$gen(1)[0]) as u64;
+        }
+    }
+}
+
+sort!(par_sort, par_sort_ascending, gen_ascending, 50_000);
+sort!(par_sort, par_sort_descending, gen_descending, 50_000);
+sort!(par_sort, par_sort_mostly_ascending, gen_mostly_ascending, 50_000);
+sort!(par_sort, par_sort_mostly_descending, gen_mostly_descending, 50_000);
+sort!(par_sort, par_sort_random, gen_random, 50_000);
+sort!(par_sort, par_sort_big, gen_big_random, 50_000);
+sort_strings!(par_sort, par_sort_strings, gen_strings, 50_000);
+sort_expensive!(par_sort_by, par_sort_expensive, gen_random, 50_000);
+
+sort!(par_sort_unstable, par_sort_unstable_ascending, gen_ascending, 50_000);
+sort!(par_sort_unstable, par_sort_unstable_descending, gen_descending, 50_000);
+sort!(par_sort_unstable, par_sort_unstable_mostly_ascending, gen_mostly_ascending, 50_000);
+sort!(par_sort_unstable, par_sort_unstable_mostly_descending, gen_mostly_descending, 50_000);
+sort!(par_sort_unstable, par_sort_unstable_random, gen_random, 50_000);
+sort!(par_sort_unstable, par_sort_unstable_big, gen_big_random, 50_000);
+sort_strings!(par_sort_unstable, par_sort_unstable_strings, gen_strings, 50_000);
+sort_expensive!(par_sort_unstable_by, par_sort_unstable_expensive, gen_random, 50_000);
--- a/rayon-demo/src/sort.rs
+++ b/rayon-demo/src/sort.rs
@@ -86,13 +86,18 @@ macro_rules! sort_expensive {
             b.iter(|| {
                 let count = AtomicUsize::new(0);
                 let mut v = v.clone();
+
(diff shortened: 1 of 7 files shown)


nikomatsakis:
So...this looks really cool! :)

stjepang:
Regarding code review: The PR is enormous (sorry about that!) and I don't expect anyone to review it completely. Since both sorts are based on previous implementation that were already reviewed in the Rust repo, I believe just taking a look at the parallelization procedures will suffice.

More concretely:

1. `mergesort.rs` - Function `mergesort` and everything above it were copied from `slice::sort`, but everything below it is new.
2. `quicksort.rs` - The only difference from `slice::sort_stable` is the call to `rayon_core::join`, that's all. This one is trivial.
3. `sort-panic-safe.rs` - This one was copied from the Rust repo as well and tweaked to support parallelism. I also added new checks for internally mutable `T`s (I'll backport those improvements to the Rust repo). Just skim over it.

src/slice/mergesort.rs
```diff
@@ -0,0 +1,749 @@
...
+
+    // Now attempt to concatenate adjacent chunks that were left intact.
+    let mut chunks = Vec::with_capacity(iter.len());
+
+    while let Some((a, mut b, res)) = iter.next() {
+        // If this chunk was not modified by the sort procedure...
+        if res != MergesortResult::Sorted {
+            while let Some(&(x, y, r)) = iter.peek() {
+                // If the following chunk is of the same type and can be concatenated...
+                if r == res && (r == MergesortResult::Descending) == is_less(&v[x], &v[x - 1]) {
```
cuviper: It's a minor optimization, but I think there's a lost case here for two descending chunks with *equal* neighbors, that could be concatenated.

- both ascending and `v[x-1] <= v[x]`
- both descending and `v[x-1] >= v[x]`

I don't think that can be written with a single `is_less` call.

src/slice/mergesort.rs
```diff
@@ -0,0 +1,749 @@
...
+/// The allocated temporary buffer is of the same length as is `v`.
+pub fn par_mergesort<T, F>(v: &mut [T], is_less: F)
+where
+    T: Send,
+    F: Fn(&T, &T) -> bool + Sync,
+{
+    // Slices of up to this length get sorted using insertion sort.
+    const MAX_INSERTION: usize = 20;
+    // The length of initial chunks.
+    const CHUNK_LENGTH: usize = 2000;
```
cuviper: magic

src/slice/quicksort.rs
```diff
@@ -0,0 +1,787 @@
...
+/// this function will immediately switch to heapsort.
+fn recurse<'a, T, F>(mut v: &'a mut [T], is_less: &F, mut pred: Option<&'a mut T>, mut limit: usize)
+where
+    T: Send,
+    F: Fn(&T, &T) -> bool + Sync,
+{
+    // Slices of up to this length get sorted using insertion sort.
+    const MAX_INSERTION: usize = 20;
+    // If both partitions are up to this length, we continue sequentially.
+    const MAX_SEQUENTIAL: usize = 2000;
```
cuviper: magic

src/slice/mod.rs
```diff
@@ -76,6 +84,139 @@ pub trait ParallelSliceMut<T: Send> {
             slice: self.as_parallel_slice_mut(),
         }
     }
+
+    /// Sorts the slice.
```
stjepang: I've extended these comments now.

src/slice/mergesort.rs
```diff
@@ -0,0 +1,749 @@
...
+///
+/// Even if `is_less` panics at any point during the merge process, this function will fully copy
+/// all elements from `left` and `right` into `dest` (not necessarily in sorted order).
+unsafe fn par_merge<T, F>(left: &mut [T], right: &mut [T], dest: *mut T, is_less: &F)
+where
+    T: Send,
+    F: Fn(&T, &T) -> bool + Sync,
+{
+    // Slices whose lengths sum up to this value are merged sequentially.
+    const MAX_SEQUENTIAL: usize = 5000;
```
stjepang: I've added a comment explaining how these magic numbers were chosen.

src/slice/mergesort.rs
stjepang: That is true, but descending (not strictly descending) runs are a bit tricky - you can't just reverse them because that would break stability guarantees of the algorithm. :) In any case, this problem is probably not worth solving...

stjepang:
@cuviper I've added another benchmark (the "expensive" one) that sorts using an expensive comparison function. The comments are improved as you suggested, benchmarks use `XorShiftRng` and vectors are cloned rather than generated from scratch every time.

Regarding heaps of unsafe code... yeah, this is one of those cases where there's no way to avoid unsafe without paying a considerable performance price. If that makes you feel any better, I've spent months working with those sort implementations and feel pretty confident about them. :) Sequential mergesort has been in libstd for quite some time now, and quicksort (aka pdqsort) is used by Servo.

rayon-demo/src/sort.rs
```diff
@@ -49,32 +51,69 @@ fn gen_strings(len: usize) -> Vec<String> {
...
+        #[bench]
+        fn $name(b: &mut Bencher) {
+            let v = $gen($len);
+            b.iter(|| {
+                let count = AtomicUsize::new(0);
+                let mut v = v.clone();
+                v.$f(|a: &u64, b: &u64| {
+                    count.fetch_add(1, SeqCst);
+                    if count.load(SeqCst) % 1_000_000_000 == 0 {
+                        panic!("should not happen");
```
cuviper: Is this making sure we don't hit O(n<sup>2</sup>) comparisons? (since 50,000<sup>2</sup> > 10<sup>9</sup>)  A comment would be nice.

The `fetch_add` followed by `load` is strange too, but doesn't matter much since this is just toy benchmark code...

src/slice/mod.rs
```diff
@@ -85,10 +85,14 @@ pub trait ParallelSliceMut<T: Send> {
         }
     }
 
-    /// Sorts the slice.
+    /// Sorts the slice in parallel.
     ///
     /// This sort is stable (i.e. does not reorder equal elements) and `O(n log n)` worst-case.
     ///
+    /// When applicable, unstable sorting is preferred because it is generally faster than stable
+    /// sorting and it doesn't allocate auxilliary memory.
```
cuviper: *auxiliary

rayon-demo/src/sort.rs
stjepang: Not really. :) This comparison function simply packs a lot of costly stuff:

1. Atomic operations.
2. A modulo operation.
3. A possible panic (they need landing pads).
4. A stack-referenced variable.
5. Trigonometric operations.

All that is there just to slow things down and make optimizer's job more difficult so we can see how the sort algorithms behave with very expensive comparison functions. 

I've added a comment explaining the intentions behind those several lines of code.

src/slice/mod.rs
stjepang: Fixed. (and there were a few other places with the same mistake :))

stjepang: I think GitHub is acting up... I pushed a new commit but it's not yet visible. Anyways, if you don't see the 6th commit in this PR, you can find it in the [original branch](https://github.com/stjepang/rayon/tree/par-sort). :/

nikomatsakis:
I looked things over. I am happy.

Sindisil:
*GREAT* work! However, there is an error in the benchmark table: Java's sorts **are** stable.

cuviper:
> (*) The sorts in Java are stable, but the implementation switches to dual-pivot quicksort when sorting integers, which is actually an unstable algorithm.

It might be interesting to have a trait in `std` on types where stable ordering is irrelevant (not observable), then we could specialize for that too.  `par_sort_unstable` is not _always_ faster, as the `_expensive` benchmarks show, but that might change if we were also specializing `T: Sync` for parallel partitioning.

rvolgers:
> It might be interesting to have a trait in std on types where stable ordering is irrelevant (not observable)

This may be trying to be too clever, but it seems like that would be okay for at least all Copy types?

stjepang:
> It might be interesting to have a trait in std on types where stable ordering is irrelevant (not observable), then we could specialize for that too.

It is tempting to do that, but I don't think it's worth it. It is true that unstable sorting is *generally* faster than stable sorting, but not strictly always (e.g. say, when the input consists of several pre-sorted sequences concatenated one after another). The same is true in C++ for `std::sort` and `std::stable_sort`.

Sometimes advanced users that care a lot about performance might have a perfectly valid reason to  use stable sort even if stability doesn't matter. By overriding stable sort with unstable sort in that case we'd simply take the choice away. :)

CryZe:
You could have `sort`, `stable_sort` and `unstable_sort`, with `sort` being specialized certain cases where unstable sort makes sense and if you want to specifically opt into the stable sort algorithm, you can do so by calling the `stable_sort` method instead. Similar to the overflow methods that allow you to opt into different behavior if needed.

carlomilanesi:
I think that sorting a slice of numbers is unrealistic. The benchmarks should sort by `id` the following structure:
`struct Record { descr: String, id: u64 }`
where `descr` contains the value of `id` converted to a decimal string. For example:
`Record { descr: format!("{}", 12345), id: 12345 }`
But if small-string optimization is used, a longer string should be used.

cuviper:
Rust's `String` doesn't do small-string optimization, but that wouldn't matter anyway if you're not using that in the comparison.  The cost of moving values basically just depends on the size of the struct itself, and any heap pointers will be unchanged.

carlomilanesi:
cuviper, I was thinking about the other languages in the comparison. Actually, string types have too many several possible implementations to use in a fair comparison. Therefore, I presume a better basic comparison of sorting algorithms would use the following structure:
```
struct Record {
    payload1: u64,
    payload2: u64,
    payload3: u64,
    key: u64,
}
```

cuviper:
OK, so you're talking about [stjepang/par-sort](https://github.com/stjepang/par-sort/), not Rayon's in-tree benchmarks, right?  We don't have any language comparisons in this repo -- it was only referenced in this PR summary.

carlomilanesi:
cuviper, yes, I was commenting on stjepang comment about par-sort, talking about "Sorting 100 million random 64-bit integers".

stjepang:
@carlomilanesi You're raising a fair point. These benchmarks are quite simplistic. They give us *some* idea of how fast our sorts are, but aren't extensive by any means and therefore cannot be taken for granted.

m4b:
Fwiw, I've seen absolutely incredible speed ups using this, from ~200-300 seconds to 4-8... for extraordinarily large dataset of 50-80GB datafile with approximately 200million number of records.

I'm not going back to linear sorting :D

PR #387, opened 2017-06-29
Add flatten iterator

### LINKED ISSUE #383: implement `flatten` from itertools
The `itertools` crate offers [a `flatten` method](https://docs.rs/itertools/0.6.0/itertools/trait.Itertools.html#method.flatten). From the docs, "This is more or less equivalent to .flat_map with an identity function."

Implementing this in Rayon should be fairly straightforward, since we support `flat_map`. You basically have to extend [the `ParallelIterator` trait](https://github.com/nikomatsakis/rayon/blob/29ab2edf82b68b7dcd6370ee296e741d487ba9bd/src/iter/mod.rs#L123) with a new adapter that [invokes `flat_map`](https://github.com/nikomatsakis/rayon/blob/29ab2edf82b68b7dcd6370ee296e741d487ba9bd/src/iter/mod.rs#L209-L216). Naturally we should include some documentation.

There is however one complication. Lacking `impl Trait`, we can't embed a closure type in the signature. Moreover, implementing the `Fn` trait directly is not possible on stable Rust. This will therefore require creating a custom adapter `Flatten` -- this could be a cut-and-paste job from [`flat_map`](https://github.com/nikomatsakis/rayon/blob/29ab2edf82b68b7dcd6370ee296e741d487ba9bd/src/iter/flat_map.rs#L10-L13), but that's unappealing. Another option would be to implement the `ParallelIterator` trait for `flatten` and have it invoke `flat_map` during the `drive` function. Something like this:

```rust
impl<I, F, PI> ParallelIterator for Flatten<I>
    where I: ParallelIterator<Item=PI>,
          PI: IntoParallelIterator,
{
    type Item = PI::Item;

    {
        self.flat_map(|x| x).drive_unindexed(consumer)
    }
}
```

For a general overview of how the parallel iterator traits work, see [the README](https://github.com/nikomatsakis/rayon/blob/29ab2edf82b68b7dcd6370ee296e741d487ba9bd/src/iter/README.md).

mgattozzi:
Closes #383

### DIFF
--- a/src/iter/flatten.rs
+++ b/src/iter/flatten.rs
@@ -1,0 +1,35 @@
+use super::internal::*;
+use super::*;
+
+/// `Flatten` turns each element to an iterator, then flattens these iterators
+/// together. This struct is created by the [`flatten()`] method on
+/// [`ParallelIterator`].
+///
+/// [`flatten()`]: trait.ParallelIterator.html#method.flatten
+/// [`ParallelIterator`]: trait.ParallelIterator.html
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct Flatten<I: ParallelIterator> {
+    base: I,
+}
+
+/// Create a new `Flatten` iterator.
+pub fn new<I>(base: I) -> Flatten<I>
+    where I: ParallelIterator
+{
+    Flatten {
+        base: base,
+    }
+}
+
+impl<I, PI> ParallelIterator for Flatten<I>
+    where I: ParallelIterator<Item = PI>,
+          PI: IntoParallelIterator + Send + Sync
+{
+    type Item = PI::Item;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+        where C: UnindexedConsumer<Self::Item>
+    {
+        self.base.flat_map(|x| x).drive_unindexed(consumer)
+    }
+}
--- a/src/iter/flatten.rs
+++ b/src/iter/flatten.rs
@@ -13,8 +13,9 @@ pub struct Flatten<I: ParallelIterator> {
 }
 
 /// Create a new `Flatten` iterator.
-pub fn new<I>(base: I) -> Flatten<I>
-    where I: ParallelIterator
+pub fn new<I, PI>(base: I) -> Flatten<I>
+    where I: ParallelIterator<Item = PI>,
+          PI: IntoParallelIterator + Send
 {
     Flatten {
         base: base,
@@ -23,7 +24,7 @@ pub fn new<I>(base: I) -> Flatten<I>
 
 impl<I, PI> ParallelIterator for Flatten<I>
     where I: ParallelIterator<Item = PI>,
-          PI: IntoParallelIterator + Send + Sync
+          PI: IntoParallelIterator + Send
 {
     type Item = PI::Item;
 
--- a/src/iter/flatten.rs
+++ b/src/iter/flatten.rs
@@ -13,6 +13,8 @@ pub struct Flatten<I: ParallelIterator> {
 }
 
 /// Create a new `Flatten` iterator.
+///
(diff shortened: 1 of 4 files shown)


src/iter/flatten.rs
```diff
@@ -0,0 +1,35 @@
+use super::internal::*;
+use super::*;
+
+/// `Flatten` turns each element to an iterator, then flattens these iterators
+/// together. This struct is created by the [`flatten()`] method on
+/// [`ParallelIterator`].
+///
+/// [`flatten()`]: trait.ParallelIterator.html#method.flatten
+/// [`ParallelIterator`]: trait.ParallelIterator.html
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
```
cuviper: Can you add a matching line to `tests/compile-fail/must_use.rs`?

mgattozzi: Yeah that shouldn't be hard to do.

src/iter/flatten.rs
```diff
@@ -0,0 +1,35 @@
...
+pub struct Flatten<I: ParallelIterator> {
+    base: I,
+}
+
+/// Create a new `Flatten` iterator.
+pub fn new<I>(base: I) -> Flatten<I>
+    where I: ParallelIterator
```
cuviper: Please add the same `Item` bounds as `impl ParallelIterator for Flatten`.

src/iter/flatten.rs
```diff
@@ -0,0 +1,35 @@
...
+{
+    Flatten {
+        base: base,
+    }
+}
+
+impl<I, PI> ParallelIterator for Flatten<I>
+    where I: ParallelIterator<Item = PI>,
+          PI: IntoParallelIterator + Send + Sync
```
cuviper: I don't think this needs `Sync`.

src/iter/mod.rs
```diff
@@ -215,6 +217,13 @@ pub trait ParallelIterator: Sized + Send {
         flat_map::new(self, map_op)
     }
 
+    /// Reduces a collection of items into one collection. If you have
+    /// a `Vec<Vec<i32>>` it can be used to turn it into a `Vec<i32>`.
+    fn flatten(self) -> Flatten<Self>
```
cuviper: This should also have something like `where Self::Item: IntoParallelIterator`.

cuviper: Oh, and I forgot to mention, I think the documentation would be better if it spoke in terms of iterators, not collections.  Something like the `flat_map` docs without the mapping part. :)

Perhaps: "An adaptor that flattens iterable `Item`s into one large iterator."

nikomatsakis: I'd like to see an example here. I find that is always super helpful to understand what's going on. e.g.,

```
use rayon::prelude::*;
let x: Vec<Vec<i32>> = vec![vec![1, 2], vec![3, 4]];
let y: Vec<i32> = x.into_par_iter().flatten().collect();
assert_eq!(y, vec![1, 2, 3, 4]);
```

mgattozzi:
@cuviper and @nikomatsakis I've updated the docs, code, and examples as asked!

src/iter/flatten.rs
```diff
@@ -0,0 +1,36 @@
...
+/// [`ParallelIterator`].
+///
+/// [`flatten()`]: trait.ParallelIterator.html#method.flatten
+/// [`ParallelIterator`]: trait.ParallelIterator.html
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct Flatten<I: ParallelIterator> {
+    base: I,
+}
+
+/// Create a new `Flatten` iterator.
```
nikomatsakis: Can you add a comment like "/// NB: Not part of public API"?h

nikomatsakis:
I forgot that I can do that just as easily.

PR #388, opened 2017-06-29
Fix doc error in flat_map

mgattozzi:
I found this small while working on #387.


cuviper:
😆

Thanks!

mgattozzi:
You're welcome! :D

Issue #386, opened 2017-06-28
implement `zip_longest` and `zip_eq`


nikomatsakis:
The `itertools` crate offers [a `zip_longest` method](https://docs.rs/itertools/0.6.0/itertools/trait.Itertools.html#method.zip_longest) and a `zip_eq`. These are variants on zip. `zip_eq` basically just asserts that the lengths of the two iterators are equal -- with parallel iterators, since we only zip things for which the lengths are known, we can trivially check this. `zip_longest` is a bit more complex. 

The implementation of `zip_longest` looks very much like [the existing `zip` iterator][z], I expect, except that -- [when we fall back to sequential execution](https://github.com/nikomatsakis/rayon/blob/29ab2edf82b68b7dcd6370ee296e741d487ba9bd/src/iter/zip.rs#L118) -- we can use the itertools `zip_longest` function instead of `zip`.

[z]: https://github.com/nikomatsakis/rayon/blob/29ab2edf82b68b7dcd6370ee296e741d487ba9bd/src/iter/zip.rs#L7-L10

nikomatsakis:
After some discussion with @cuviper, we are having second thoughts about the value of zip_longest. It seems a bit obscure. I'm not sure if people really use it. Given that it raises kind of interesting questions about whether to take itertools as a public dependency, maybe best to avoid it.

PR #389, opened 2017-06-29
Add zip_eq iterator adapter

chrisvittal:
This PR Implements a rayon version of zip_eq. Through discussion with Niko, I decided to implement it as an assertion on the length the inputs and return `Zip` rather than creating a thin wrapper around `Zip` to use as an extra type.

Implements part of #386.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -809,7 +809,6 @@ pub trait IndexedParallelIterator: ParallelIterator {
     ///
     /// // we should never get here
     /// assert_eq!(1, zipped.len());
-    ///
     /// ```
     fn zip_eq<Z>(mut self, zip_op: Z) -> Zip<Self, Z::Iter>
         where Z: IntoParallelIterator,
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -791,6 +791,36 @@ pub trait IndexedParallelIterator: ParallelIterator {
         zip::new(self, zip_op.into_par_iter())
     }
 
+    /// Same as `Zip` but requires that both iterators be the same length.
+    ///
+    /// # Panics
+    /// Will panic if `self` and `zip_op` are not the same length.
+    ///
+    /// ```should_panic
+    /// use rayon::prelude::*;
+    ///
+    /// let one = [1u8];
+    /// let two = [2u8, 2];
+    /// let one_iter = one.par_iter();
+    /// let two_iter = two.par_iter();
+    ///
+    /// // this will panic
+    /// let zipped: Vec<(&u8, &u8)> = one_iter.zip_eq(two_iter).collect();
+    ///
+    /// // we should never get here
+    /// assert_eq!(1, zipped.len());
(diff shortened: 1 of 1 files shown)


src/iter/mod.rs
```diff
@@ -791,6 +791,36 @@ pub trait IndexedParallelIterator: ParallelIterator {
...
+    /// let two = [2u8, 2];
+    /// let one_iter = one.par_iter();
+    /// let two_iter = two.par_iter();
+    ///
+    /// // this will panic
+    /// let zipped: Vec<(&u8, &u8)> = one_iter.zip_eq(two_iter).collect();
+    ///
+    /// // we should never get here
+    /// assert_eq!(1, zipped.len());
+    ///
```
nikomatsakis: Nit: remove blank line here

nikomatsakis:
@chrisvittal after some discussion with @cuviper, we are having second thoughts about the value of `zip_longest`. It seems a bit obscure. I'm not sure if people really use it. Given that it raises kind of interesting questions about whether to take itertools as a public dependency, maybe best to avoid it.

cuviper:
> Thinking on it more, I can come up with one argument: if we did change Zip so that the "base iterator" didn't necessarily have a known length (just the ability to drive an IndexedConsumer), then we wouldn't be able to do this assertion up front, in which case we would need a distinct return type.

I don't think that particular concern is a problem as long as `zip_eq` is still indexed -- it can do the assertion and still pass off to a more relaxed `Zip`.  But if you want `zip_eq` to also maybe be more relaxed, then yes it needs its own type.

I would lean that way anyway, towards a `ZipEq` wrapper, if only because that's more consistent with the rest of the crate where we're using distinct types even for things that likely would never diverge their implementations. e.g. `option::{Iter, IterMut, IntoIter}`

We have some funny macros in `src/delegate.rs` that should make this easier.  Something like:

```rust
delegate_indexed_iterator_item!{
    #[doc = "blah blah `zip_eq` blah..."]
    ZipEq<A, B> => Zip<A, B> : (A::Item, B::Item),
    impl<A: IndexedParallelIterator, B: IndexedParallelIterator>
}
```

But it's not *too* bad to write directly either.  I only wrote the macros because I had a lot of these to do for the collection iterators. :)

chrisvittal:
@nikomatsakis, I was considering much the same issues with respect to `zip_longest` this morning and believe I came to a similar conclusion.

@cuviper I spent a bit of time implementing `zip_eq` with a `ZipEq` type. That work can bee see at [this commit](https://github.com/chrisvittal/rayon/commit/02fa7decef2054295367459edd8745462240fe94), if both you and Niko think it's good, I can close this PR and open a new one.

cuviper:
@chrisvittal looks good to me!  Only thing I would add is documentation and `#[must_use]` on `ZipEq`, and a test for that must-use.  You can see how that was done for `Flatten` in #387.

chrisvittal:
Closed in favor of #392

PR #397, opened 2017-07-11
Add Repeat and RepeatN

cuviper:
This adds:
- unindexed, infinite `iter::repeat(elt) -> Repeat`
- indexed, finite `iter::repeatn(elt, n) -> RepeatN`
- `Repeat` methods `take(n)` and `zip(iter)` convert to `RepeatN`

This extends #337, opened separately for review since it's a fairly major change.

### DIFF
--- a/src/iter/repeat.rs
+++ b/src/iter/repeat.rs
@@ -1,11 +1,6 @@
-use super::internal::*;
 use super::*;
-use iter::*;
-use iter::internal::*;
-use std;
-use std::iter;
+use super::internal::*;
 use std::usize;
-
 
 pub struct Repeat<T> {
     element: T,
--- a/src/iter/repeat.rs
+++ b/src/iter/repeat.rs
@@ -1,0 +1,113 @@
+use super::internal::*;
+use super::*;
+use iter::*;
+use iter::internal::*;
+use std;
+use std::iter;
+use std::usize;
+
+
+pub struct Repeat<T> {
+    element: T,
+}
+
+pub fn repeat<T: Clone>(elt: T) -> Repeat<T> {
+    Repeat{element: elt}
+}
+
+impl<T: Clone> Clone for Repeat<T>{
+    fn clone(&self) -> Repeat<T> {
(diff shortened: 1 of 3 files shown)


cuviper:
I wanted to add some symmetry such that `IndexedParallelIterator::zip` could also accept iterators that might be truncatable, but I couldn't work out the coherence.  I wrote a new trait for this, with a blanket implementation to cover the existing `I: IntoParallelIterator, I::Iter: IndexedParallelIterator`, but rustc couldn't be convinced that `Repeat<T>` doesn't overlap that.  I suppose *in theory* a downstream crate could implement `IndexedParallelIterator` for their own `Repeat<Foo>`?

cuviper:
Rebased.

src/iter/repeat.rs
```diff
@@ -0,0 +1,206 @@
...
+    element: T,
+}
+
+/// Creates a parallel iterator that endlessly repeats `elt` (by
+/// cloning it). Note that this iterator has "infinite" length, so
+/// typically you would want to use `zip` or `take` or some other
+/// means to shorten it, or consider using
+/// [the `repeatn()` function](fn.repeatn.html) instead.
+///
+/// Example:
```
nikomatsakis: I think that people write `# Examples` or something conventionally?

src/iter/repeat.rs
```diff
@@ -0,0 +1,206 @@
...
+/// Iterator adaptor for [the `repeatn()` function](fn.repeatn.html).
+pub struct RepeatN<T: Clone + Send> {
+    element: T,
+    count: usize,
+}
+
+/// Creates a parallel iterator that produces `n` repeats of `elt`
+/// (by cloning it).
+///
+/// Example:
```
nikomatsakis: Same here.

cuviper:
Updated.

nikomatsakis:
Needs rebase. =)

cuviper:
bors r=nikomatsakis

PR #404, opened 2017-08-24
document the `Scope` struct

vishalsodani:
I see that a pull request for #233 was not accepted. So, added the details for the issue as required.


cuviper:
Thanks!

PR #409, opened 2017-09-04
Clarify signatures of sum and product

dtolnay:
In sum, we need to be able to compute Item+Item->S as well as S+S->S.


cuviper:
Ah, making the default parameter explicit - yes, I see how that helps clarity.  Thanks!

PR #410, opened 2017-09-04
Remove stray lifetime on IndexedParallelIterator::drive

dtolnay:


cuviper:
Strange that this didn't have a warning, even after I fixed the `allow(warnings)` conditional.

Thanks!

PR #411, opened 2017-09-05
Add a examples to the module docs

dns2utf8:
Hi

I wanted to add examples to the docs. My first example works as I expected.

My second example behaves very strange. It compiles when I use `i32` as the type of the `Range<T>`.
With any other type such as `usize`, `u64` or `i64` it does not compile.
Do you have an idea how to fix it?

Regards

---
Error message:
```
---- src/range.rs - range::Iter (line 23) stdout ----
        error[E0599]: no method named `zip` found for type `rayon::range::Iter<i64>` in the current scope
 --> src/range.rs:6:20
  |
6 |                   .zip(0..25)
  |                    ^^^
  |
  = note: the method `zip` exists but the following trait bounds were not satisfied:
          `&mut rayon::range::Iter<i64> : std::iter::Iterator`
```

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -783,6 +783,8 @@ impl<T: ParallelIterator> IntoParallelIterator for T {
 /// An iterator that supports "random access" to its data, meaning
 /// that you can split it at arbitrary indices and draw data from
 /// those points.
+///
+/// **Note:** Not implemented for `u64` and `i64` ranges
 pub trait IndexedParallelIterator: ParallelIterator {
     /// Collects the results of the iterator into the specified
     /// vector. The vector is always truncated before execution
--- a/src/range.rs
+++ b/src/range.rs
@@ -2,12 +2,42 @@
 //! (`Range<T>`); this is the type for values created by a `a..b`
 //! expression. You will rarely need to interact with it directly
 //! unless you have need to name one of the iterator types.
+//! 
+//! ```
+//! use rayon::prelude::*;
+//! 
+//! let r = (0..100u64).into_par_iter()
+//!                    .sum();
+//! 
+//! // compare result with sequential calculation
+//! assert_eq!((0..100).sum::<u64>(), r);
+//! ```
 
 use iter::*;
 use iter::internal::*;
 use std::ops::Range;
 
-/// Parallel iterator over a range
+/// Parallel iterator over a range, implemented for all integer types.
+///
+/// **Note:** The `zip` operation requires `IndexedParallelIterator`
+/// which is not implemented for `u64` or `i64`.
+///
+/// ```
+/// use rayon::prelude::*;
+///
+/// let p = (0..25usize).into_par_iter()
+///                   .zip(0..25usize)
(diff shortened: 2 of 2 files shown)


cuviper:
`Range<usize>` should work, but it's expected that `i64` and `u64` ranges don't implement `IndexedParallelIterator`, for the same reason they don't implement `std::iter::ExactSizeIterator`: a 64-bit range may overflow a `usize` length on 32-bit platforms.

dns2utf8:
Ah, I did not know that. I added a note to the docs.

cuviper:
I cherry-picked your CI fixes in #436.  If you'll please rebase the rest of your changes, we can see how CI fares now.

dns2utf8:
Rebased and tested. Thank you for the cerry-picking.

src/range.rs
```diff
@@ -2,12 +2,42 @@
 //! (`Range<T>`); this is the type for values created by a `a..b`
 //! expression. You will rarely need to interact with it directly
 //! unless you have need to name one of the iterator types.
+//! 
+//! ```
+//! use rayon::prelude::*;
+//! 
+//! let r = (0..100u64).into_par_iter()
+//!                    .fold(|| 0, |acc, cur| acc + cur)
+//!                    .sum();
```
cuviper: I notice in both examples you use `.fold(...).sum()`, when just `.sum()` should work fine.  Was this intentional?

src/range.rs
```diff
@@ -2,12 +2,42 @@
...
+//! // compare result with sequential calculation
+//! assert_eq!((0..100).sum::<u64>(), r);
+//! ```
 
 use iter::*;
 use iter::internal::*;
 use std::ops::Range;
 
 /// Parallel iterator over a range
+/// Implemented for `u32`, `i32`, `usize` and `isize`.
```
cuviper: We should have a blank doc-line between the intro and additional description.  Or if you want it as one intro paragraph, then it needs punctuation between the clauses.

Also, `ParallelIterator` is implemented for `range::Iter<T>` of _all_ integer types.  The difference is just that `i64` and `u64` don't implement `IndexedParallelIterator`, which you'd need for `zip`, for example.

dns2utf8:
Hm the `fold().sum()` was a strange idea.

dns2utf8:
Would you like me to squash the commits?

src/iter/mod.rs
```diff
@@ -781,6 +781,8 @@ impl<T: ParallelIterator> IntoParallelIterator for T {
 /// An iterator that supports "random access" to its data, meaning
 /// that you can split it at arbitrary indices and draw data from
 /// those points.
+///
+/// **Note:** Not implemented for `u64` and `i64`
```
cuviper: Small clarification, please: "Not implemented for `u64` and `i64` ranges."

cuviper:
Yeah, let's squash this one to its final state, thanks!

dns2utf8:
Fixed and squashed. Greetings from the RustFest impl days

cuviper:
Thanks!

PR #412, opened 2017-09-06
Raise the min version requirement because of the `?` operator

dns2utf8:
I fixed the tests for the latest nightly compiler in #411 but since the `?` operator is in the main code the minimal required version must be raised.

TODO
- [x] Update README
- [x] Update CHANGELOG

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.12.0
+  - 1.13.0
   - stable
   - nightly
 os:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: rust
 rust:
+  - 1.12.0
   - 1.13.0
   - stable
   - nightly
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,6 @@ default-features = false
 [dev-dependencies]
 compiletest_rs = "0.2.1"
 docopt = "0.7"
-futures = "0.1.7"
+futures = "=0.1.14"
 rand = "0.3"
 rustc-serialize = "0.3"
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ part, adding calls to Rayon should not change how your programs works
 at all, in fact. However, if you operate on mutexes or atomic
 integers, please see the [notes on atomicity](#atomicity).
 
-Rayon currently requires `rustc 1.12.0` or greater.
+Rayon currently requires `rustc 1.13.0` or greater.
 
 ### Using Rayon
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,7 @@
+# Release rayon 0.8.3
+
+- `?` operator requires rust 1.13 or greater
+
 # Release rayon 0.8.2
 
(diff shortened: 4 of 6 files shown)


dns2utf8:
I updated the readme to [1.13](https://blog.rust-lang.org/2016/11/10/Rust-1.13.html)

nikomatsakis:
Yeah, so I should probably not have merged that PR that used `?` -- however, I do think it may be worth raising the minimum requirement so that we can use `pub(crate)`, in which case we could also use `?`.

(I know that @cuviper and I have differing points of view on this matter, however, as I tend to view the minimum version requirement as more of a *courtesy* than a hard semver requirement, so I'd like to get his take.)

cuviper:
> I should probably not have merged that PR that used ?

Which one was that?

AFAICS the only problem here is actually in futures, where 0.1.15 uses `?` and 0.1.14 didn't.  If I manually run `cargo update -p futures --precise 0.1.14`, rust-1.12 can build rayon again.  Maybe we can just do that for CI, and eventually we're going to split rayon-futures anyway.

But this highlights my real gripe, that such updated requirements are viral, and there's no tooling to deal with it.  Any crate that tries to maintain compatibility is at the mercy of its dependencies to do the same.  I'll be a less grumpy about it when cargo learns how to ignore too-new crates, e.g. [RFC 1953](https://github.com/rust-lang/rfcs/pull/1953).

> I do think it may be worth raising the minimum requirement so that we can use pub(crate)

That would be rust-1.18, which is a big jump to me.  And IMO `pub(crate)` is a relatively minor benefit, that doesn't really enable us to do anything new.  Some of our hokey "free `pub fn` in private module" constructors could be `pub(crate)` methods instead, but this is just style, not functionality.  Do you have a more compelling use case?

dns2utf8:
You are correct. I grepped the rayon code and it contains no `?` operator.
I set the crate requirement to `=0.1.14` for testing now.

Imho the futures crate has about the same age as 1.12 and does not guarantee any compatibility with older compilers. Sure one could ask @alexcrichton to yank `0.1.15` and release it as `0.2.0` again.
Moving forward I assume the futures crate will use more features in upcoming releases. 
So raising the min version will be required at some point.

From my point of view it boils down to the questions
* who relies on 1.12? 
* or more general what versions are currently used?

For my projects, I check the debian repo where the current stable release is [1.14.0](https://packages.debian.org/stretch/rustc)

alexcrichton:
I would highly discourage dependencies like `=1.1.14` as it tends to be pretty hostile to downstream dependencies (foces everyone to use an older version), and I don't think I'll be yanking 1.1.15 because Rust with `?` is pretty old at this point

cuviper:
We don't really know what compiler folks may be stuck with, and that's part of the problem. The 2017 survey showed most respondents with the current version through rustup, but there were still a tail of people behind, even "1.10 or older".

I wouldn't suggest that futures should yank 0.1.15.  I just wish that their compatibility choice wasn't so viral.  And I also agree that pinning an older version in rayon has its own problems. I suggest just stepping down the version in the CI scripts for now.

cuviper:
#436 separated rayon-futures, so now we don't need to bump the min version for rayon.

PR #435, opened 2017-09-08
Demo cleanup

### LINKED ISSUE #419: doc: Examples should use `?` not `try!`
Looks like `rayon-demo/src/tsp/mod.rs` may still be using `try!`

gsquire:
I don't know if you want to take it a step further and remove `try!` in the library code, but for now I just removed it from the demo.

Fixes #419


cuviper:
> I don't know if you want to take it a step further and remove try! in the library code

Nah, I don't want to bump the minimum rustc for rayon if we can help it.  In rayon-demo though, that's fine.  Thanks!

PR #436, opened 2017-09-09
Extract rayon-futures as a separate crate

cuviper:
This is a rebase of #344, also moving the `futures` support to a separate
`rayon-futures` crate.  No other changes were made to the actual code.


cuviper:
I think we should go ahead and do this soon, since the core additions are still `rayon_unstable`.  We don't even have to publish the new crate right now, although I did go ahead and reserve the name.

cuviper:
I added more CI fixes, and finally we're green again! Merging...

Issue #424, opened 2017-09-06
doc: Crate missing `html_root_url` attribute


alexcrichton:
This'll help cross-crate links in foreign crates! It should be set to https://docs.rs/rayon/$version

KodrAus:
Looks like this one's all done @cuviper.

cuviper:
Let's get the same addition for rayon-core and the pending rayon-futures.

smt923:
@cuviper, I'm working on adding the attribute to rayon-core and rayon-futures, should I just match the docs.rs version to whatever is in the Cargo.toml? For both crates, that current version doesn't have published docs, but should I just leave it and assume when a release gets pushed it'll all work?

PR #437, opened 2017-09-11
Fix example and add missing attribute to crate

### LINKED ISSUE #419: doc: Examples should use `?` not `try!`

jdr023:
Fixes #419 and #424.


cuviper:
I merged the same `try!` change in #435, but it looks like we can still merge cleanly to get the attribute update.  Thanks!

PR #438, opened 2017-09-12
Add Crate level documentation and Cargo attributes

### LINKED ISSUE #418: doc: Add crate-level documentation
Would be great to have some of the README-like information (though perhaps not quite as thorough) along with some items like:

* Examples
* Common types
* Rationale for crate layout
* etc

### LINKED ISSUE #423: doc: Missing attributes in Cargo.toml
The manifest should specify  `readme`, `keywords`, and `categories`

AndyGauge:
Fixes #418 
Fixes #423

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,9 @@ description = "Simple work-stealing parallelism for Rust"
 license = "Apache-2.0/MIT"
 repository = "https://github.com/nikomatsakis/rayon"
 documentation = "https://docs.rs/rayon/"
+readme = "README.md"
+keywords = ["parallel", "thread", "concurrency", "join", "performance"]
+categories = ["concurrency"]
 
 [workspace]
 members = ["rayon-demo", "rayon-core", "rayon-futures"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,33 +18,37 @@
 //!
 //! # Parallel Iterators
 //!
-//! Parallel iterators are formed using [`par_iter`] and `par_iter_mut` functions
-//! followed by a computation chain.  Computations on the iterator can take the
+//! Parallel iterators are formed using [`par_iter`], [`par_iter_mut`], and [`into_par_iter`]
+//! functions to iterate by shared reference, mutable reference, or by value respectively.
+//! These iterators are chained with computations that can take the
 //! shape of `map` or `for_each` as an example.  This solves [embarrassingly]
 //! parallel tasks that are completely independent of one another.
 //!
-//! [`par_iter`]: iter/trait.ParallelIterator.html
+//! [`par_iter`]: iter/trait.IntoParallelRefIterator.html
+//! [`par_iter_mut`]: iter/trait.IntoParallelRefMutIterator.html
+//! [`into_par_iter`]: iter/trait.IntoParallelIterator.html#tymethod.into_par_iter
 //! [embarrassingly]: https://en.wikipedia.org/wiki/Embarrassingly_parallel
 //!
 //! # Examples
+//!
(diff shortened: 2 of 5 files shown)


nikomatsakis:
@AndyGauge I see some travis failures; I suspect they are unrelated to this PR, but would you mind rebasing?

AndyGauge:
The upgrade from try! to ? has caused 1.12 to error.  I think that's a travis.yml change that is required.  I'll submit an issue.

AndyGauge:
OK, this is ready for review.

Thanks

src/lib.rs
```diff
@@ -9,6 +9,94 @@
...
+//! Data-parallelism library that is easy to convert sequential computations into parallel.
+//!
+//! Rayon is lightweight and convenient for application to existing code.  It guarantees
+//! data-race free executions, and takes advantage of parallelism when sensible, based
+//! on work-load at runtime.  There are two categories of rayon workloads: parallel
+//! iterators and multi-branched recursion (`join` method).
+//!
+//! # Parallel Iterators
+//!
+//! Parallel iterators are formed using [`par_iter`] and `par_iter_mut` functions
```
cuviper: It's strange to me that you've linked `par_iter` to the `ParallelIterator` trait.  It is good to have that link up front, but `par_iter` itself comes from `IntoParallelRefIterator`.  There are actually three such conversions, `into_par_iter()`, `par_iter()`, and `par_iter_mut()`, to iterate respectively by value, by shared reference, or by mutable reference.

src/lib.rs
```diff
@@ -9,6 +9,94 @@
...
+//! # fn main() {
+//! let mut chars: Vec<char> = "A man, a plan, a canal - Panama!".chars().collect();
+//! let encrypted: String = chars.par_iter().map(|c| {
+//!        match *c {
+//!            'A' ... 'M' | 'a' ... 'm' => ((*c as u8) + 13) as char,
+//!            'N' ... 'Z' | 'n' ... 'z' => ((*c as u8) - 13) as char,
+//!            _ => *c
+//!        }
+//!    }).collect();
+//! # }
```
cuviper: It would be good to `assert_eq!` the result so readers know what to expect, and so it also serves as a doctest.

src/lib.rs
```diff
@@ -9,6 +9,94 @@
...
+//! ```
+//!
+//! Here a string is encrypted using ROT13 leveraging parallelism.  Once all the
+//! threads are complete, they are collected into a string.
```
cuviper: I think it reads better to place such descriptions *before* the example code.

Cargo.toml
```diff
@@ -8,6 +8,9 @@ description = "Simple work-stealing parallelism for Rust"
 license = "Apache-2.0/MIT"
 repository = "https://github.com/nikomatsakis/rayon"
 documentation = "https://docs.rs/rayon/"
+readme = "README.md"
+keywords = ["parallel", "thread", "concurrency", "join", "performance"]
+categories = ["concurrency"]
```
cuviper: Can you also add these attributes to rayon-core and rayon-futures?

AndyGauge: I'm not sure if readme attribute is appropriate.  I added '..\readme.md'.  Not sure if that is helpful or if the other 'crates' need their own readmes.

cuviper: I have no idea if crates.io will work with `../README.md` -- I doubt it though, because it's probably looking at the individual uploaded crates, not the repository.

We do have a readme in rayon-futures, but not rayon-core.  The latter could just be a simple placeholder, saying something like "Core APIs for Rayon; see the rayon crate for more details."

cuviper:
Thanks!

PR #442, opened 2017-09-18
Implement `Debug` for public types

### LINKED ISSUE #427: Public types should implement `Debug`
This can be linted with `#![deny(missing_debug_implementations)]` and then basically hitting "go" until everything compiles again!

tmccombs:
Fixes #427.

I included what I though was important in the implementations, but I'm open to suggestions.

### DIFF
--- a/rayon-core/src/scope/mod.rs
+++ b/rayon-core/src/scope/mod.rs
@@ -350,11 +350,10 @@ impl<'scope> Scope<'scope> {
 
 impl<'scope> fmt::Debug for Scope<'scope> {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        use latch::LatchProbe;
         fmt.debug_struct("Scope")
             .field("owner_thread", &self.owner_thread)
             .field("panic", &self.panic)
-            .field("job_completed", &self.job_completed_latch.probe())
+            .field("job_completed_latch", &self.job_completed_latch)
             .finish()
     }
 }
--- a/rayon-core/src/scope/mod.rs
+++ b/rayon-core/src/scope/mod.rs
@@ -2,6 +2,7 @@ use latch::{Latch, CountLatch};
 use log::Event::*;
 use job::HeapJob;
 use std::any::Any;
+use std::fmt;
 use std::marker::PhantomData;
 use std::mem;
 use std::ptr;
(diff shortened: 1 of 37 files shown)


nikomatsakis:
@tmccombs hmm -- I'm seeing a bunch of travis failures! do you mind taking a look?

KodrAus:
That'll be the `deny(missing_debug)`, which includes `pub` structs even if they aren't re-exported.

@nikomatsakis would you rather keep the lint and work around `pub(not pub)` structs or just remove it?

tmccombs:
Ah I didn't run with rayon_unstable feature turned on locally. I'll look into them.

tmccombs:
hmm, I had a typo, so it didn't actually get everything, I'll work on getting all the structs I missed.

rayon-futures/src/lib.rs
```diff
@@ -113,6 +115,11 @@ impl<T, E> Drop for RayonFuture<T, E> {
     }
 }
 
+impl<T, E> fmt::Debug for RayonFuture<T, E> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.write_str("RayonFuture(...)")
```
KodrAus: This could probably just be `fmt.debug_struct("RayonFuture").finish()`.

rayon-core/src/scope/mod.rs
```diff
@@ -346,3 +347,14 @@ impl<'scope> Scope<'scope> {
...
     }
 }
+
+impl<'scope> fmt::Debug for Scope<'scope> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        use latch::LatchProbe;
+        fmt.debug_struct("Scope")
+            .field("owner_thread", &self.owner_thread)
+            .field("panic", &self.panic)
+            .field("job_completed", &self.job_completed_latch.probe())
```
cuviper: I think `probe()` will always be true until the scope is out of the user's hands.  For debugging purposes, the actual counter value would be more interesting.  How about adding `Debug` to `CountLatch` and show that here?

tmccombs: alright

cuviper:
I added a couple more that hit master in the meantime.

Thanks!

PR #443, opened 2017-09-19
Tighten Send and Sync requirements

### LINKED ISSUE #430: Scope::spawn doesn't require `Send`
I think it should though!

cuviper:
This is a **breaking change** to `rayon-core`, necessary for soundness.

- `Scope::spawn` now requires `Send` for the closure.  Fixes #430.

- `ThreadPool::install` now requires `Send` for the return value.

- (unstable) `internal::Task` now requires `Send + Sync` instead of just
`Send`, so its use of `Arc` properly implements `Send` and `Sync` too.

### DIFF
--- a/rayon-core/src/scope/mod.rs
+++ b/rayon-core/src/scope/mod.rs
@@ -6,8 +6,9 @@ use std::fmt;
 use std::marker::PhantomData;
 use std::mem;
 use std::ptr;
+use std::sync::Arc;
 use std::sync::atomic::{AtomicPtr, Ordering};
-use registry::{in_worker, WorkerThread};
+use registry::{in_worker, WorkerThread, Registry};
 use unwind;
 
 #[cfg(test)]
@@ -21,7 +22,10 @@ pub struct Scope<'scope> {
     /// thread where `scope()` was executed (note that individual jobs
     /// may be executing on different worker threads, though they
     /// should always be within the same pool of threads)
-    owner_thread: *const WorkerThread,
+    owner_thread_index: usize,
+
+    /// thread registry where `scope()` was executed.
+    registry: Arc<Registry>,
 
     /// if some job panicked, the error is stored here; it will be
     /// propagated to the one who created the scope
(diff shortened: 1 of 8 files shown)


rayon-core/src/scope/mod.rs
```diff
@@ -278,7 +285,7 @@ impl<'scope> Scope<'scope> {
 
             // the `Scope` is not send or sync, and we only give out
             // pointers to it from within a worker thread
-            debug_assert!(!WorkerThread::current().is_null());
+            debug_assert!(!worker_thread.is_null());
```
cuviper: Note, `Scope` now *does* implement `Sync`, so this comment isn't true anymore.  We need to evaluate whether that's actually safe.  I think here, this `!is_null()` should probably be a full `assert!`, and perhaps we assert that the current registry matches that saved in the `Scope` too.

rayon-core/src/scope/mod.rs
```diff
@@ -278,7 +285,7 @@ impl<'scope> Scope<'scope> {
...
 
             let worker_thread = &*worker_thread;
             worker_thread.push(job_ref);
```
cuviper: Maybe this can just switch to `scope.registry.inject_or_push(job_ref)`, so if the scope does get carried out of the threadpool, it still spawns back to the right pool.  The `'scope` lifetime should still ensure the right stuff...

cuviper: Done.

rayon-core/src/scope/mod.rs
```diff
@@ -268,20 +275,13 @@ impl<'scope> Scope<'scope> {
...
                 .as_job_ref();
-            let worker_thread = WorkerThread::current();
-
-            // the `Scope` is not send or sync, and we only give out
-            // pointers to it from within a worker thread
-            debug_assert!(!WorkerThread::current().is_null());
-
-            let worker_thread = &*worker_thread;
-            worker_thread.push(job_ref);
+            self.registry.inject_or_push(job_ref);
```
nikomatsakis: Let's add a comment here saying that it would be unsafe to just invoke `push` before we don't know that the scope was not somehow sent to another thread.

cuviper: Done.

cuviper:
Added a fix for #446, since that also needed the change to `install()` here.

PR #444, opened 2017-09-21
Add html_root_url for core and futures, with a reminder in Cargo.toml

### LINKED ISSUE #424: doc: Crate missing `html_root_url` attribute

smt923:
Should close #424 - I've used some shorter version links for the docs, but if I should just make these exactly what version is specified in the Cargo.toml I can change it, those versions just don't have any published docs yet and making a PR with links that 404'd didn't feel right (but perhaps this is totally fine)

### DIFF
--- a/rayon-futures/src/lib.rs
+++ b/rayon-futures/src/lib.rs
@@ -2,6 +2,8 @@
 //!
 //! See `README.md` for details.
 
+#![doc(html_root_url = "https://docs.rs/rayon-futures/0")]
+
 extern crate futures;
 extern crate rayon_core;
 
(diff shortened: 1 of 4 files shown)


rayon-futures/src/lib.rs
```diff
@@ -2,6 +2,8 @@
 //!
 //! See `README.md` for details.
 
+#![doc(html_root_url = "https://docs.rs/rayon-futures/0")]
```
KodrAus: Let's make this `0.1` so it's in line with what's in the `Cargo.toml`, even if that hasn't been released yet.

cuviper:
Thanks!

PR #445, opened 2017-09-22
Implement interleave_shortest()

### LINKED ISSUE #384: implement `interleave` and `interleave_shortest` from itertools
The `itertools` crate offers [a `interleave` method](https://docs.rs/itertools/0.6.0/itertools/trait.Itertools.html#method.interleave) and `interleave_shortest`. From the docs, "Alternate elements from two iterators until both run out." First task is probably to figure out the *precise* specification here (e.g., by reading the code), since I find that a bit ambiguous, and both methods use the same example, so it's hard to see the difference.

Of the existing adapters, `interleave` most resembles [`zip`](https://github.com/nikomatsakis/rayon/blob/29ab2edf82b68b7dcd6370ee296e741d487ba9bd/src/iter/zip.rs#L7-L10). I believe this means that, like `zip`, it will require that the two input iterators support `IndexedParallelIterator` (see the [parallel iterator README](https://github.com/nikomatsakis/rayon/blob/29ab2edf82b68b7dcd6370ee296e741d487ba9bd/src/iter/README.md) for more details).

The basic sketch, like zip, would be to convert both of the input iterators into producers. They can then be split in lockstep. At the point where we convert to sequential execution we can just leverage the `itertools` crate (which is already a dependency of `rayon`).

laumann:
Add interleave_shortest() to IndexedParallelIterator that acts like
interleave, except it terminates as soon as one of the iterators is
exhausted.

The other part of/Fixes #384

### DIFF
--- a/src/iter/interleave.rs
+++ b/src/iter/interleave.rs
@@ -58,7 +58,10 @@ impl<I, J> IndexedParallelIterator for Interleave<I, J>
     }
 
     fn len(&mut self) -> usize {
-        self.i.len() + self.j.len()
+        self.i
+            .len()
+            .checked_add(self.j.len())
+            .expect("overflow")
     }
 
     fn with_producer<CB>(mut self, callback: CB) -> CB::Output
--- a/src/iter/interleave_shortest.rs
+++ b/src/iter/interleave_shortest.rs
@@ -1,0 +1,81 @@
+use super::internal::*;
+use super::*;
+use std::cmp;
+use std::iter;
+
+/// `InterleaveShortest` is an iterator that works similarly to
+/// `Interleave`, but this version stops returning elements once one
+/// of the iterators run out.
+///
+/// This struct is created by the [`interleave_shortest()`] method on
+/// [`IndexedParallelIterator`].
+///
+/// [`interleave_shortest()`]: trait.IndexedParallelIterator.html#method.interleave_shortest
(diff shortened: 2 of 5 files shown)


laumann:
Just wondering, why does the [`size_hint()` implementation for InterleaveShortest](https://docs.rs/itertools/0.6.3/src/itertools/adaptors/mod.rs.html#148-180) in itertools look so complicated? Makes me wonder if I'm missing something here...

cuviper:
Looks good!  Just one thing, can you add this to `tests/compile-fail/must_use.rs`?  I forgot this for `interleave` too.

> Just wondering, why does the size_hint() implementation for InterleaveShortest in itertools look so complicated?

Probably because itertools doesn't require them to be `ExactSizeIterator`, so it's going through hoops to combine the min/max limits from the two iterators.  We're also lax about overflow when two iterators combine to overflow `usize`.  Actually, let's improve that.  `Interleave` can copy [what `Chain` does](https://github.com/nikomatsakis/rayon/blob/ff57a0217992a49fbd18b07fdcdd5b683589d6f6/src/iter/chain.rs#L79-L84), and while we're here, it should also pass through [`min_len` and `max_len`](https://github.com/nikomatsakis/rayon/blob/ff57a0217992a49fbd18b07fdcdd5b683589d6f6/src/iter/chain.rs#L177-L183).

cuviper:
Thanks!

PR #447, opened 2017-09-25
Transition from deprecated Unpark to Notify

tmccombs:
I'm not entirely sure if this is the correct way to create a NotifyHandle from a ScopeFuture, but I do think storing the NotifyHandle in the ScopeFutureContes is the correct thing to do.


cuviper:
#357 is already working on this... Maybe you can offer a review?

tmccombs:
ah, didn't see that.

PR #450, opened 2017-09-29
Added documentation for iter types and structs

Hittherhod:
Added documentation for all iter traits and structs. Updated links to par_iter and par_iter_mut so they link directly to the methods (previously they linked to the trait page, not the specific method).

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -86,12 +86,13 @@ mod unzip;
 mod test;
 
 /// `IntoParallelIterator` implements the conversion to a [`ParallelIterator`].
+///
 /// By implementing `IntoParallelIterator` for a type, you define how it will
 /// transformed into an iterator. This is a parallel version of the standard
-/// library's `std::iter::IntoIterator` trait.
+/// library's [`std::iter::IntoIterator`] trait.
 ///
 /// [`ParallelIterator`]: trait.ParallelIterator.html
-/// [`into_par_iter()`]: trait.IntoParallelIterator.html#method.into_par_iter
+/// [`std::iter::IntoIterator`]: https://doc.rust-lang.org/std/iter/trait.IntoIterator.html
 pub trait IntoParallelIterator {
     type Iter: ParallelIterator<Item = Self::Item>;
     type Item: Send;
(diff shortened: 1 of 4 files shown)


src/iter/mod.rs
```diff
@@ -85,13 +85,26 @@ mod unzip;
 #[cfg(test)]
 mod test;
 
+/// `IntoParallelIterator` implements the conversion to a [`ParallelIterator`].
+/// By implementing `IntoParallelIterator` for a type, you define how it will
+/// transformed into an iterator. This is a parallel version of the standard
+/// library's `std::iter::IntoIterator` trait.
```
KodrAus: We could make `std::iter::IntoIterator` link to https://doc.rust-lang.org/std/iter/trait.IntoIterator.html

Hittherhod: @KodrAus I added that link as well

src/iter/mod.rs
```diff
@@ -110,6 +124,13 @@ impl<'data, I: 'data + ?Sized> IntoParallelRefIterator<'data> for I
     }
 }
 
+
+/// `IntoParallelRefMutIterator` implements the conversion to a [`ParallelIterator`] providing mutable references to the data.
+///
+/// This is a parallel version of the `iter_mut()` method 
+/// defined by various collections.
```
cuviper: We should probably also mention that this trait is automatically implemented `for I where &mut I: IntoParallelIterator`.  In other words, one should probably go implement that rather than implementing this trait directly.

src/iter/mod.rs
```diff
@@ -85,13 +85,27 @@ mod unzip;
...
 pub trait IntoParallelIterator {
     type Iter: ParallelIterator<Item = Self::Item>;
     type Item: Send;
 
     fn into_par_iter(self) -> Self::Iter;
 }
 
+/// `IntoParallelRefIterator` implements the conversion to a [`ParallelIterator`].
+/// This is a parallel version of the `iter()` method 
+/// defined by various collections.
```
cuviper: I'd like this to mirror the `IntoParallelRefMutIterator` docs, i.e. "providing shared references ...", blank line to break the paragraph after the first sentence, and mention a similar auto-implementation.

src/iter/mod.rs
```diff
@@ -85,13 +85,27 @@ mod unzip;
...
+///
+/// [`ParallelIterator`]: trait.ParallelIterator.html
+/// [`par_iter()`]: trait.IntoParallelRefIterator.html#method.par_iter
```
cuviper: Is this `par_iter()` link used?

src/iter/mod.rs
```diff
@@ -85,13 +85,27 @@ mod unzip;
 #[cfg(test)]
 mod test;
 
+/// `IntoParallelIterator` implements the conversion to a [`ParallelIterator`].
+/// By implementing `IntoParallelIterator` for a type, you define how it will
+/// transformed into an iterator. This is a parallel version of the standard
+/// library's [`std::iter::IntoIterator`] trait.
+///
+/// [`ParallelIterator`]: trait.ParallelIterator.html
+/// [`into_par_iter()`]: trait.IntoParallelIterator.html#method.into_par_iter
```
cuviper: Is this `into_par_iter()` link used?

Hittherhod:
@cuviper I updated the docs to be more consistent. I removed the unused links and added in the suggestion to just implement IntoParallelIterator

cuviper:
Looks great, thanks!

PR #451, opened 2017-10-06
Enable bors-ng

cuviper:


cuviper:
I've now pruned pull-request CI down to just Linux 1.12 and nightly+unstable, for faster feedback and to reduce queue delays (especially from the overburdened OSX).  We should still get full Travis and Appveyor test results when bors pushes to its branches.

bors r+

PR #452, opened 2017-10-08
Fix unused warnings in interleave

cuviper:


cuviper:
Trivial fixes.

bors r+

PR #456, opened 2017-10-12
Implement Clone for parallel iterators

### LINKED ISSUE #416: Implement `Clone` for iterators where possible
If possible, `Clone` is a great trait to have? Maybe Rayon's use case though precludes the ability to do this? (if so please feel free to close!)

cuviper:
Most are simply `#[derive(Clone)]`, except for `&T` sorts of iterators where we don't want the derived `T: Clone`.

Fixes #416.


cuviper:
I did some `Debug` cleanup that touches a lot of the same lines, so I'm just adding it here.

PR #458, opened 2017-10-14
Inherent mutability -> Interior mutablility

MajorBreakfast:
The rust book calls this "interior mutability". I think that the two terms were just confused because they sound awfully similar.

First edition: https://doc.rust-lang.org/book/first-edition/mutability.html
Second edition: https://doc.rust-lang.org/book/second-edition/ch15-05-interior-mutability.html#refcellt-and-the-interior-mutability-pattern


cuviper:
Ah, yep, even the [`std::cell` docs](https://doc.rust-lang.org/stable/std/cell/index.html) call it "interior". Thanks!

bors r+

PR #459, opened 2017-10-15
Document that the sum in the examples is a parallel reduction

marmistrz:


README.md
```diff
@@ -101,6 +101,9 @@ fn sum_of_squares(input: &[i32]) -> i32 {
 }
 ```
 
+Note that the `sum()` call is not a sequential one. It is, precisely,
+a logarithmic, tree-like reduction known from, for example, OpenMP.
```
cuviper: I appreciate what you're trying to say, but it's worded a bit awkwardly, especially with two parentheticals.  Can you try simplifying the second sentence?

marmistrz: I completely forgot about it. Is the change still relevant or should I go ahead simplifying it?

cuviper: There's a new paragraph following the example you're describing, but I'm not sure if that leads the reader in the same direction that you're addressing.  I'll let you look at the context to see if you think your addition is still needed.  (My own familiarity makes it harder to judge what could be confusing for newcomers...)

cuviper:
Ping @marmistrz, did you get another chance to look?  If we don't hear from you, I may try to reword this myself in a new PR.

PR #460, opened 2017-10-15
Document setting the number of threads

marmistrz:


README.md
```diff
@@ -85,6 +85,16 @@ For more information on demos, try:
 
 **Note:** While Rayon is usable as a library with the stable compiler, running demos or executing tests requires nightly Rust.
 
+### How many threads will Rayon spawn?
+
+By default, Rayon uses the same number of threads as the number of CPUs
+available. Note that on systems with hyperthreading enabled this
+equals to the number of logical cores and the physical ones.
```
cuviper: Do you mean "and not the physical ones"?  The way it's written sounds like perhaps you'd add these numbers, but they're overlapping values.  e.g. I would say my current CPU has 4 physical cores and 8 logical cores.

marmistrz: Of course there should be a *not* here! I'll fix it.

cuviper:
Thanks!

bors r+

cuviper:
Meh, spurious appveyor failure, and this change doesn't touch code anyway.

PR #462, opened 2017-10-17
Fix some markdown formatting in docs

mbrubeck:


mbrubeck:
I discovered this when building with `--enable-commonmark` (rust-lang/rust#44229) but it's actually an issue with the current default renderer (hoedown) that is fixed by the new one (pulldown-cmark).

cuviper:
Thanks!

bors r+

PR #463, opened 2017-10-19
rename rayon::iter::internal to rayon::iter::plumbing

### LINKED ISSUE #415: Rename/move the iter::internal module
This module, while an internal implementation detail, still adheres to the semver guidelines of the crate and can be useful to implementors! In the triage meeting the name "porcelain" was discussed for this as well.

lvillani:
This should fix #415.


cuviper:
r? @nikomatsakis - are you OK with the "plumbing" name?

We should update this header too:

> These should be considered highly unstable: users of parallel iterators should not need to interact with them directly.

Let's drop the "highly unstable" part, because we're not actually making breaking changes here any differently than the rest of the API.  Instead, we should clarify that these parts are only needed for implementing new parallel iterators.

cuviper:
I rebased and tweaked that header.

bors r+

PR #464, opened 2017-10-24
Add "parallel" iter::empty() and iter::once()

cuviper:
These are simplistic producers, mirroring their `std` namesakes.  They
are essentially just optimized equivalents to iterating `None` and
`Some(x)`, where `empty()` is a ZST, `once()` doesn't need the enum tag
for an `Option`, and neither have the associated enum branches.


cuviper:
I need to add docs before we merge this.

PR #466, opened 2017-10-29
Add documentation examples for parallel iterator methods

mamuleanu:
Closes #420

PR #467, opened 2017-11-05
Add examples to the ParallelIterator documentation

Kerollmops:
The last remaining examples to do are `map_with` and `fold_with` that I don't really understand.

Could you please explain to me the mecanics ?

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -179,9 +179,12 @@ pub trait ParallelIterator: Sized + Send {
 
     /// Executes `OP` on each item produced by the iterator, in parallel.
     ///
+    /// # Examples
+    ///
     /// ```
     /// use rayon::prelude::*;
-    /// (0..100).into_par_iter().for_each(|x| println!("{}", x));
+    ///
+    /// (0..100).into_par_iter().for_each(|x| println!("{:?}", x));
     /// ```
     fn for_each<OP>(self, op: OP)
         where OP: Fn(Self::Item) + Sync + Send
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -1020,6 +1020,19 @@ pub trait ParallelIterator: Sized + Send {
     /// efficiently with precise knowledge of how many elements the
     /// iterator contains, and even allows you to reuse existing
     /// vectors' backing stores rather than allocating fresh vectors.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    ///
+    /// let a = [(0, 1), (1, 2), (2, 3), (3, 4)];
+    ///
+    /// let (left, right): (Vec<_>, Vec<_>) = a.par_iter().cloned().unzip();
+    ///
+    /// assert_eq!(left, [0, 1, 2, 3]);
+    /// assert_eq!(right, [1, 2, 3, 4]);
+    /// ```
     fn unzip<A, B, FromA, FromB>(self) -> (FromA, FromB)
         where Self: ParallelIterator<Item = (A, B)>,
               FromA: Default + Send + ParallelExtend<A>,
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -881,10 +881,15 @@ pub trait ParallelIterator: Sized + Send {
     /// just want the first match that discovered anywhere in the iterator,
     /// `find_any` is a better choice.
     ///
+    /// # Exmaples
+    ///
     /// ```
     /// use rayon::prelude::*;
+    ///
     /// let a = [1, 2, 3, 3];
+    ///
     /// assert_eq!(a.par_iter().find_first(|&&x| x == 3), Some(&3));
+    ///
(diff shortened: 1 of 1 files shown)


cuviper:
Thanks for the examples!  However, this is surely going to collide in a big way with #466, which I just approved since it came in first.  When that lands, please do compare and merge.  Where your examples might show some distinctly interesting case, feel free to keep both, otherwise we should keep whichever example is most easily understood.

The CI failure should be fixed by #468.

cuviper:
As for `map_with`, `for_each_with`, and `fold_with`, the idea is these can be used to pass around some supplemental data type which implements `Clone` and `Send`, but not necessarily `Sync` where it could just be shared directly.

The primary motivating example of this is the [`mpsc::Sender`](https://doc.rust-lang.org/nightly/std/sync/mpsc/struct.Sender.html) with `Clone` and `Send` but `!Sync`.  Using these `_with` methods, you can get a clone of the sender in each thread as-needed.

There aren't many options to do this sort of thing manually -- e.g. maybe a `Mutex<Sender>` that you lock and clone in the `fold()` identity function, but this is kind of ugly.

Kerollmops:
Thanks for the indications. I have rebased my branch and done some improvement on the previously merged examples.

I will improve some of the examples already present here and add examples to the `map_with`, `for_each_with`, and `fold_with` methods, so please don't merge it for now.

Thank you you for the details ebout theses functions.

Kerollmops:
I will just add an example to the `for_each_with`. `map_with` and `fold_with` does need much more example to use and like you said the `Sender` method is kind of ugly.

You can review this PR and merge it if you think it's good 😄

src/iter/mod.rs
```diff
@@ -825,6 +1025,18 @@ pub trait ParallelIterator: Sized + Send {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    ///
+    /// let a = [Some(1), Some(2), Some(4), None, Some(6), Some(8)];
+    ///
+    /// let sum: usize = a.par_iter().cloned().while_some().sum::<usize>();
+    ///
+    /// assert_eq!(sum, 7);
```
cuviper: I think this test is getting lucky too, but perhaps we need to clarify the documented behavior.  It halts as soon as a `None` is found *anywhere*, in arbitrary parallel order.  You can't assume that it will see all of the sequentially-prior `Some` values before finding a `None`, and it may also visit `Some` values that are sequentially after a `None`.  In the extreme, this `while_some` could produce nothing if it hit the `None` first, or it may even produce all of the `Some` items if it hit the `None` last!

src/iter/mod.rs
```diff
@@ -195,6 +198,21 @@ pub trait ParallelIterator: Sized + Send {
...
+    /// use std::sync::mpsc::channel;
+    /// use rayon::prelude::*;
+    ///
+    /// let (sender, receiver) = channel();
+    ///
+    /// (0..5).into_par_iter().for_each_with(sender, |s, x| s.send(x).unwrap());
+    ///
+    /// let res: Vec<_> = receiver.iter().collect();
+    ///
+    /// assert_eq!(&res[..], &[0, 1, 2, 3, 4])
```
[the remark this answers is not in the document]
Kerollmops: Ho ! Yes you are right...

src/iter/mod.rs
Kerollmops: So what kind of example can I use here ? It's some sort of random that occurs here...

cuviper: Perhaps `while_some().count() < a.len()` is a better example?  It's somewhat random which `Some` elements will make it through, but we can guarantee that it will be less than the total vector length.

You can make the same assertion about length with `filter_map` though - I'm not sure how to clearly demonstrate that `while_some` short-circuits.  Maybe we could show an infinite source iterator getting stopped, like `rayon::iter::repeat(None).while_some().count() == 0`.  Or for another example, [`check_while_some`](https://github.com/rayon-rs/rayon/blob/e16ae7779fd67bcb971cfd4d6a9f03a8df16b626/src/iter/test.rs#L1253-L1273) uses an atomic to count how many values were produced behind the scenes.

As a practical use-case, `FromParallelIterator` [for `Option`](https://github.com/rayon-rs/rayon/blob/e16ae7779fd67bcb971cfd4d6a9f03a8df16b626/src/option.rs#L156) and [for `Result`](https://github.com/rayon-rs/rayon/blob/e16ae7779fd67bcb971cfd4d6a9f03a8df16b626/src/result.rs#L105) each use `while_some()` to short-circuit their collection when they see a `None` or `Err`.

Kerollmops: Ok, so I will use the test code, it is clear and short.

Thank you for the explaination, really usefull and clear too.
Powerfull `while_some` method !

cuviper:
Looks great, thanks!

bors r+

cuviper:
> error: unable to get packages from source
> 
> Caused by:
> 
>   [18] Transferred a partial file (transfer closed with 897629 bytes remaining to read)

Looks spurious.

bors r+

PR #468, opened 2017-11-06
cargo update

cuviper:
- docopt -> 0.8
  - rustc-serialize -> serde
- glium -> 0.18


cuviper:
Should fix (yet another) GL break on OSX...

bors r+

PR #471, opened 2017-11-14
bump coco to 0.3

ignatenkobrain:


cuviper:
If stjepang/coco#20 gets merged and published, we won't have to worry about bumping rayon-core's minimum rustc for this.

stjepang:
@cuviper Done! Merged and published.

ignatenkobrain:
@stjepang https://crates.io/crates/coco still doesn't list 0.3.2 ;)

stjepang:
Oops, sorry! Now it does. :)

cuviper:
Here are some cargo-chrono plots before and after:

rayon-pr471-points

rayon-pr471-points-medians

Mostly wins, but two big regressions:

- `fibonacci_split_recursive`: this is a kind of weird one, that I'm willing to write off.  FWIW, the related `fibonacci_join` benchmarks (not shown here) are about twice as fast now!
- `parallel_find_last`: the really weird part about this is that it's now performing even worse than `parallel_find_missing`, which itself stayed about the same.  Parallel finding the last element should be able to short-circuit and return sooner than with a missing element that has to check everything.  I'm really baffled by this one!

cuviper:
On a whim, I just tried compiling rayon-demo with all test modules except `find` commented out, and now `parallel_find_last` does perform better than `parallel_find_missing` as expected, and about the same as with the old coco.  But `parallel_find_middle` got worse.

My best guess is that these are just sensitive to changes in overall code generation and/or layout. Meh.

bors r+

PR #484, opened 2017-11-30
bump lazy_static to 1

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


cuviper:
Hmm, we're still indirectly getting `lazy_static 0.2` from dependencies, and `compiletest_rs` is too sensitive to having multiple copies of a library.

```
error[E0464]: multiple matching crates for `lazy_static`
 --> tests/run-pass/sort-panic-safe.rs:2:1
  |
2 | extern crate lazy_static;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: candidates:
          crate `lazy_static`: /home/travis/build/rayon-rs/rayon/target/debug/deps/liblazy_static-58ed3e94141b3559.rlib
          crate `lazy_static`: /home/travis/build/rayon-rs/rayon/target/debug/deps/liblazy_static-ccedce43d96797cc.rlib
```

Normally cargo passes specific library paths to rustc, so this isn't a problem, but `compiletest_rs` calls rustc directly and doesn't know which libraries should be used.  So if rustc is left with multiple options in the `-L` path, it fails.

I think we can just avoid `lazy_static` in that test though.  I'll try it...

cuviper:
Better yet, I just made it a regular test and added the proper dev-dependency.

bors r+

PR #488, opened 2017-12-08
Add Update adaptor and `update()` method

### LINKED ISSUE #487: Add an Itertools::update equivalent
https://docs.rs/itertools/0.7.4/itertools/trait.Itertools.html#method.update

It's basically an in-place `map`, or a mutable version of `inspect`.

Enet4:
This resolves #487. The sequential version created by the producer is based on the `itertools` implementation. Tests are included, but let me know if you'd like better coverage.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -67,6 +67,8 @@ mod interleave_shortest;
 pub use self::interleave_shortest::InterleaveShortest;
 mod intersperse;
 pub use self::intersperse::Intersperse;
+mod update;
+pub use self::update::Update;
 
 mod noop;
 mod rev;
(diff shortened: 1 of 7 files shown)


src/iter/update.rs
```diff
@@ -0,0 +1,295 @@
...
+    fn complete(self) -> C::Result {
+        self.base.complete()
+    }
+
+    fn full(&self) -> bool {
+        self.base.full()
+    }
+}
+
+/// Standard Update adaptor, based on `itertools::adaptors::Update`
```
cuviper: Did you consider using itertools directly?  We wouldn't want a public dependency, but this is behind the scenes.  But the implementation is pretty straightforward, so I'm just wondering what you think.

src/iter/mod.rs
```diff
@@ -1183,6 +1185,25 @@ pub trait ParallelIterator: Sized + Send {
...
+    /// ```
+    /// use rayon::prelude::*;
+    ///
+    /// let mut par_iter = (0..5).into_par_iter().update(|mut x| {*x *= 2;});
+    ///
+    /// let doubles: Vec<_> = par_iter.collect();
+    ///
+    /// assert_eq!(&doubles[..], &[0, 2, 4, 6, 8]);
+    /// ```
+    fn update<F>(self, update_op: F) -> Update<Self, F>
```
cuviper: Also, just to keep related behavior grouped together, let's move this up after `inspect`.

src/iter/update.rs
Enet4: I did think about that while reading the issue. I chose to replicate the implementation because `interleave` did the same, although with additional reasons.

If there is willingness to depend on `itertools` in future versions, I suppose this adaptor would take advantage of that. Although it's just my opinion, leaving these adaptors  behind a feature gate to keep `itertools` an optional dependency feels a bit unnecessary.

Enet4:
The requested adjustments were made. Please let me know if you'd like a commit squash.

As for the actual use of `itertools`, if you find that keeping it as a dependency is ok, I can make the necessary changes.

cuviper:
Thanks!  We don't need to worry much about itertools -- we can always change this later if wanted.

bors r+

PR #498, opened 2017-12-21
FromParallelIterator and ParallelExtend Cow for String

cuviper:
Parallel version of rust-lang/rust#41449.


cuviper:
>  Ran for 8 hrs 5 min 37 sec

That Travis macos queue is painful - maybe we should trim back to just one macos build?

But it did succeed, so...

PR #499, opened 2017-12-21
rayon-futures: update the error location in a compile-fail test

cuviper:


cuviper:
Argh, just one build away!!!  That macos queue is painful...

bors r+

PR #502, opened 2017-12-31
Bump rand to 0.4

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


cuviper:
It looks like rand 0.4 requires rustc 1.15.  Since it requires no code change, can we make this dependency `>= 0.3, < 0.5`? Then we don't have to bump the minimum rustc.  The CI scripts will have to explicitly downgrade for testing the older rustc though.

ignatenkobrain:
@cuviper done.

Bilkow:
I believe it's better to change it in rayon-demo too for consistency.

ignatenkobrain:
@Bilkow I think rayon-demo is ancient....

cuviper:
Sadly, rust-lang-nursery/rand#239 bumped the minimum rustc all the way to **1.22**!  And rand-0.3.22 is now re-exporting items from 0.4, so any hope for older rustc has to downgrade to 0.3.20.

I'll go ahead and add that to this PR so we can merge, because it's now breaking CI in other PRs.

cuviper:
I acknowledge this is ugly, but still...

bors r+

cuviper:
Oh, hang on, need to figure out appveyor CI too... ugh

bors r-

cuviper:
Meh, we'll just be lazier about 1.12, testing Linux only.

bors r+

cuviper:
In the very last Windows target:

```
---- [compile-fail] compile-fail\future_escape.rs stdout ----
error: tests/compile-fail/future_escape.rs:2: unexpected error: '2:1: 2:20: multiple matching crates for `rayon` [E0464]'
```

This is a common problem with compiletest-rs, but usually we just avoid compiling the same crate multiple times.  But for some reason the build does show that it compiled `rayon` more than once.  No idea why, especially only on this one target.  Let's try again.

bors r+

cuviper:
Same failure... 😠

PR #503, opened 2017-12-31
Adapt for wasm32 target

Bilkow:
Changed from libc::abort() to std::process::abort() in unwind.rs (rayon-core) so it can be compiled for wasm32-unknown-unknown. As a consequence, both an unsafe block and the libc crate dependency for rayon-core can be removed.

Version bump in rand to 0.4, as 0.3.19 doesn't compile for wasm. It seems I'm late for the version bump (#502), so I'm not sure if I should remove it from the commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,16 @@ matrix:
       env: RUSTFLAGS='--cfg rayon_unstable'
       if: NOT type = pull_request
 
+    # wasm won't actually work without threading, but it builds
+    - rust: nightly
+      os: linux
+      env: TARGET=wasm32-unknown-unknown
+      script:
+        - rustup target add $TARGET
+        - cargo build --target $TARGET
+      if: NOT type = pull_request
+
+
 script:
   - cargo build
   - |
--- a/rayon-core/src/unwind.rs
+++ b/rayon-core/src/unwind.rs
@@ -2,10 +2,6 @@
 //! place, you can use the `AbortIfPanic` helper to protect against
 //! accidental panics in the rayon code itself.
 
-#[cfg(any(not(target_arch = "wasm32"), target_os = "emscripten"))]
-use libc;
-#[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
-use std::process;
 use std::any::Any;
 use std::panic::{self, AssertUnwindSafe};
 use std::io::stderr;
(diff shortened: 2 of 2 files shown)


cuviper:
AIUI, wasm doesn't have threads at all, does it? Can you actually use rayon with these changes?

That `abort` was only added in rust 1.17, and I'd like to avoid increasing the minimum if it's not too much trouble. Can you make that change only for `#[cfg(target = "wasm")]`? (that's probably not exactly how to write it.)

Let's indeed keep rand separate, since that PR already exists.

Bilkow:
Thanks for your fast answer!
I believe it doesn't have threads, but I'm trying to compile "image", that depends on jpeg-decoder that depends on rayon. I'm not sure how will threads work in wasm, if they'll just block, for instance. 
With those changes my crate compiles with the `wasm32-unknown-unknown` target.
I'll remove the rand bump and make it only for wasm32 then.

cuviper:
https://github.com/rust-lang/rust/blob/master/src/libstd/sys/wasm/thread.rs

It looks like wasm threads are stubbed out to an "unsupported" error. `Threadpool::new()` should propagate that, or else using rayon's global threadpool will panic when it unwraps the error.

The rayon use in `jpeg-decoder` is optional, but it doesn't look like `image` offers a way to control that feature. You could disable jpeg support altogether if you don't need it.

Bilkow:
Oh, I see! I'll probably just disable jpeg support for now and patch image to reexport the rayon feature of jpeg-decoder later.

While I think that thread support is coming to wasm32 soon, I'm not sure if you'd rather have a panic instead of a compile time error for now.

Should I continue?

cuviper:
It's up to you, if you want to keep working on it now. I wouldn't claim to support wasm until we can actually use it, but we can merge preliminary fixes at any time.

Bilkow:
Oops, I accidentally closed it by undoing my commit.
I did it only for wasm32, excluding emscripten and (with the rand bump) it's compiled both for wasm32 and x86_64 linux.
`cargo +nightly test` will sometimes fail with the errors I described in #504, but it also happens in the main repo so I believe it isn't related.

Diggsey:
I just ran into this too. Can this be merged?

cuviper:
I rebased and then moved the `cfg` for 1.12 compat.  Let's see if it passes...

bors r+

PR #506, opened 2018-01-13
`take` is a function on IndexedParallelIterator

phimuemue:
So, I updated the links to the proper positions.


cuviper:
Good catch, thanks!  I also updated `skip` and `enumerate` that had the same problem.

bors r+

Issue #465, opened 2017-10-24
Look for more opportunities to `fold_with` and `consume_iter`


cuviper:
These functions have reasonable default implementations, but sometimes they can be improved.

- The default [`Producer::fold_with`](https://github.com/rayon-rs/rayon/blob/cdf876d903e948ba5be440c54eac0ca3c30c0b75/src/iter/internal.rs#L46) just converts to an iterator and passes that to the `Folder`.
  - [`ChainProducer::fold_with`](https://github.com/rayon-rs/rayon/blob/cdf876d903e948ba5be440c54eac0ca3c30c0b75/src/iter/chain.rs#L199) does better by folding its two parts separately, which can reduce conditional branching.  (This is much like the motivation for the custom `std::iter::Chain::fold`.)
  - Most of the other producers that wrap a base type could probably have a custom `fold_with` too, if only to allow a chain base type to optimize itself as before.
    - e.g. `ClonedProducer::fold_with` could call `self.base.fold_with(ClonedFolder { base: folder })`
- The default [`Folder::consume_iter`](https://github.com/rayon-rs/rayon/blob/cdf876d903e948ba5be440c54eac0ca3c30c0b75/src/iter/internal.rs#L81) just picks off items one by one to `consume()` them until `full()`.  (If it weren't for that "full" short-circuiting, it could just call `Iterator::fold` instead.)
  - [`SumFolder::consume_iter`](https://github.com/rayon-rs/rayon/blob/cdf876d903e948ba5be440c54eac0ca3c30c0b75/src/iter/sum.rs#L86) just calls the standard `Iterator::sum()`. (which actually is a fold()... here we know it will never be full.)
  - There may be other folders which can do a better job in `consume_iter` too.

Issue #352, opened 2017-05-31
should `skip` preserve side-effects when running in parallel?


nikomatsakis:
In the code for `skip()`, I noticed [this call to `bridge_producer_consumer()`](https://github.com/nikomatsakis/rayon/blob/40e7efa8cd51464e9c8511ccfad21713a5d411b4/src/iter/skip.rs#L74):

```rust
let (before_skip, after_skip) = base.split_at(self.n);
bridge_producer_consumer(self.n, before_skip, NoopConsumer::new());
self.callback.callback(after_skip)
```

It seems odd to me that we do this, rather than just throw away `before_skip`. In particular, I do think that consumers can expect to be "fed" all the items that a parallel iterator promised (i.e., what it returned for `len()` or `opt_len()`). But I don't think that *producers* should necessarily expect to be "driven" by having all of their items pulled (and, in this case, discarded).

I'm inclined to drop that call to `bridge_producer_consumer()`. Thoughts?

cuviper:
That was to maintain side-effect symmetry with `Iterator::skip` : https://github.com/nikomatsakis/rayon/pull/162#issuecomment-261684136

If you want to weaken that symmetry, then sure, we could throw the first half away, just like `take` throws away its second half.  (IOW `take` already proves that producers may be incompletely driven.)

cuviper:
For that matter, we also have many short-circuiting ops like `find_any`.  So I don't think there's any general expectation that producers will fully produce -- `skip` just needs additional consideration for side effects.

nikomatsakis:
ah, yeah, forgot about the side-effects

nikomatsakis:
I changed the title since I think we agree that producers are not forced to produce.

nikomatsakis:
We decided to keep compatibility with sequential iterators. If we want a version that ignores side-effects, we can always add another method later with a different name.

nikomatsakis:
This came up in a conversation recently -- there is some need for a method that, given an `IndexedParallelIterator`, lets you extract a subset. The person was using `skip` and `take` in conjunction to do that, but it's imperfect (particularly when it comes to side-effects). It seems like we could offer a more direct method.

cuviper:
Maybe add something like `trim(start, end)` that we'll say explicitly doesn't run side-effects?

nikomatsakis:
Discussed in meeting:

skip will retain its sequential-compatible semantics.

[...]

PR #508, opened 2018-01-13
Specialize some Folder::consume_iter methods

MaloJaffre:
Fixes the second item of #465.

### DIFF
--- a/src/iter/for_each.rs
+++ b/src/iter/for_each.rs
@@ -45,6 +45,11 @@ impl<'f, F, T> Folder<T> for ForEachConsumer<'f, F>
         self
     }
 
+    fn consume_iter<I>(self, iter: I) -> Self where I: IntoIterator<Item=T> {
+        iter.into_iter().fold((), |_, item| (self.op)(item));
+        self
+    }
+
     fn complete(self) {}
 
     fn full(&self) -> bool {
--- a/src/iter/noop.rs
+++ b/src/iter/noop.rs
@@ -33,6 +33,11 @@ impl<T> Folder<T> for NoopConsumer {
         self
     }
 
+   fn consume_iter<I>(self, iter: I) -> Self where I: IntoIterator<Item=T> {
+        iter.into_iter().fold((), |_, _| ());
+        self
+   }
+
     fn complete(self) {}
 
     fn full(&self) -> bool {


src/iter/find.rs
```diff
@@ -84,6 +84,14 @@ impl<'p, T, P> Folder<T> for FindFolder<'p, T, P>
         self
     }
 
+    fn consume_iter<I>(mut self, iter: I) -> Self where I: IntoIterator<Item=T> {
```
nikomatsakis: Hmm. I'm not sure about this method. Two things:

- First, I don't know if it will be faster than the default impl. That default impl already just iterates and invokes `consume` until `full()` is true -- this is basically the same thing we are doing here.
- Second, in this version, we are not checking for the possibility that `self.found` gets set to true by somebody else, so we can wind up doing quite a bit more work. 

To fix the second part, you already can't use the built-in iterator `find` operation, since we want to be checking for `self.found` being set to true as well. Once we make those changes, I'm even less sure this will pay off. I think i'd be inclined *not* to include `consume_iter` here (and perhaps to leave a comment, instead, explaining that it likely doesn't pay off here).

Or am I missing something? (cc @cuviper -- thoughts?)

src/iter/for_each.rs
```diff
@@ -45,6 +45,13 @@ impl<'f, F, T> Folder<T> for ForEachConsumer<'f, F>
         self
     }
 
+    fn consume_iter<I>(self, iter: I) -> Self where I: IntoIterator<Item=T> {
+        for item in iter {
```
nikomatsakis: Here, I guess we could do `iter.into_iter().for_each(self.op)` -- this gives the iterator the maximal chance to optimize, I believe. (I'm not sure if we would get a win, as I'd expect the optimizer to handle stripping out the calls to `full` anyway, but can't hurt.)

src/iter/noop.rs
```diff
@@ -33,6 +33,11 @@ impl<T> Folder<T> for NoopConsumer {
         self
     }
 
+   fn consume_iter<I>(self, iter: I) -> Self where I: IntoIterator<Item=T> {
+        for _ in iter {}
```
nikomatsakis: Man, seeing this silliness really makes me revisit the [question of side-effects](https://github.com/rayon-rs/rayon/issues/352). It'd be so nice to just "do nothing" here. In any case, I think I would prefer `iter.for_each(|_| ())`.

src/iter/find.rs
cuviper: I agree, this seems like a case that shouldn't be changed.  We want `find` to short-circuit.

src/iter/for_each.rs
cuviper: It would have to be `fold`, until we bump the minimum rustc to reach `Iterator::for_each`.

src/iter/noop.rs
cuviper: Ditto on using `fold` instead.

src/iter/for_each.rs
nikomatsakis: or `.all(|x| { self.op(x); true })`, I suppose.

MaloJaffre:
Thanks @nikomatsakis and @cuviper, I've rebased and fixed my PR according to your review.

cuviper:
bors r=nikomatsakis

PR #509, opened 2018-01-18
Add chunks method to IndexedParallelIterator

### LINKED ISSUE #501: Implement `chunks`
Right now if I have a slice I can do:

```rust
slice.par_chunks(500)....
```

but if I have an iterator I can't chunk it:

```rust
(0..100000).par_chunks(500)... // Error
(0..100000).into_par_iter().chunks(500).... // Error
```

So I end up needing to work around this as:

```rust
let indices: Vec<_> = (0..100000).collect(); // :(
indices.par_chunks(500)... // Works
```

which is horrible.

tmccombs:
Fixes #501 

Note: I am only using a Vec<T> for each chunk because if I defined a Chunk iterator like itertools does, it would have to reference the type of `Producer::IntoIter` for the producer given to the ProducerCallback, which isn't accessible from the IndexedParallelIterator type. And the Chunk type needs to be the Item of the Chunks iterator.  An alternative implementation could possibly use a Box<Iterator> instead of using generics.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -30,6 +30,8 @@ mod find;
 mod find_first_last;
 mod chain;
 pub use self::chain::Chain;
+mod chunks;
+pub use self::chunks::Chunks;
 mod collect;
 mod enumerate;
 pub use self::enumerate::Enumerate;
@@ -1350,6 +1352,25 @@ pub trait IndexedParallelIterator: ParallelIterator {
               I::Iter: IndexedParallelIterator<Item = Self::Item>
     {
         interleave_shortest::new(self, other.into_par_iter())
+    }
+
+    /// Split an iterator up into fixed-size chunks.
+    ///
+    /// Returns an iterator that returns `Vec`s of the given number of elements.
+    /// If the number of elements in the iterator is not divisible by `chunk_size`,
+    /// the last chunk may be shorter than `chunk_size`.
+    ///
+    /// Example:
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    /// let a = vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+    /// let r: Vec<Vec<i32>> = a.into_par_iter().chunks(3).collect();
+    /// assert_eq!(r, vec![vec![1,2,3], vec![4,5,6], vec![7,8,9], vec![10]]);
+    /// ```
+    fn chunks(self, chunk_size: usize) -> Chunks<Self> {
+        assert_ne!(chunk_size, 0, "chunk_size must not be zero");
+        chunks::new(self, chunk_size)
     }
 
     /// Lexicographically compares the elements of this `ParallelIterator` with those of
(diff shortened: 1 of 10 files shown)


src/iter/mod.rs
```diff
@@ -1352,6 +1354,25 @@ pub trait IndexedParallelIterator: ParallelIterator {
...
+    /// Example:
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    /// let a = vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+    /// let r: Vec<Vec<i32>> = a.into_par_iter().chunks(3).collect();
+    /// assert_eq!(r, vec![vec![1,2,3], vec![4,5,6], vec![7,8,9], vec![10]]);
+    /// ```
+    fn chunks(self, chunk_size: usize) -> Chunks<Self> {
+        assert_ne!(chunk_size, 0, "chunk_size must not be zero");
```
cuviper: CI failed Rust 1.12 here -- `assert!(chunk_size != 0, "...")` should be fine.
(We might want to add this assertion on `slice::Chunks` and `ChunksMut` too.)

tmccombs:
> If you could sketch out your alternative more, we can consider that.

My idea was basically instead of returning a Vec it would return an object that looks something like:

```rust
struct Chunk<T> {
  inner: Box<Iterator<Item=T>>
}
```

I don't think it would be too difficult to implement something like this.

cuviper:
I was *going* to point out that you can't really pass ownership of the iterator into `Chunks`, boxed or not, but your commit to "Refactor chunks to make use of wrapped producers split_at" is clever!  This looks like a promising approach to me.

src/iter/mod.rs
```diff
@@ -1352,6 +1354,25 @@ pub trait IndexedParallelIterator: ParallelIterator {
         interleave_shortest::new(self, other.into_par_iter())
     }
 
+    /// Split an iterator up into fixed-size chunks.
+    ///
+    /// Returns an iterator that returns `Vec`s of the given number of elements.
+    /// If the number of elements in the iterator is not divisible by `chunk_size`,
+    /// the last chunk may be shorter than `chunk_size`.
+    ///
+    /// Example:
```
cuviper: For consistency, let's make this `# Examples` - a header, and plural even though it's just one.

tmccombs:
I added the tests you requested.

As for using a boxed iterator. I ran into an issue. The Item of ParallelIterator must be Send, but the IntoIter of a producer doesn't have to be send. I'm going to leave it as returning a Vec for now.

src/math.rs
```diff
@@ -0,0 +1,24 @@
+/// Divide `n` by `divisor`, and round up to the nearest integer
+/// if not evenly divisable.
+#[inline]
+pub fn div_round_up(n: usize, divisor: usize) -> usize {
+    if n == 0 {
+        0
```
cuviper: It's a _little_ bit sketchy to have `div_round_up(0, 0) -> 0`, but the places we're using it will never have a 0 `divisor`.  Maybe just add a `debug_assert!(divisor != 0)`?

cuviper:
Oh, also this will need an update for #514.

nikomatsakis:
r+ from me =)

cuviper:
I pushed an addition to satisfy `#![deny(missing_docs)]`, and also cross-referenced `par_chunks`.

bors r+

cuviper:
Fixed for #521.

bors r+

Issue #420, opened 2017-09-06
doc: Documentation examples for parallel iterator methods


alexcrichton:
Ideally all methods on the parallel iterator traits should have an example of how they can be used!

cuviper:
I'm reopening this per https://github.com/rayon-rs/rayon/pull/466#issuecomment-342289823

nikomatsakis:
Assigning @cuviper to do a sweep over these methods.

cuviper:
> Ideally all methods on the parallel iterator traits should have an example of how they can be used!

"traits" plural -- I'm inclined to include all traits in the prelude, but we're missing a lot of examples...

- [ ] `ParallelIterator`
- [ ] `FromParallelIterator`
- [ ] `IntoParallelIterator`
- [ ] `IntoParallelRefIterator`
- [ ] `IntoParallelRefMutIterator`
- [ ] `IndexedParallelIterator`
- [ ] `ParallelExtend`
- [ ] `ParallelIterator` will be finished by #511 
- [ ] `ParallelSlice`
- [ ] `ParallelSliceMut`
- [ ] `ParallelString`

(#512 does add documentation to some of these, but not examples.)

Also, the iter functions are in good shape except for `iter::split()`.

cuviper:
All checked!

PR #511, opened 2018-01-20
420 document parallel iterator methods

oddg:
Contribution for closing https://github.com/rayon-rs/rayon/issues/420.

There is one method left in the `ParallelIterator` trait that has no example: `opt_len`.
Also the example for `fold_with` might not be necessary.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -275,10 +275,16 @@ pub trait ParallelIterator: Sized + Send {
     ///
     /// let (sender, receiver) = channel();
     ///
-    /// let mut par_iter = (0..5).into_par_iter().map_with(sender, |s, x| { s.send(x).unwrap(); x });
-    /// let a: Vec<_> = par_iter.collect();
+    /// let a: Vec<_> = (0..5)
+    ///                 .into_par_iter()            // iterating over i32
+    ///                 .map_with(sender, |s, x| {
+    ///                     s.send(x).unwrap();     // sending i32 values through the channel
+    ///                     x                       // returning i32
+    ///                 })
+    ///                 .collect();                 // collecting the returned values into a vector
     ///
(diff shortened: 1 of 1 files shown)


src/iter/mod.rs
```diff
@@ -266,6 +266,23 @@ pub trait ParallelIterator: Sized + Send {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::sync::mpsc::channel;
+    /// use rayon::prelude::*;
+    ///
+    /// let (sender, receiver) = channel();
+    ///
+    /// let mut par_iter = (0..5).into_par_iter().map_with(sender, |s, x| { s.send(x).unwrap(); x });
```
nikomatsakis: Nit: can we format this a bit more clearly? Maybe run the example through rustfmt (on play) and paste it back? I feel like it's a bit hard to follow what's happening. (Also, some comments explaining what's going on wouldn't hurt.)

src/iter/mod.rs
```diff
@@ -484,6 +501,18 @@ pub trait ParallelIterator: Sized + Send {
     /// efficient. If possible, it is better to call `reduce()`, which
     /// requires an identity element.
     ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    /// let sums = [(0, 1), (5, 6), (16, 2), (8, 9)]
+    ///            .par_iter()        // iterating over &(i32, i32)
+    ///            .cloned()          // iterating over (i32, i32)
```
nikomatsakis: formatting like this is great =)

oddg:
@nikomatsakis Thanks for your review.
I hope the example is clearer now.

cuviper:
bors r=nikomatsakis

Issue #421, opened 2017-09-06
doc: Modules should have documentation


alexcrichton:
Ideally all modules in the crate should have documentation as to their purpose and maybe even an example or two if appropriate

nikomatsakis:
Most of them have comments but they are perhaps a bit lacking. 

Major exception:

The threadpool docs might be an area where we could use help.

PR #515, opened 2018-01-26
document spawn return value, modules

### LINKED ISSUE #431: rayon::spawn should document the lack of return value
This seemed to be at least very similar to `std::thread::spawn`, but it may be worth documenting why a return value isn't exposed here. (and recommend using the rayon-futures crate if that's what you need)

nikomatsakis:
Document the spawn return vaule and add doc-comments to some of the modules. Also cleanup the style of doc-comments to be a bit more consistent -- e.g., don't name the function being documented in the first sentence, which reads a bit oddly in rustdoc.

Fixes #431 
Fixes https://github.com/rayon-rs/rayon/issues/421

### DIFF
--- a/rayon-core/src/join/mod.rs
+++ b/rayon-core/src/join/mod.rs
@@ -10,8 +10,8 @@ use FnContext;
 #[cfg(test)]
 mod test;
 
-/// The `join` function takes two closures and *potentially* runs them
-/// in parallel. It returns a pair of the results from those closures.
+/// Takes two closures and *potentially* runs them in parallel. It
+/// returns a pair of the results from those closures.
 ///
 /// Conceptually, calling `join()` is similar to spawning two threads,
 /// one executing each of the two closures. However, the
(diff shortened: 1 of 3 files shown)


rayon-core/src/join/mod.rs
```diff
@@ -58,12 +58,13 @@ pub fn join<A, B, RA, RB>(oper_a: A, oper_b: B) -> (RA, RB)
     join_context(|_| oper_a(), |_| oper_b())
 }
 
-/// The `join_context` function is identical to `join`, except the closures
-/// have a parameter that provides context for the way the closure has been
-/// called, especially indicating whether they're executing on a different
-/// thread than where `join_context` was called.  This will occur if the second
-/// job is stolen by a different thread, or if `join_context` was called from
-/// outside the thread pool to begin with.
+/// Indentical to `join`, except that the closures have a parameter
```
KodrAus: Identical?

cuviper: https://www.urbandictionary.com/define.php?term=indentical

cuviper: I just pushed the correction.

nikomatsakis: sigh :)

Issue #417, opened 2017-09-06
Move to concrete Send/Sync error types


alexcrichton:
One of the API guidelines is that errors should generally be `Send` and `Sync`, but it looks like `Box<Error>` is being returned in a few locations? Could a concrete error be returned instead?

cuviper:
The places we currently use `Box<Error>` are in rayon-core, which we'd rather not bump semver.

Would it be a breaking change to make those `Box<Error + Send + Sync>`?  I think that may be OK -- at least this example does compile:

```rust
use std::error::Error;
fn foo() -> Box<Error + Send + Sync> {
    unimplemented!()
}
fn main() {
    let x: Box<Error> = foo();
}
```

alexcrichton:
I think technically, yes, but practically, most likely no? There's a few options here though to do if a rayon-core bump is off the table. Most of them involve not directly reexporting rayon-core but rather wrapping it, but one example is that you could add a concrete, cloneable type in rayon-core. The rayon crate would then downcast to that type, clone it, and then return it (maybe wrapped again)

nikomatsakis:
Hmm, the only place I see that we are using `Box<Error>` is the `initialize` and `ThreadPool::new` methods, right?

Maybe we can just deprecate those and create new names for them? Then we can use some concrete types or something? I'd like to be "failure compatible" I guess. Ultimately this doesn't seem like a super big deal though, those are pretty peripheral methods.

I do highly doubt that changing to `Box<Error + Send + Sync>` would break anyone. But I guess you never know.

cuviper:
`initialize()`, `ThreadPool::new()`, and then `Configuration::build()` returns `ThreadPool::new()`.  One concrete type can cover all of these.

nikomatsakis:
Decision:

`Configuration::build`, `ThreadPool::new`, and `initialize` would be deprecated. 

Add:

- `Configuration::init(self) -> ThreadPool`
- `Configuration::init_global(self)`

Alternative names:

- `Configuration::into_thread_pool(self) -> ThreadPool`
- `Configuration::into_global_thread_pool(self)`

nikomatsakis:
Happy to bikeshed names further =)

cuviper:
Those new methods should return `Result` with concrete error types.
(i.e. the whole point of this issue...)

cuviper:
Fixed by #516.

PR #516, opened 2018-01-26
Introduce concrete error type

jwass:
Addresses #417. Still a work in progress.
Remaining questions:
- [x] * Documentation on new error type
- [x] * Should we remove `pub use rayon_core::initialize` in `rayon/src/lib.rs`
- [x] * Finalize name decision for `ThreadPoolInitError` and `Configuration::init()`

Define ThreadPoolInitError. Deprecate methods that returned
Box<Error> types in favor of new methods returning the concrete
ThreadPoolInitError.

Deprecations in rayon-core:
initialize() => Configuration::init_global()
Configuration::build() => Configuration::init()
ThreadPool::new() => Configuration::init()

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ equals to the number of logical cores and not the physical ones.
 
 If you want to alter the number of threads spawned, you can set the
 environmental variable `RAYON_NUM_THREADS` to the desired number of threads
-or use the [`initialize` function](https://docs.rs/rayon/*/rayon/fn.initialize.html) method
+or use the [`ThreadPoolBuilder::build_global` function](https://docs.rs/rayon/*/rayon/struct.ThreadPoolBuilder.html#method.build_global) method
 
 ### Parallel Iterators
 
--- a/rayon-core/src/join/test.rs
+++ b/rayon-core/src/join/test.rs
@@ -1,9 +1,8 @@
 //! Tests for the join code.
 
-use Configuration;
+use ThreadPoolBuilder;
 use join::*;
 use rand::{Rng, SeedableRng, XorShiftRng};
-use thread_pool::*;
 use unwind;
 
 fn quick_sort<T: PartialOrd + Send>(v: &mut [T]) {
@@ -44,7 +43,7 @@ fn sort_in_pool() {
     let mut rng = XorShiftRng::from_seed([0, 1, 2, 3]);
     let mut data: Vec<_> = (0..12 * 1024).map(|_| rng.next_u32()).collect();
 
-    let pool = ThreadPool::new(Configuration::new()).unwrap();
+    let pool = ThreadPoolBuilder::new().build().unwrap();
     let mut sorted_data = data.clone();
     sorted_data.sort();
     pool.install(|| quick_sort(&mut data));
@@ -89,7 +88,7 @@ fn join_context_both() {
 #[test]
 fn join_context_neither() {
     // If we're already in a 1-thread pool, neither job should be stolen.
-    let pool = ThreadPool::new(Configuration::new().num_threads(1)).unwrap();
+    let pool = ThreadPoolBuilder::new().num_threads(1).build().unwrap();
     let (a_migrated, b_migrated) = pool.install(|| {
         join_context(|a| a.migrated(), |b| b.migrated())
     });
@@ -103,7 +102,7 @@ fn join_context_second() {
 
     // If we're already in a 2-thread pool, the second job should be stolen.
     let barrier = Barrier::new(2);
-    let pool = ThreadPool::new(Configuration::new().num_threads(2)).unwrap();
+    let pool = ThreadPoolBuilder::new().num_threads(2).build().unwrap();
     let (a_migrated, b_migrated) = pool.install(|| {
         join_context(|a| { barrier.wait(); a.migrated() },
                      |b| { barrier.wait(); b.migrated() })
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -26,6 +26,7 @@
 
 use std::any::Any;
 use std::env;
+use std::io;
 use std::error::Error;
 use std::marker::PhantomData;
 use std::str::FromStr;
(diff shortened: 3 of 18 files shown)


rayon-core/src/join/test.rs
```diff
@@ -44,7 +43,7 @@ fn sort_in_pool() {
     let mut rng = XorShiftRng::from_seed([0, 1, 2, 3]);
     let mut data: Vec<_> = (0..12 * 1024).map(|_| rng.next_u32()).collect();
 
-    let pool = ThreadPool::new(Configuration::new()).unwrap();
+    let pool = Configuration::new().init().unwrap();
```
cuviper: (bike-shedding) It doesn't seem great that this call says nothing about being a thread pool -- though this was true of `::build()` too.  Maybe it's OK for folks who would call it like `rayon::Configuration::new().init()` so the context isn't entirely lost.

rayon-core/src/lib.rs
```diff
@@ -84,6 +85,18 @@ pub fn current_num_threads() -> usize {
     ::registry::Registry::current_num_threads()
 }
 
+#[derive(Debug)]
+pub struct ThreadPoolInitError {
+    kind: ErrorKind,
+
```
cuviper: rogue blank line

rayon-core/src/lib.rs
```diff
@@ -306,27 +347,30 @@ impl Configuration {
...
+        match self.kind {
+            ErrorKind::GlobalPoolAlreadyInitialized => "The global thread pool has already been initialized.",
+            ErrorKind::IOError(ref e) => e.description(),
+        }
+    }
+}
+
+impl fmt::Display for ThreadPoolInitError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.description().fmt(f)
```
cuviper: `io::Error` is more informative in `Display`, so it would be nice to forward that directly.

jwass: Like this? https://github.com/rayon-rs/rayon/pull/516/files#diff-f42c354e8416b0ad30f65243a93d84a0R369

rayon-core/src/test.rs
```diff
@@ -118,3 +117,19 @@ fn handler_panics_handled_correctly() {
...
+
+#[test]
+#[allow(deprecated)]
+fn check_config_build() {
+    let pool = Configuration::new().num_threads(22).build().unwrap();
+    assert_eq!(pool.current_num_threads(), 22);
+}
+
+
+/// Helper used by check_error_send_sync to ensure ThreadPoolInitError is Send + Sync
```
jwass: Is this the right way to write a test to ensure the type is Send + Sync?

jwass:
@cuviper Addressed all the requested changes + a few other things - with the exception of coming up with a final name for `Configuration::init`. I figure we can decide on that once the rest of the code is good to go.

cuviper:
Another option we were discussing on gitter is to rename `Configuration` to something like `ThreadPoolBuilder`.  (Or really a "soft" rename that still leaves a deprecated `Configuration` behind.)  Then I think we'll have better context for what a `build()`/`init()` method is doing.

jwass:
@cuviper @nikomatsakis 
Here's a first pass with the new `ThreadPoolBuilder` with `Configuration` delegating to it.

* Rename `Configuration` to `ThreadPoolBuilder` and add `build(self) -> Result<ThreadPool, ThreadPoolInitError>`
* Make a deprecated `Configuration` that's a thin wrapper around `ThreadPoolBuilder` while maintaining its original interface.

Still left to do:
* [ ] Update any lingering `Configuration` references in the documentation.
* [ ] Add some tests to ensure the deprecated `Configuration` is still fully supported.

If you agree with the approach I'll finish out the rest.

cuviper:
> Add some tests to ensure the deprecated Configuration is still fully supported.

I've had a vague idea that we ought to be testing past eras of `rayon`.  We could create some more local repos under `ci/` that require old `rayon` versions, but override their `rayon-core` to the local path, then make sure they at least still build.  (We don't need to make this idea part of your PR though.)

rayon-core/src/lib.rs
```diff
@@ -128,15 +148,38 @@ type StartHandler = Fn(usize) + Send + Sync;
...
+    ///   yield slightly more consistent results, since the worker threads
+    ///   will already be ready to go even in the first iteration.  But
+    ///   this cost is minimal.
+    ///
+    /// Initialization of the global thread pool happens exactly
+    /// once. Once started, the configuration cannot be
+    /// changed. Therefore, if you call `init_global` a second time, it
+    /// will return an error. An `Ok` result indicates that this
+    /// is the first initialization of the thread pool.
+    pub fn init_global(self) -> Result<(), ThreadPoolInitError> {
```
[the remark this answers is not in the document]
jwass: That was exactly my thinking to leaving it as `init_global` - but I have no objections to `build_global` or anything else if that's what makes sense. I'll defer to your judgment.

README.md
```diff
@@ -93,7 +93,7 @@ equals to the number of logical cores and not the physical ones.
 
 If you want to alter the number of threads spawned, you can set the
 environmental variable `RAYON_NUM_THREADS` to the desired number of threads
-or use the [`initialize` function](https://docs.rs/rayon/*/rayon/fn.initialize.html) method
+or use the [`Configuration::init_global` function](https://docs.rs/rayon/*/rayon/struct.Configuration.html#method.init_global) method
```
nikomatsakis: Nit: it's not `ThreadPoolBuilder::init_global`, right?

rayon-core/src/lib.rs
```diff
@@ -84,9 +85,21 @@ pub fn current_num_threads() -> usize {
...
+    kind: ErrorKind,
+}
+
+#[derive(Debug)]
+enum ErrorKind {
+    GlobalPoolAlreadyInitialized,
+    IOError(io::Error),
+}
+
 /// Contains the rayon thread pool configuration.
```
nikomatsakis: I think we could do a bit better here in this documentation. =) Maybe something like this:

---

Used to configure and create a new thread-pool, or to configure the global rayon thread-pool. 

# Examples

The following creates a Rayon thread-pool using 22 threads:

```rust
let pool = ThreadPoolBuilder::new().num_threads(22).build();
```

You can also configure the global thread pool by using `init_global` instead of `build`:

```rust
ThreadPoolBuilder::new().num_threads(22).init_global();
```

jwass:
@nikomatsakis I think I addressed everything:
* Renamed `init_global` to `build_global`
* Updated all `Configuration` references, both in code and docs, to use `ThreadPoolBuilder`. (At least I hope I got them all.)
* Added a small test to ensure that using all public methods on `Configuration` at least compile, now that it's deprecated. We can do something more robust like @cuviper's suggestion above but maybe a different PR.
* Fleshed out docs for `ThreadPoolBuilder`.
* Updated `rayon-futures` to use `ThreadPoolBuilder`.
* No compiler warnings.

I'd like to rebase/squash before merging.

cuviper:
Looks great to me!

jwass:
Rebased/squashed and ready to go.

PR #517, opened 2018-01-28
Add Examples to Documentation

iwillspeak:
Updates the docs for the `ParallelString` and `ParallelSlice` trait to add
examples for each method.

part of #420

### DIFF
--- a/src/str.rs
+++ b/src/str.rs
@@ -50,6 +50,14 @@ pub trait ParallelString {
     fn as_parallel_string(&self) -> &str;
 
     /// Returns a parallel iterator over the characters of a string.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    /// let max = "hello".par_chars().max_by_key(|c| *c as i32);
+    /// assert_eq!(Some('o'), max);
+    /// ```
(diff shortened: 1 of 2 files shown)


src/str.rs
```diff
@@ -70,6 +89,16 @@ pub trait ParallelString {
...
     /// It is implemented for `char` and any `F: Fn(char) -> bool + Sync + Send`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    /// let parts: Vec<_> = "((1 + 3) * 2)"
+    ///     .par_split_terminator(|c| c == '(' || c == ')')
+    ///     .collect();
+    /// assert_eq!(vec!["", "", "1 + 3", " * 2"], parts);
```
cuviper: I have to say, this is a peculiar example!  I do like that it shows a somewhat complicated pattern though.

cuviper:
Thanks!

bors r+

src/str.rs
iwillspeak: Yeah, I think by then I was kind of fed up of splitting simpler things! I also wanted to show how it would handle the trailing bracket differently to the leading ones.

PR #525, opened 2018-02-06
revise the "overall documentation strategy"

nikomatsakis:
The iter module and join/scope functions now host the primary docs. The README and crate-level docs mostly forward to those. Some of the detailed material from the README is moved to a FAQ and lightly edited.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,0 +1,227 @@
+# Rayon FAQ
+
+These file is for general questions that don't fit into the README or
+crate docs.
+
+## How many threads will Rayon spawn?
+
+By default, Rayon uses the same number of threads as the number of CPUs
+available. Note that on systems with hyperthreading enabled this
+equals to the number of logical cores and not the physical ones.
+
+If you want to alter the number of threads spawned, you can set the
+environmental variable `RAYON_NUM_THREADS` to the desired number of
+threads or use the
+[`ThreadPoolBuilder::build_global` function](https://docs.rs/rayon/*/rayon/struct.ThreadPoolBuilder.html#method.build_global)
+method.
+
+## How does Rayon balance work between threads?
+
+Behind the scenes, Rayon uses a technique called **work stealing** to
+try and dynamically ascertain how much parallelism is available and
+exploit it. The idea is very simple: we always have a pool of worker
+threads available, waiting for some work to do. When you call `join`
+the first time, we shift over into that pool of threads. But if you
+call `join(a, b)` from a worker thread W, then W will place `b` into
+its work queue, advertising that this is work that other worker
+threads might help out with. W will then start executing `a`.
+
+While W is busy with `a`, other threads might come along and take `b`
+from its queue. That is called *stealing* `b`. Once `a` is done, W
+checks whether `b` was stolen by another thread and, if not, executes
+`b` itself. If W runs out of jobs in its own queue, it will look
+through the other threads' queues and try to steal work from them.
+
+This technique is not new. It was first introduced by the
+[Cilk project][cilk], done at MIT in the late nineties. The name Rayon
+is an homage to that work.
+
+[cilk]: http://supertech.csail.mit.edu/cilk/
+
+## What should I do if I use `Rc`, `Cell`, `RefCell` or other non-Send-and-Sync types?
+
+There a number of non-threadsafe types in the Rust standard library,
+and if your code is using them, you will not be able to combine it
+with Rayon. Similarly, even if you don't have such types, but you try
+to have multiple closures mutating the same state, you will get
+compilation errors; for example, this function won't work, because
+both closures access `slice`:
+
+```rust
+/// Increment all values in slice.
+fn increment_all(slice: &mut [i32]) {
+    rayon::join(|| process(slice), || process(slice));
+}
+```
+
+The correct way to resolve such errors will depend on the case.  Some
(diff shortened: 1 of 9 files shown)


nikomatsakis:
cc @KodrAus

rayon-core/src/join/mod.rs
```diff
@@ -32,7 +32,44 @@ mod test;
...
+///        let (lo, hi) = v.split_at_mut(mid);
+///        rayon::join(|| quick_sort(lo),
+///                    || quick_sort(hi));
+///    }
+/// }
+///
+/// let mut v = vec![5, 1, 8, 22, 0, 44];
+/// quick_sort(&mut v);
+/// assert_eq!(v, vec![0, 1, 5, 8, 22, 44]);
+///
```
KodrAus: This extra space will show up in the docs output.

FAQ.md
```diff
@@ -0,0 +1,227 @@
...
+fn increment_all(slice: &mut [i32]) {
+    rayon::join(|| process(slice), || process(slice));
+}
+```
+
+The correct way to resolve such errors will depend on the case.  Some
+cases are easy: for example, uses of [`Rc`] can typically be replaced
+with [`Arc`], which is basically equivalent, but thread-safe.
+
+Code that uses `Cell` or `RefCell`, however, can be a mite more complicated.
```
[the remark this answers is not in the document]
cuviper: "a mite" is an idiom meaning "somewhat".  "a bit" would be fine too.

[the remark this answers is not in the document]
KodrAus: Ah ok, I haven't heard that one before :)

rayon-core/src/join/mod.rs
nikomatsakis: Question: do you think I should include the `partition` method? I expected there to be a link to play, but I don't think there actually is...?

nikomatsakis: Decided to just uncomment it.

src/iter/mod.rs
```diff
@@ -30,6 +33,30 @@
...
+//! ways to create (or consume) parallel iterators:
+//!
+//! - Slices (`&[T]`, `&mut [T]`) offer methods like `par_split` and
+//!   `par_windows`, as well as various parallel sorting
+//!   operations. See [the `ParallelSlice` trait] for the full list.
+//! - Strings (`&str`) offer methods like `par_split` and `par_lines`.
+//!   See [the `ParallelString` trait] for the full list.
+//! - Various collections offer [`par_extend`], which grows a collection
+//!   given a parallel iterator. (You can use [`collect()`] to create a parallel
+//!   iterator.)
```
cuviper: Presumably you mean, "to create *a collection from* a parallel iterator."

src/iter/mod.rs
```diff
@@ -30,6 +33,30 @@
...
+//! [the `ParallelString` trait]: ../str/trait.ParallelString.html
+//! [`par_extend`]: trait.ParallelExtend.html
+//! [`collect()`]: trait.ParallelIterator.html#method.collect
+//!
+//! These are the full set of methods you can use to create a parallel iterator:
+//!
+//! - `par_iter` -- iterates over `&T` references to the items in the collection
+//! - `par_iter_mut` -- iterates over `&mut T` references to the items in the collection
+//! - `into_par_iter` -- iterates over the items in the collection, taking ownership of each
+//! - `par_split`
```
cuviper: Were you starting to reference `rayon::iter::split()` here?  There is a `par_split()` on slices and strings though.  I'm not sure we can realistically list a "full set of methods" anyway...

src/option.rs
```diff
@@ -7,8 +7,15 @@ use iter::plumbing::*;
 use std;
 use std::sync::atomic::{AtomicBool, Ordering};
 
-
-/// Parallel iterator over an option
+/// A parallel iterator over the value in [`Some`] variant of an [`Option`].
+///
+/// The iterator yields one value if the [`Option`] is a [`Some`], otherwise none.
+///
+/// This `struct` is created by the [`par_iter_mut`] function.
```
cuviper: Should be `into_par_iter`.

src/iter/mod.rs
nikomatsakis: Hmm yeah I think I had decided to move that complete list into a link to the trait.

src/iter/mod.rs
nikomatsakis: Yep.

src/option.rs
nikomatsakis: sigh

nikomatsakis:
@cuviper ok

cuviper:
bors delegate=KodrAus

FAQ.md
```diff
@@ -0,0 +1,227 @@
+# Rayon FAQ
+
+These file is for general questions that don't fit into the README or
```
KodrAus: This file?

FAQ.md
```diff
@@ -0,0 +1,227 @@
...
+crate docs.
+
+## How many threads will Rayon spawn?
+
+By default, Rayon uses the same number of threads as the number of CPUs
+available. Note that on systems with hyperthreading enabled this
+equals to the number of logical cores and not the physical ones.
```
KodrAus: s/equals to/equals

KodrAus:
Let's see if this works!

bors r+

PR #526, opened 2018-02-07
Add examples to more traits

cuviper:
This covers the simpler one-method traits for #420.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -1671,6 +1671,33 @@ pub trait IndexedParallelIterator: ParallelIterator {
 ///
 /// [`ParallelIterator`]: trait.ParallelIterator.html
 /// [`collect()`]: trait.ParallelIterator.html#method.collect
+///
+/// # Examples
+///
+/// Implementing `FromParallelIterator` for your type:
+///
+/// ```
+/// use rayon::prelude::*;
+/// use std::mem;
+///
+/// struct BlackHole {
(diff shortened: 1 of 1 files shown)


src/iter/mod.rs
```diff
@@ -1671,6 +1716,33 @@ pub trait IndexedParallelIterator: ParallelIterator {
...
+///     {
+///         let par_iter = par_iter.into_par_iter();
+///         BlackHole {
+///             mass: par_iter.count() * mem::size_of::<T>(),
+///         }
+///     }
+/// }
+///
+/// let bh: BlackHole = (0i32..1000).into_par_iter().collect();
+/// assert_eq!(bh.mass, 4000);
```
nikomatsakis: cute

src/iter/mod.rs
```diff
@@ -1671,6 +1716,33 @@ pub trait IndexedParallelIterator: ParallelIterator {
 ///
 /// [`ParallelIterator`]: trait.ParallelIterator.html
 /// [`collect()`]: trait.ParallelIterator.html#method.collect
+///
+/// # Examples
```
nikomatsakis: fwiw I often think that the idiomatic style for iterators out to be `from_iter` (and hence `from_par_iter`). I like that writing 

```rust
let x =
    BlackHole::from_par_iter(...);
```

kind of "automatically" gives the needed type-annotation. But oh well.

PR #530, opened 2018-02-12
Add examples to par_split_mut and par_chunks_mut

cuviper:
Also add an odd tail to the `par_chunks` example.

cc #420

### DIFF
--- a/src/slice/mod.rs
+++ b/src/slice/mod.rs
@@ -73,8 +73,8 @@ pub trait ParallelSlice<T: Sync> {
     ///
     /// ```
     /// use rayon::prelude::*;
-    /// let windows: Vec<_> = [1, 2, 3, 4].par_chunks(2).collect();
-    /// assert_eq!(vec![[1, 2], [3, 4]], windows);
+    /// let chunks: Vec<_> = [1, 2, 3, 4, 5].par_chunks(2).collect();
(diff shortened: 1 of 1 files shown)


src/slice/mod.rs
```diff
@@ -73,8 +73,8 @@ pub trait ParallelSlice<T: Sync> {
     ///
     /// ```
     /// use rayon::prelude::*;
-    /// let windows: Vec<_> = [1, 2, 3, 4].par_chunks(2).collect();
-    /// assert_eq!(vec![[1, 2], [3, 4]], windows);
+    /// let chunks: Vec<_> = [1, 2, 3, 4, 5].par_chunks(2).collect();
+    /// assert_eq!(chunks, vec![&[1, 2][..], &[3, 4], &[5]]);
```
KodrAus: We don't need the trailing `[..]` here, right?

cuviper: It needs to be forced as a slice instead of array, or else the type of `[5]` won't match.

I think I could do this like `let expected: Vec<&[_]> = ...` if that's clearer.

KodrAus: Ah yep. I personally think the way it is now is fine to read, I was just nit-picking for the sake of the review.

PR #531, opened 2018-02-12
Add examples to IndexedParallelIterator

cuviper:
cc #420


cuviper:
I didn't bother with `eq`/`lt`/etc., but in my defense neither does `Iterator` upstream... 😅

Issue #236, opened 2017-03-03
[docs] the `rayon` crate needs documentation


nikomatsakis:
This is the main entry-point for `rayon`. It should have some nice docs, probably quite similar to the README. That is, we would want to explain what rayon is all about, give some quick examples, and then direct people to the most important functions to learn more.

I would probably include direct links to:

- The `ParallelIterator` trait
- The `join()` and `scope()` methods (located in rayon-core)
- The `spawn_async()` and `spawn_future_async()` methods (located in rayon-core)
- The `ThreadPool` type for creating custom thread-pools (located in rayon-core)

KodrAus:
I like the idea of introducing a reader to the breadth of the `rayon` ecosystem using a top-down, task-based approach, like @nikomatsakis has pretty much described in the set of bullet points.

I'm not sure if it's already been discussed before, but another idea might be to keep the docs focused on API discovery and work on a separate book for exploring `rayon` in a more structured way, like `https://serde.rs`. Although it's probably easier to fill out all that information in the crate docs first, and then pull it out into a book at another point.

Is the idea of this issue that we flesh out the _Parallel Iterators_ and _Divide and conquer with `join`_ sections in the crate root and add some new ones for `futures` and the `ThreadPool`?

cuviper:
FWIW, the futures stuff is now in its own `rayon-futures` crate, and given that it's still unstable, I wouldn't bother mentioning it in `rayon` or `rayon-core` docs at all yet.

A book sounds great, but I'd prioritize improving the crate docs first. :)

nikomatsakis:
We have such documentation but we still want to revisit it. This is the initial landing place, after all.

nikomatsakis:
@KodrAus is going to do a survey.

KodrAus:
I'm making a start on this now in a branch including #511 and #512.

KodrAus:
Ok I've done a first pass over the docs and made some notes in [a Dropbox Paper](https://paper.dropbox.com/doc/rayon-docs-HoJk4CftX7Fl8LfnVoMBb). It's organised by module/type with some general thoughts on the docs in each. Some of it I think is probably already covered by issues here. I figured we could go over that and try clarify any other points into a concrete list.

cuviper:
Let's call this done, for the milestone's sake. 😄 
(Future work can be filed in more specific issues.)

PR #532, opened 2018-02-14
Release Rayon 1.0! (and rayon-core 1.4)

cuviper:


cuviper:
[rendered release notes](https://github.com/cuviper/rayon/blob/rayon-1.0/RELEASES.md#release-rayon-100--rayon-core-140)

The only outstanding milestone issue is #236, but completion on that seems nebulous.  I don't think we really need to block the release on docs anyway.

The last thing I noticed is that `rayon_core::ThreadPoolInitError` is not re-exported in `rayon`.  Should we add that?

jwass:
@cuviper I think the reason `ThreadPoolInitError` wasn't re-exported was that the old `GlobalPoolAlreadyInitialized` was never exported either. (At least that was my thinking at the time). Although maybe it still should be.

cuviper:
Also, those methods are now `build` -- should it be called `ThreadPoolBuildError`?

cuviper:
@jwass fair point, but still worth considering.

KodrAus:
> The only outstanding milestone issue is #236, but completion on that seems nebulous. I don't think we really need to block the release on docs anyway.

I agree, the docs are in pretty nice shape, and some of the tooling work going on in `rustdoc` this year could offer some new dimensions to expand the docs in anyway.

nikomatsakis:
Very exciting!

cuviper:
bors r=nikomatsakis

cuviper:
Published!

PR #533, opened 2018-02-14
Rename ThreadPoolInitError to ThreadPoolBuildError

jwass:
Re-export ThreadPoolBuildError from rayon


cuviper:
We need to either do this before #532 is published, or not at all.

PR #534, opened 2018-02-15
fix the relative links to the iter module

nikomatsakis:
Browsing the Rayon 1.0 docs I hit a few snags.


cuviper:
"links", "a few snags", but there's only one fix here?

cuviper:
Meh, that one fix is still correct.

bors r+

PR #480, opened 2017-11-26
Replace coco with crossbeam-deque

stjepang:
Coco is now deprecated and I don't intend to maintain it anymore. You should use Crossbeam instead.

This PR switches the dependency from `coco` to [`crossbeam-deque`](https://github.com/crossbeam-rs/crossbeam-deque). We still haven't re-exported the deque into the main `crossbeam` crate, but you can start using `crossbeam-deque` right now.

Benchmarks:

```
name                                                                before ns/iter         after ns/iter          diff ns/iter   diff %
factorial::factorial_iterator                                       24,802,712             24,427,893                 -374,819   -1.51%
factorial::factorial_join                                           10,636,501             10,459,655                 -176,846   -1.66%
factorial::factorial_par_iter                                       10,243,680             10,196,058                  -47,622   -0.46%
factorial::factorial_recursion                                      12,202,069             11,909,615                 -292,454   -2.40%
fibonacci::fibonacci_iterative                                      21                     21                                0    0.00%
fibonacci::fibonacci_join_1_2                                       60,217,053             58,844,783               -1,372,270   -2.28%
fibonacci::fibonacci_join_2_1                                       58,738,255             58,375,573                 -362,682   -0.62%
fibonacci::fibonacci_recursive                                      13,976,426             13,532,543                 -443,883   -3.18%
fibonacci::fibonacci_split_iterative                                24,218                 24,097                         -121   -0.50%
fibonacci::fibonacci_split_recursive                                5,582,896              6,771,896                 1,189,000   21.30%
find::size1::parallel_find_common                                   8,543                  8,466                           -77   -0.90%
find::size1::parallel_find_first                                    5,095                  4,828                          -267   -5.24%
find::size1::parallel_find_last                                     4,232,507              4,211,524                   -20,983   -0.50%
find::size1::parallel_find_middle                                   2,827,812              2,814,623                   -13,189   -0.47%
find::size1::parallel_find_missing                                  4,267,884              4,261,439                    -6,445   -0.15%
find::size1::serial_find_common                                     3,858                  3,811                           -47   -1.22%
find::size1::serial_find_first                                      1                      1                                 0    0.00%
find::size1::serial_find_last                                       4,258,009              4,251,158                    -6,851   -0.16%
find::size1::serial_find_middle                                     2,840,747              2,829,840                   -10,907   -0.38%
find::size1::serial_find_missing                                    4,259,732              4,256,293                    -3,439   -0.08%
[...]

PR #535, opened 2018-02-15
deps: update crossbeam-deque to 0.3

ignatenkobrain:


jeehoonkang:
Rayon intentionally depends on an old version of crossbeam-deque.  0.3 doesn't support Rust 1.13.

cuviper:
Yep, we negotiated that compatibility arrangement in #480.  Thanks anyway!

PR #541, opened 2018-02-21
split: use floored division in example assertion

cuviper:
Rounding can falsify the previous `width < 2 * height` assertions.  For
example, given lengths `(3, 3)`, it will split to `(1, 3)` and `(2, 3)`.

cc #539 @HadrienG2


HadrienG2:
Nice catch!

cuviper:
Thank Travis-CI for aligning the stars just the wrong way... 😉 

bors r=HadrienG2

Issue #548, opened 2018-03-01
pre-RFC: rayon::yield_now


cuviper:
If you're in a situation where you need to wait for some external resource, but don't want to block, then calling `rayon::yield_now` could let the current thread search for other work in the pool.  If nothing was found, or if you're not actually in a threadpool at the moment, it can just call `std::thread::yield_now`.

fredpointzero:
Hi, do you have any update on this? It will be really usefull!

cuviper:
I implemented this yesterday, and it was straightforward enough: https://github.com/rayon-rs/rayon/compare/master...cuviper:yield_now

I'm now wondering if it makes sense for the caller to be "blind" to what happened (work executed or just yielded), or whether we should even call `thread::yield_now()` at all.

A different design could be `poll() -> bool` returning `true` if any work was executed, and if not the caller can choose to follow up with `thread::yield_now()`, or sleep with exponential backoff, etc.

fredpointzero:
I think polling makes more sense, because my actual use case is:
> I know that I need to that condition A to be true and will be mutated by another thread of the thread pool. So, I want to pause this job for a short amount of time by executing another job to avoid a deadlock.

Knowing whether a job was executed can be important to detect that deadlock situation.

So I think that a poll design is better suited. 

I don't know if there is a use case where you want more insight, like was a local job executed? was a job stolen? or is there nothing to do? It can be as well an information returned by the poll.

cuviper:
@fredpointzero (or anyone else watching), I would appreciate feedback on #1026, thanks!

Issue #46, opened 2016-02-05
Usage with “Read” or “Iterator” as provider?


flying-sheep:
I see that only ranges and slices can be converted into parallel iterators.

i wonder how to best utilize rayon while reading from a file.

ATM I read files into buffers, then parse them using an iterator. i imagine there’s a better way to use rayon than:

``` rust
let buffer = String::new();
File::open(some_path)?.read_to_string(buffer)?;  // iterate unindexed data once
let records: Vec<_> = ParseIter::new(buffer).collect();  // iterate unindexed data again
records.par_iter_mut().for_each(do_stuff);  // iterate indexed data in parallel
```

nikomatsakis:
Well, right now there isn't really a much better way (though you could handwrite a recursive function with `join`), but I'd like to add a kind of "work queue" abstraction (basically similar to scoped threads) that would probably address this use case.

Michael-F-Bryan:
This seems like a fairly popular feature judging on the number of issues people have created that are slight variations on this general `Iterator` to `ParallelIterator` adapter idea. Has there been any progress on this?

My particular use case is that I've got an iterator which yields every permutation of 4 numbers from 18 to 200 (around 167,961,600,000,000 permutations), then does a bunch of filters and maps to find a bunch of "ideal" combinations. In this case it's not feasible to save the combinations into a `Vec`, yet the filters and maps are all trivially parallelizable.

cuviper:
It's a popular request, but there's not an obvious way to implement it.  We've had a few ideas, but not any progress to report for it, sadly.

> My particular use case is that I've got an iterator which yields every permutation of 4 numbers from 18 to 200 (around 167,961,600,000,000 permutations),

Perhaps I misunderstand, but wouldn't the number of permutations be (200-18)<sup>4</sup> -- roughly 10<sup>9</sup>?  If you can set up that generator with ranges, you can parallelize that.
`(18..200).into_par_iter().for_each(|x| (180..200).into_par_iter().for_each(|y| ...))`

> In this case it's not feasible to save the combinations into a Vec,

Another strategy is batch this work.  Generate a manageable amount into a `Vec`, process those in parallel, and repeat in a loop until you've done them all.

grahame:
[...]

PR #550, opened 2018-03-07
add bridge from Iterator to ParallelIterator

QuietMisdreavus:
Half of https://github.com/rayon-rs/rayon/issues/46

This started getting reviewed in https://github.com/QuietMisdreavus/polyester/pull/6, but i decided to move my work to Rayon proper.

This PR adds a new trait, `AsParallel`, an implementation on `Iterator + Send`, and an iterator adapter `IterParallel` that implements `ParallelIterator` with a similar "cache items as you go" methodology as Polyester. I introduced a new trait because `ParallelIterator` was implemented on `Range`, which is itself an `Iterator`.

The basic idea is that you would start with a quick sequential `Iterator`, call `.as_parallel()` on it, and be able to use `ParallelIterator` adapters after that point, to do more expensive processing in multiple threads.

The design of `IterParallel` is like this:

* `IterParallel` defers background work to `IterParallelProducer`, which implements `UnindexedProducer`.
* `IterParallelProducer` will split as many times as there are threads in the current pool. (I've been told that https://github.com/rayon-rs/rayon/pull/492 is a better way to organize this, but until that's in, this is how i wrote it. `>_>`)
* When folding items, `IterParallelProducer` keeps a `Stealer` from `crossbeam-deque` (added as a dependency, but using the same version as `rayon-core`) to access a deque of items that have already been loaded from the iterator.
* If the `Stealer` is empty, a worker will attempt to lock the Mutex to access the source `Iterator` and the `Deque`.
  * If the Mutex is already locked, it will call `yield_now`. The implementation in polyester used a `synchronoise::SignalEvent` but i've been told that worker threads should not block. In lieu of https://github.com/rayon-rs/rayon/issues/548, a regular spin-loop was chosen instead.
  * If the Mutex is available, the worker will load a number of items from the iterator (currently (number of threads * number of threads * 2)) before closing the Mutex and continuing.
  * (If the Mutex is poisoned, the worker will just... stop. Is there a recommended approach here? `>_>`)

This design is effectively a first brush, has [the same caveats as polyester](https://docs.rs/polyester/0.1.0/polyester/trait.Polyester.html#implementation-note), probably needs some extra features in rayon-core, and needs some higher-level docs before i'm willing to let it go. However, i'm putting it here because it was not in the right place when i talked to @cuviper about it last time.

### DIFF
--- a/src/iter/as_parallel.rs
+++ b/src/iter/as_parallel.rs
@@ -1,0 +1,160 @@
+use crossbeam_deque::{Deque, Stealer, Steal};
+
+use std::thread::yield_now;
+use std::sync::{Mutex, TryLockError};
+use std::sync::atomic::{AtomicUsize, AtomicBool, Ordering};
+
+use iter::*;
+use current_num_threads;
+
+/// Conversion trait to convert an `Iterator` to a `ParallelIterator`.
+///
+/// This needs to be distinct from `IntoParallelIterator` because that trait is already implemented
+/// on a few `Iterator`s, like `std::ops::Range`.
+pub trait AsParallel {
+    /// What is the type of the output `ParallelIterator`?
+    type Iter: ParallelIterator<Item = Self::Item>;
+
+    /// What is the `Item` of the output `ParallelIterator`?
+    type Item: Send;
+
+    /// Convert this type to a `ParallelIterator`.
+    fn as_parallel(self) -> Self::Iter;
+}
+
+impl<T: Iterator + Send> AsParallel for T
+    where T::Item: Send
+{
+    type Iter = IterParallel<T>;
+    type Item = T::Item;
+
+    fn as_parallel(self) -> Self::Iter {
+        IterParallel {
+            iter: self,
+        }
+    }
+}
+
+/// `IterParallel` is a parallel iterator that wraps a sequential iterator.
+#[derive(Debug)]
+pub struct IterParallel<Iter> {
+    iter: Iter,
+}
+
+impl<Iter: Iterator + Send> ParallelIterator for IterParallel<Iter>
+    where Iter::Item: Send
+{
+    type Item = Iter::Item;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+        where C: UnindexedConsumer<Self::Item>
+    {
+        let split_count = AtomicUsize::new(current_num_threads());
+        let deque = Deque::new();
+        let stealer = deque.stealer();
+        let done = AtomicBool::new(false);
+        let iter = Mutex::new((self.iter, deque));
+
+        bridge_unindexed(IterParallelProducer {
+            split_count: &split_count,
+            done: &done,
+            iter: &iter,
+            items: stealer,
+        }, consumer)
+    }
+}
+
+struct IterParallelProducer<'a, Iter: Iterator + 'a> {
+    split_count: &'a AtomicUsize,
+    done: &'a AtomicBool,
+    iter: &'a Mutex<(Iter, Deque<Iter::Item>)>,
+    items: Stealer<Iter::Item>,
+}
+
+// manual clone because T doesn't need to be Clone, but the derive assumes it should be
+impl<'a, Iter: Iterator + 'a> Clone for IterParallelProducer<'a, Iter> {
+    fn clone(&self) -> Self {
+        IterParallelProducer {
+            split_count: self.split_count,
+            done: self.done,
+            iter: self.iter,
+            items: self.items.clone(),
+        }
+    }
+}
+
+impl<'a, Iter: Iterator + Send + 'a> UnindexedProducer for IterParallelProducer<'a, Iter>
+    where Iter::Item: Send
+{
+    type Item = Iter::Item;
+
+    fn split(self) -> (Self, Option<Self>) {
+        let mut count = self.split_count.load(Ordering::SeqCst);
+
+        loop {
+            let done = self.done.load(Ordering::SeqCst);
+            match count.checked_sub(1) {
+                Some(new_count) if !done => {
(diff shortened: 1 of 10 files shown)


src/iter/as_parallel.rs
```diff
@@ -0,0 +1,161 @@
...
+                    } else {
+                        // our cache is out of items, time to load more from the iterator
+                        match self.iter.try_lock() {
+                            Ok(mut guard) => {
+                                let count = current_num_threads();
+                                let count = (count * count) * 2;
+
+                                let (ref mut iter, ref deque) = *guard;
+
+                                for _ in 0..count {
```
cuviper: Instead of reading exactly `count` items, how about `while deque.len() < count`?  This way, I'm imagining that one thread could become the de facto reader thread, just pushing items as fast as the others can pop them out.  But if the other threads are too busy, this thread will fill the deque up to `count` and then break out to resume processing items itself.  This will be dynamic too, so when other threads get the time they'll start cooperating again.

QuietMisdreavus:
I don't have any specific performance numbers for this on hand, at least not any that don't just count the thread-spawn/mutex-lock costs on a job that would be handled much better sequentially. Maybe i should borrow a task from one of the people that requested this, so i have a proper task to measure. :P

nikomatsakis:
I'd love to review this too! I'm adding it to my calendar for later this week, so I can give it some time.

nikomatsakis:
OK -- I reviewed this with @QuietMisdreavus. Looks pretty good! I think we should change the name, though I don't know what name I want. Maybe `parallelize`? `adapt_into_par_iter`? Something else? Maybe an RFC?! :)

Also, we added some benchmarks. We found that for nbody is was approx. the same as the traditional par iter, but for game of life is was radically slower. Meh.

nikomatsakis:
@QuietMisdreavus @cuviper I'm sorry i'm super slow, but I'd like to see this land -- I think we had settle on something involving the word "bridge" for the name, right?

I was thinking that most of the iterator combinators (e.g., [`map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map)) take `fn(self)` but are not named `into`. Perhaps we should follow suite here, and name this `bridge_to_par_iter`?

cuviper:
On gitter, the "bridge" ideas were `into_par_bridge` and `into_par_bridge_iter`.

QuietMisdreavus:
I kinda like the idea of just `par_bridge`? There's a parallel with `par_iter`, but it's distinct enough to highlight that it's separate from other parallel iterators.

QuietMisdreavus:
I've pushed a commit that renames the trait from `AsParallel` to `ParallelBridge` (and the method name from `as_parallel` to `par_bridge`). What do you think?

src/iter/par_bridge.rs
```diff
@@ -0,0 +1,162 @@
...
+
+use iter::ParallelIterator;
+use iter::plumbing::{UnindexedConsumer, UnindexedProducer, bridge_unindexed, Folder};
+use current_num_threads;
+
+/// Conversion trait to convert an `Iterator` to a `ParallelIterator`.
+///
+/// This needs to be distinct from `IntoParallelIterator` because that trait is already implemented
+/// on a few `Iterator`s, like `std::ops::Range`.
+pub trait ParallelBridge {
```
nikomatsakis: The name is great, but we definitely need better rust-docs. We should include something like this:

---

That this "bridges" from a standard sequential iterator to a parallel one. This has the advantage of letting you parallelize just about anything, but it can be distinctly less efficient than the "native" parallel iterators produced by `par_iter`. 

---

We also need an example or two I think — maybe one showing some combinators that don't work in parallel land?

src/iter/par_bridge.rs
```diff
@@ -0,0 +1,198 @@
...
+/// output.sort_unstable();
+///
+/// assert_eq!(&*output, &["one!", "three!", "two!"]);
+/// ```
+pub trait ParallelBridge {
+    /// What is the type of the output `ParallelIterator`?
+    type Iter: ParallelIterator<Item = Self::Item>;
+
+    /// What is the `Item` of the output `ParallelIterator`?
+    type Item: Send;
```
cuviper: Are these associated types actually useful?  We could just have:

    fn par_bridge(self) -> IterParallel<Self>;

and be done with it.  That also offers some amount of control over who can `impl` this trait, as only we can construct that type.

For comparison, the `ParallelSlice` and `ParallelString` extensions just return their types directly.

src/iter/par_bridge.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/// `IterParallel` is a parallel iterator that wraps a sequential iterator.
+///
+/// This type is created when using the `par_bridge` method on `ParallelBridge`. See the
+/// [`ParallelBridge`] documentation for details.
+///
+/// [`ParallelBridge`]: trait.ParallelBridge.html
+#[derive(Debug)]
+pub struct IterParallel<Iter> {
```
cuviper: Should we "bridge" this name too?  Maybe `IterBridge`, or even just `Bridge`?

src/iter/par_bridge.rs
QuietMisdreavus: I think i wrote those in when i wasn't sure whether this trait would be used on other items, so i just made it a clone of `ParallelIterator`. I didn't realize that about `ParallelSlice`/`ParallelString`, so i can go in and change this up.

src/iter/par_bridge.rs
QuietMisdreavus: That would make more sense - this is still the same name it had from when the trait was named `AsParallel`. Between those two, I'm more of a fan of `IterBridge`.

cuviper:
Time to cross this bridge... thanks!

bors r+

PR #552, opened 2018-03-12
Dependency updates:

golem131:
* crossbeam-epoch 0.3.0 -> 0.4.0
* libc

### DIFF
--- a/rayon-core/Cargo.toml
+++ b/rayon-core/Cargo.toml
@@ -16,8 +16,8 @@ categories = ["concurrency"]
 [dependencies]
 rand = ">= 0.3, < 0.5"
 num_cpus = "1.2"
-crossbeam-deque = "0.2.0"
-libc = "0.2.16"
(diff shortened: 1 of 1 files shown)


cuviper:
We're deliberately holding back crossbeam-deque to maintain compatibility with older rustc.  They only check the most recent stable rustc in their development, but they gave us extra support in https://github.com/rayon-rs/rayon/pull/480#issuecomment-364661178.  We probably should add a comment about this though, as you're not the first to try updating it -- like #535.

I see no reason to update the libc dependency either.  0.2.16 is just the version we chose when that dependency was added, but this acts like a minimum.  Cargo will automatically choose the latest compatible version possible, in this case anything from 0.2.16 up to (but not including) 0.3.0.

cuviper:
I pushed commit 0c8be8d2b82f5540465c7d9ff25f057743a88c22 to explain the older crossbeam-deque.

PR #554, opened 2018-03-12
Fix typos

matthiasbeyer:
I'm not entirely sure about the second patch in here, please confirm that this is correct.


cuviper:
Looks good to me, thanks!

bors r+

PR #558, opened 2018-03-22
Document how par_chunks lengths may be shorter

### LINKED ISSUE #557: Document that only last chunk from par_chunks(_mut) may be smaller than chunk_size
Transcript from Gitter:

Andrew Gaspar @AndrewGaspar Mar 16 16:05
Are the “leftovers” for par_chunks and par_chunks_mut guaranteed to fall at the end? I.e. will the first and only chunk smaller than chunk_size fall at the logical end of the parallel iterator
I’m trying to compose par_chunks with enumerate, and it’s important that I can multiple the index from enumerate by chunk_size to get the offset into the original slice of the slice in my parallel for_each.

Andrew Gaspar @AndrewGaspar Mar 16 16:18
I assume this is the case, but it would be great if it was documented.

Josh Stone @cuviper 17:21
@AndrewGaspar yes, only the last chunk may be smaller
that is documented on IndexedParallelIterator::chunks
but not on the slice methods, you're right

cuviper:
cc @AndrewGaspar
Fixes #557.

### DIFF
--- a/src/slice/mod.rs
+++ b/src/slice/mod.rs
@@ -49,8 +49,8 @@ pub trait ParallelSlice<T: Sync> {
         }
     }
 
-    /// Returns a parallel iterator over all contiguous windows of
-    /// length `size`. The windows overlap.
+    /// Returns a parallel iterator over all contiguous windows of length
+    /// `window_size`. The windows overlap.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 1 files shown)


AndrewGaspar:
lgtm!

cuviper:
Great, thanks!

bors r=@AndrewGaspar

PR #560, opened 2018-03-26
Parity with std::str iterators

### LINKED ISSUE #559: Chars does not implement IndexedParallelIterator#enumerate
[Playground](https://play.rust-lang.org/?gist=fbfd4660501e9807394a10119555f318&version=stable)

It would be great if I could use `"foobar".par_chars().enumerate()`.

Thanks for considering!

cuviper:
This mimics the rest of the iterators in `std::str` with methods on the
`ParallelString` trait.

Closes #559, as `par_char_indices` is the closest we can do for that.


src/str.rs
```diff
@@ -63,6 +64,55 @@ pub trait ParallelString {
...
+    /// ```
+    fn par_bytes(&self) -> Bytes {
+        let bytes = self.as_parallel_string().as_bytes();
+        Bytes { inner: bytes.par_iter().cloned() }
+    }
+
+    /// Returns a parallel iterator over a string encoded as UTF-16.
+    ///
+    /// Note that surrogate pairs (for codepoints greater than `U+FFFF`) are
+    /// produced as separate items, but will not be split across threads.
```
cuviper: This feature of not splitting pairs across threads came naturally from the way the string is split.  But I just realized that I didn't make a similar guarantee on `par_bytes` -- perhaps we should?  We could only do that as an unindexed producer though, to avoid arbitrary splits.

cuviper: I decided I am going to make that guarantee for `par_bytes()` too.  I think this is a more interesting feature to provide.  If someone really wanted the indexed `Bytes`, they can still use a raw `s.as_bytes().par_iter().cloned()` to get that effect.

src/str.rs
```diff
@@ -448,3 +716,113 @@ impl<'ch> ParallelIterator for SplitWhitespace<'ch> {
...
+        let producer = MatchesProducer { chars: self.chars, pattern: &self.pattern };
+        bridge_unindexed(producer, consumer)
+    }
+}
+
+impl<'ch, 'pat, P: Pattern> UnindexedProducer for MatchesProducer<'ch, 'pat, P> {
+    type Item = &'ch str;
+
+    fn split(self) -> (Self, Option<Self>) {
+        match split(self.chars) {
```
nikomatsakis: So what am I missing here -- how do we know we are not splitting right in the middle of a match?

nikomatsakis: Maybe this is just out of cache for me...

cuviper: At the moment, we only support single-character patterns for splits and matches.

PR #561, opened 2018-03-28
Try another approach to hacking rand in CI

cuviper:
Using `sed` to hack rand-0.3.20 in `Cargo.lock` worked for a while, but
now it's giving us an invalid lockfile, because some of the test/demo
dependencies need rand-0.4 themselves.

Instead, we now just tweak `rayon-core/Cargo.toml` long enough to
generate a complete lock file, which is enough to get a working CI build
on Rust 1.13 again.


cuviper:
Bad download on Windows.

bors r+

cuviper:
CI did ultimately succeed, merging...

PR #563, opened 2018-04-07
[WIP] Add try_fold, try_for_each, try_reduce

### LINKED ISSUE #495: Implement functionality like `Iterator::try_fold`
[`Iterator::try_fold`](https://doc.rust-lang.org/nightly/std/iter/trait.Iterator.html#method.try_fold) is currently unstable -- rust-lang/rust#45594.

For `Iterator::fold`, rayon has `fold` that operates per-split, which you must `reduce` to get down to one final accumulator.  I expect `try_fold` should be similar.

We should wait until it's stabilized in `std` before we add it to rayon, so we can mimic the final signature.

cuviper:
There are six variations here, matching the existing non-try suite:

- `try_fold` and `try_fold_with`
- `try_for_each` and `try_for_each_with`
- `try_reduce` and `try_reduce_with`

All of them operate on `Try::Ok` values similar to the exiting non-try
methods, and short-circuit early to return any `Try::Error` value seen.
This `Try` is a pub-in-private clone of the unstable `std::ops::Try`,
implemented for `Option<T>` and `Result<T, E>`.

TODO and open questions:

- [ ] Needs documentation, examples, and tests.
- [x] Should we wait for `Iterator::try_fold` and `try_for_each` to
      reach rust stable?  They were stabilized in rust-lang/rust#49607,
      but there's always a chance this could get backed out.
    - **Resolved**: they're stable in 1.27
- [x] Should we wait for stable `std::ops::Try`?  We could just keep
      ours private for now, and change to publicly use the standard
      trait later (on sufficiently new rustc).
    - **Resolved**: keep our pub-in-private `Try` for now.
- [x] Should `try_fold` and `try_fold_with` continue to short-circuit
      globally, or change to only a local check?
  - When I first implemented `try_reduce_with`, I use a `try_fold` +
    `try_reduce` combination, like `reduce_with`'s implementation, but
    I didn't like the idea of having double `full: AtomicBool` flags
    in use.
  - If `try_fold` only errors locally, then other threads can continue
    folding normally, and you can decide what to do with the error
    when you further reduce/collect/etc.  e.g. A following `try_reduce`
    will still short-circuit globally.
  - **Resolved**: changed to just a local check.

Closes #495.

### DIFF
--- a/src/iter/try_reduce.rs
+++ b/src/iter/try_reduce.rs
@@ -1,7 +1,7 @@
 use super::ParallelIterator;
 use super::plumbing::*;
 
-use std::ops::Try;
+use super::private::Try;
 use std::sync::atomic::{AtomicBool, Ordering};
 
 pub fn try_reduce<PI, R, ID, T>(pi: PI, identity: ID, reduce_op: R) -> T
@@ -78,7 +78,10 @@ impl<'r, R, ID, T> Reducer<T> for TryReduceConsumer<'r, R, ID>
           T: Try
 {
     fn reduce(self, left: T, right: T) -> T {
-        (self.reduce_op)(left?, right?)
+        match (left.into_result(), right.into_result()) {
+            (Ok(left), Ok(right)) => (self.reduce_op)(left, right),
+            (Err(e), _) | (_, Err(e)) => T::from_error(e),
+        }
     }
 }
 
@@ -97,7 +100,7 @@ impl<'r, R, T> Folder<T> for TryReduceFolder<'r, R, T>
     fn consume(self, item: T) -> Self {
         let reduce_op = self.reduce_op;
         let result = self.result.and_then(|left| {
-            reduce_op(left, item?).into_result()
+            reduce_op(left, item.into_result()?).into_result()
         });
         TryReduceFolder {
             result: result,
@@ -111,7 +114,7 @@ impl<'r, R, T> Folder<T> for TryReduceFolder<'r, R, T>
         let reduce_op = self.reduce_op;
         let result = self.result.and_then(|mut left| {
             for right in iter {
-                left = reduce_op(left, right?)?;
+                left = reduce_op(left, right.into_result()?).into_result()?;
             }
             Ok(left)
         });
--- a/src/iter/try_reduce.rs
+++ b/src/iter/try_reduce.rs
@@ -102,22 +102,9 @@ impl<'r, R, T> Folder<T> for TryReduceFolder<'r, R, T>
         let result = self.result.and_then(|left| {
             reduce_op(left, item.into_result()?).into_result()
         });
-        TryReduceFolder {
-            result: result,
-            ..self
+        if result.is_err() {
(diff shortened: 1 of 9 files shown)


nikomatsakis:
I'm torn about the proper behavior of `try_fold`; that's a very interesting question. So let me summarize to make sure I understand:

- Today, `try_fold` yields up an iterator over `Option<T>` values (or whatever);
- But once it encounters a `None` anywhere, it makes a best effort to stop producing further values *anywhere*;
- So typically you would have at most one `None` in the sequence (though it's probably possible to have to more than one, if the timing is poor).

The alternative being proposed is that, once fold has settled on a sequential list of things to process, it will always finish, and hence you can very easily get multiple `None` in the sequence, and `fold` avoids global communication via an `AtomicBool`.

**Given these two options, I suspect that we want the current behavior.** It seems to me that the are only reason to want the second behavior is if you somehow wanted to ensure that some elements were processed even though a `None` would occur at some other point; but given that you don't control where the splits occur in a fold operation, you are not actually given that assurance.

The only property that you do get, afaict, is that if a `None` occurs *later* in the sequence, then the *side-effects* of earlier operations will occur (the side-effects of later operations *may* occur). That seems pretty thin and unlikely to be what you wanted.

In contrast, what we *give up* is that we potentially do a bunch of extra work, producing `Some(_)` values that will later be discarded.

cuviper:
> The alternative being proposed is that, once fold has settled on a sequential list of things to process, it will always finish, and hence you can very easily get multiple `None` in the sequence, and fold avoids global communication via an `AtomicBool`.

I'm proposing that `try_fold` would still short-circuit its local work, and just lose the global check.  My thought is that you'd most often combine with `try_reduce`, which has its own global check.  But that won't know there's any `Err`/`None` until the preceding `try_fold` passes it through, so we still need that to happen ASAP.

I don't know if there's an actual useful semantic here, beyond avoiding redundant `AtomicBool`s.  This is a small optimization for the `try_fold().try_reduce()` case, but the change would be noticeable if anybody did `try_fold()` followed by something that *doesn't* short circuit.  I don't know if that's realistic.  You're right that folks don't have control over splitting in general, but `with_min/max_len` do offer some control.

cuviper:
I guess what you thought I was proposing was that all sequential runs would be processed up to an `Err`/`None`, then resume that local sequence with a new accumulator?  So a local fold might actually produce multiple accumulations for the next consumer... that's interesting too.  I don't know if that's useful either.

src/iter/try_reduce.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    fn complete(self) -> T {
+        match self.result {
+            Ok(ok) => T::from_ok(ok),
+            Err(error) => T::from_error(error),
+        }
+    }
+
+    fn full(&self) -> bool {
+        self.result.is_err() || self.full.load(Ordering::Relaxed)
```
nikomatsakis: Doesn't `self.result.is_err()` *impliy* `self.full` is true? (Or will be shortly?)

PR #568, opened 2018-04-23
link to par_sort resolves

AndyGauge:
I found a dead link.

### DIFF
--- a/rayon-core/src/join/mod.rs
+++ b/rayon-core/src/join/mod.rs
@@ -39,7 +39,7 @@ mod test;
 /// sort for real, you should prefer [the `par_sort` method] offered
 /// by Rayon).
 ///
-/// [the `par_sort` method]: ../slice/trait.ParallelSliceMut.html#method.par_sort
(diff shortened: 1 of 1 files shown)


cuviper:
Ah, this is awkward.  Your fix is correct for the re-export as [`rayon::join`](https://docs.rs/rayon/1.0.1/rayon/fn.join.html#examples), but it will still be a broken link for [`rayon_core::join`](https://docs.rs/rayon-core/1.4.0/rayon_core/fn.join.html#examples).  I'm not sure if there's any good solution for cross-crate links like that.

AndyGauge:
I forgot about rayon_core with the links.  I tested this in both.

cuviper:
It will still be broken when rayon-core is documented on its own, as docs.rs does, but I don't know what to do about that.  Anyway, this is a clear improvement, thanks!

bors r+

AndyGauge:
I didn't look carefully at the link you sent, but now I see what you are talking about.  I don't know how to solve that.

PR #569, opened 2018-04-25
[Do not merge] Changes to Rayon in fork for rustc

Zoxc:
This just shows to changes intended for use in rustc via a rustc_rayon and rustc_rayon_core fork of Rayon.

### DIFF
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -50,6 +50,7 @@ mod scope;
 mod sleep;
 mod spawn;
 mod test;
+mod worker_local;
 mod thread_pool;
 mod unwind;
 mod util;
@@ -65,6 +66,7 @@ pub use thread_pool::current_thread_has_pending_tasks;
 pub use join::{join, join_context};
 pub use scope::{scope, Scope};
 pub use spawn::spawn;
+pub use worker_local::WorkerLocal;
 
 /// Returns the number of threads in the current registry. If this
 /// code is executing within a Rayon thread-pool, then this will be
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -141,6 +141,9 @@ pub struct ThreadPoolBuilder {
     /// Closure invoked on worker thread exit.
     exit_handler: Option<Box<ExitHandler>>,
 
+    /// Closure invoked on worker thread start.
+    main_handler: Option<Box<MainHandler>>,
+
     /// If false, worker threads will execute spawned jobs in a
     /// "depth-first" fashion. If true, they will do a "breadth-first"
     /// fashion. Depth-first is the default.
@@ -169,6 +172,12 @@ type StartHandler = Fn(usize) + Send + Sync;
 /// closure is passed the index of the thread on which is is invoked.
 /// Note that this same closure may be invoked multiple times in parallel.
 type ExitHandler = Fn(usize) + Send + Sync;
+
+/// The type for a closure that gets invoked with a
+/// function which runs rayon tasks.
+/// The closure is passed the index of the thread on which it is invoked.
+/// Note that this same closure may be invoked multiple times in parallel.
+type MainHandler = Fn(usize, &mut FnMut()) + Send + Sync;
 
 impl ThreadPoolBuilder {
     /// Creates and returns a valid rayon thread pool builder, but does not initialize it.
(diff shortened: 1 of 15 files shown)


cuviper:
Can you give a high level summary of what the changes are, and why they are needed?

Zoxc:
I've made a commit for each feature added.
- The thread local value is just a thread-local which gets read when jobs are created and set when they are executed. This means that jobs start out with the same thread-local value as their parent. It is added to allow Rayon jobs to access [rustc's ImplicitCtxt](https://github.com/rust-lang/rust/blob/1eab9c5a1b85db4ca850114814a7df3e289fcd67/src/librustc/ty/context.rs#L1755) which is  a thread-local value which is updated during the lifetime of the thread-pool
- `main_handler` is added to allow for an ergonomic way to set scoped thread-local state for worker threads, since APIs dealing with scoped things use closures and `begin_handler` and `end_handler` doesn't let you do this
- The scoped thread pool is used so that we can point thread-locals in the work threads to the stack on the main thread. This allows us to have temporary "globals" for the entire thread pool.
- `ThreadLocal` is added as an ergonomic and efficient way to create thread-locals with a lifetime shorter than the thread-pool

rayon-core/src/tlv.rs
```diff
@@ -0,0 +1,30 @@
...
+            TLV.with(|tlv| tlv.set(self.0));
+        }
+    }
+    let _reset = Reset(get());
+    TLV.with(|tlv| tlv.set(value));
+    f()
+}
+
+/// Sets the current thread-local value
+pub fn set(value: usize) {
```
nikomatsakis: ok so this is some sort of "token" where:

- it starts as 0
- you can set it and get it
- when you spawn a job, it inherits the value at the time that job was spawned?

rayon-core/src/lib.rs
```diff
@@ -170,6 +173,12 @@ type StartHandler = Fn(usize) + Send + Sync;
 /// Note that this same closure may be invoked multiple times in parallel.
 type ExitHandler = Fn(usize) + Send + Sync;
 
+/// The type for a closure that gets invoked with a
+/// function which runs rayon tasks.
+/// The closure is passed the index of the thread on which it is invoked.
+/// Note that this same closure may be invoked multiple times in parallel.
```
nikomatsakis: How is this distinct from the `StartHandler`?

nikomatsakis: Oh, I see, it kind of "takes over" the thread

Zoxc: It allows up to run code both before and after the Rayon worker, so it is more powerful than `StartHandler` and `ExitHandler`. [You can see it used here](https://github.com/rust-lang/rust/pull/50235/files#diff-698522bf32a92da4331b4bd3f6ed31c6R115).

rayon-core/src/tlv.rs
Zoxc: Yes.

rayon-core/src/thread_local.rs
```diff
@@ -0,0 +1,74 @@
...
+                panic!("ThreadLocal can only be used on the thread pool it was created on")
+            }
+            &self.locals[(*worker_thread).index].0
+        }
+    }
+}
+
+impl<T> ThreadLocal<Vec<T>> {
+    /// Joins the elements of all the thread locals into one Vec
+    pub fn join(self) -> Vec<T> {
```
nikomatsakis: Apart from this method, I believe we could just use a "regular" thread-local value, no?

nikomatsakis: I wonder if @cuviper's "broadcast" primitive could be used to simulate this from a regular thread-local. I don't remember if it lets you return values — but it probably *could*. =)

Zoxc: If by "regular" thread-local value, you mean something like [the thread_local crate](https://crates.io/crates/thread_local), then yes. However that is less efficient since it doesn't exploit the fact that our thread pool has a fixed number of threads.

nikomatsakis:
My take on these changes:

- I wouldn't want to land them in rayon proper yet (I know that's not the current plan). Mostly they seem quite reasonable but I'd want to spend some time thinking about whether this is the right approach and if we can polish the APIs.
  - In particular, I think that having some form of "task-local, inherited" state (and "worker-local") in Rayon makes a lot of sense.
  - (Reminder to myself that I need to carve out some dedicated Rayon time and in particular I think we need an RFC process, but I've been saying that for a while I know.)
- However, the changes all look technically fine. I have to go review the rustc side to see how they all fit in. I would like to figure out how the scoped thread pools in particular fit in.

PR #571, opened 2018-04-30
Implement our own PRNG for rayon-core

### LINKED ISSUE #570: Consider using our own PRNG in rayon-core
We have a [kludge in `.travis.yml`](https://github.com/rayon-rs/rayon/blob/5f98c019abc4d40697e83271c072cb2b92966f46/.travis.yml#L19-L23) to deal with `rand` and its rustc compatibility.  It would be nice to do better here.

We currently use `rand` in two ways:

- Some tests use randomized input, often seeded.
- Worker threads randomize which other threads they try to steal from.

We only run tests on nightly right now (barring #536), so only the use by worker threads matters for rustc compatibility in general.  We use `rand::weak_rng()` for that, but it would probably be just fine to implement a simple PRNG locally.  Then we could get rid of the compatibility kludge, and just keep `rand` as a dev-dependency.

FWIW, `rand` is also reconsidering its compatibility story in rust-lang-nursery/rand#413.

cuviper:
Our need for randomness is very simple, just to spread the work-stealing
somewhat equally across all possible victims.  Rather than worrying
about compatibility and updates to the `rand` crate, we now implement a
simple `xorshift*` generator for this.  We keep `rand` only as a dev-
dependency for testing purposes.

Closes #570.


rayon-core/src/registry.rs
```diff
@@ -716,3 +705,36 @@ pub fn in_worker<OP, R>(op: OP) -> R
...
+    state: Cell<u64>,
+}
+
+impl XorShift64Star {
+    fn new() -> Self {
+        // Any non-zero seed will do -- this uses a stack address, which should be unique since
+        // we're only calling this once from each new thread.  The address is also shift-XORed to
+        // give us some high bits even on 32-bit platforms.
+        let base = 0;
+        let addr = &base as *const _ as u64;
```
nikomatsakis: Hmm. I'm not crazy about this technique for getting a pointer into the stack. It feels fragile in the face of potential Rust unsafe code guidelines — e.g., I could sort of imagine `base` getting promoted to read-only memory or something. But probably we wouldn't be able to get away with such a rule anyway. It might be nice to have some kind of intrinsic for "random stack address", though — this is something I want to do from time to time. Anyhow I suppose it's fine "as is" for now, I filed https://github.com/nikomatsakis/rust-memory-model/issues/47 to track it and ponder it more deeply. =)

rayon-core/src/registry.rs
```diff
@@ -716,3 +705,36 @@ pub fn in_worker<OP, R>(op: OP) -> R
...
+        let seed = addr ^ (addr << 32);
```
nikomatsakis: can we add a `assert!(seed != 0);` just as a failsafe?

rayon-core/src/registry.rs
```diff
@@ -602,17 +600,8 @@ impl WorkerThread {
...
-        let start = {
-            // OK to use this UnsafeCell because (a) this data is
-            // confined to current thread, as WorkerThread is not Send
-            // nor Sync and (b) rand crate will not call back into
-            // this method.
-            let rng = &mut *self.rng.get();
-            rng.next_u32() % num_threads as u32
-        } as usize;
+
+        let start = self.rng.next() as usize % num_threads;
```
nikomatsakis: Nit: since we control `rng`, why not add a nice wrapper function like `self.rng.next_usize(num_threads)`  that hides all this modulo stuff?

cuviper:
OK, now I'm seeding with a hashed global counter.  The counter gives us a unique start to each PRNG, and hashing that gives us a heap of independent bits.  This will be somewhat reproducible between runs (modulo thread startup races), but we don't really care about that.

Issue #573, opened 2018-05-10
Add support for Range<i128> and Range<u128>


cuviper:
Rust 1.26 stabilizes `i128` and `u128`, so we should be able to support such parallel ranges.  This will have to be only for unindexed `ParallelIterator`, just like the case for 64-bit ranges.

If we do `RangeInclusive` for #572, then we should also consider 128-support there.

We'll have to guard this with a feature flag or a build-script cfg to avoid bumping required rustc.

cuviper:
Done in #575.

PR #575, opened 2018-05-27
u128 and i128 ranges

paulkernfeld:
This is an attempt at #573. I don't have experience with functionality that is conditional on the version of rustc, so I apologize if this PR is completely bonkers. I went with a feature flag in this case, but I would be happy to reimplement this using an alternative method.

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -188,6 +188,9 @@ indexed_range_impl!{isize}
 unindexed_range_impl!{u64}
 unindexed_range_impl!{i64}
 
+// 128-bit support uses a feature to permit rustc < 1.26
+#[cfg(feature = "range128")] unindexed_range_impl !{u128}
+#[cfg(feature = "range128")] unindexed_range_impl !{i128}
 
 #[test]
 pub fn check_range_split_at_overflow() {
--- a/src/range.rs
+++ b/src/range.rs
@@ -203,7 +203,8 @@ pub fn check_range_split_at_overflow() {
 #[cfg(has_i128)]
 #[test]
 pub fn test_i128_len_doesnt_overflow() {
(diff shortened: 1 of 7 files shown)


src/range.rs
```diff
@@ -188,6 +188,9 @@ indexed_range_impl!{isize}
 unindexed_range_impl!{u64}
 unindexed_range_impl!{i64}
 
+// 128-bit support uses a feature to permit rustc < 1.26
+#[cfg(feature = "range128")] unindexed_range_impl !{u128}
+#[cfg(feature = "range128")] unindexed_range_impl !{i128}
```
cuviper: Notice that this macro defines a `fn len(&self) -> u64`, which may overflow for 128-bit ranges.  I think we can just add a second macro parameter for the type of unsigned length, using `u64` for the 64-bit ranges and `u128` for the 128-bit ranges.  A unit test for these lengths would be good too.

src/range.rs
```diff
@@ -185,9 +185,10 @@ indexed_range_impl!{i32}
 indexed_range_impl!{isize}
 
 // other Range<T> with just Iterator
-unindexed_range_impl!{u64}
-unindexed_range_impl!{i64}
-
+unindexed_range_impl!{u64, u64}
+unindexed_range_impl!{i64, u64}
+#[cfg(has_i128)] unindexed_range_impl !{u128, u128}
+#[cfg(has_i128)] unindexed_range_impl !{i128, u128}
```
cuviper: I'm surprised this parsed correctly with the space before `!`, but I'm still going to tweak that so syntax highlighting is correct too.

cuviper:
Looks good!

bors r+

PR #583, opened 2018-07-05
Release rayon 1.0.2 and rayon-core 1.4.1

cuviper:
This updates the test and demo dependencies, then it's about time for
new rayon and rayon-core releases!


cuviper:
Published!

PR #584, opened 2018-07-11
Fix doc links to avoid warning on nightly

LukasKalbertodt:
Hi there!

- one link was just a typo (incorrect `::*`)
- the other one was a wrong intra-link (which are still unstable)

I converted the wrong intra-link to a normal one for now. Would be awesome if this can be merged, as `cargo doc` is quite noisy on nightly when you have many dependencies :)


cuviper:
Thanks!

bors r+

PR #594, opened 2018-09-11
Override `clone_from`.

pthariensflame:
This shouldn't affect most people in any way, but in case somebody *is* using `clone_from` on parallel iterators, it could potentially make it more efficient (it shouldn't make anything *worse*).  **NOTE:** I have not benchmarked this; I don't know how best to do that, so any advice on the topic would be well-received.

This pull request only contains a very basic test for these overrides; advice on how to better test them would also be much appreciated.

### DIFF
--- a/rayon-futures/src/lib.rs
+++ b/rayon-futures/src/lib.rs
@@ -180,6 +180,10 @@ where
     fn clone(&self) -> Self {
         ArcScopeFuture(self.0.clone())
     }
+
+    fn clone_from(&mut self, other: &Self) {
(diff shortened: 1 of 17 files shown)


nikomatsakis:
Based on this, I'm going to close the PR, but thanks @pthariensflame for the suggestion. Feel free to re-open if you think I'm missing something though. :)

pthariensflame:
That’s fine.  My primary reason for believing it might make a difference is that some of these things can delegate down to the `clone_from` of their components, which might be specially optimized (but I’m not sure).

Issue #586, opened 2018-07-31
Consider bumping crossbeam-deque


kpcyrd:
Hey!

I noticed rayon-core currently depends on an old version of crossbeam-deque, Cargo.toml states:
```
# This is deliberately not the latest version, because we want
# to support older rustc than crossbeam-deque 0.3+ does.
[dependencies.crossbeam-deque]
version = "0.2.0"
```

This would require us (the debian rust maintainers) to package an outdated version of crossbeam-deque and its outdated dependencies. Please reconsider bumping the dependency. This currently prevents cc from transitioning to [debian testing](https://piuparts.debian.org/sid/source/r/rust-cc.html).

This seems to be related to https://github.com/rayon-rs/rayon/pull/480

cuviper:
Alternatively, you could just patch your package to use the later version.  AFAIK the API is compatible, at least enough for Rayon's needs, just not the compiler support.  We patch it in Fedora, and it works fine.

(And yes, I'm caught in opposing roles here -- holding conservative support in the crate while using the latest in Fedora...)

kpcyrd:
@cuviper do you have a link for for that patch? It seems rayon-core doesn't build with the latest crossbeam-deque without patching the source as well :/ Thanks!

cuviper:
You're right, we've only bumped it to 0.4 without source changes:
https://src.fedoraproject.org/rpms/rust-rayon-core/blob/master/f/rayon-core-1.4.1-fix-metadata.diff

I think the newer versions have shuffled around the worker/stealer types a bit. I can try to prepare a branch for this next week, if you don't get to it first... 😀

kpcyrd:
At the rayon-rs team: Can you please check which system is currently targeted with >= 1.13.0?

debian stable now ships rustc 1.24.0 and I can't find a distro that ships anything older (openbsd 6.3 ships 1.24.0 as well). This currently makes it difficult to package cc which is required for anything that links to C libs and/or uses nix.

cuviper:
Here's the port: <https://github.com/rayon-rs/rayon/compare/master...cuviper:crossbeam-deque-0.6>.  I haven't done any benchmarking, but at least it builds and passes tests on rust 1.28.

[...]

PR #595, opened 2018-09-13
update crossbeam-deque to 0.6

debris:


debris:
travis fails, cause `crossbeam-utils` cannot be compiled with `rustc 1.13.0`. Is there any reason to be backwards compatible with 2y old rust?

stjepang:
See #586 for a previous discussion on bumping `crossbeam-utils` :)

debris:
haha, thanks :)

cuviper:
Yeah, it's a matter of policy.  And while I'm sure we will bump it eventually, 1.25 is too recent IMO.

Thanks anyway!

PR #602, opened 2018-10-05
Add ParallelIterator::{map_init,for_each_init,try_for_each_init}

cuviper:
These are like `map_with` and all, except using an `init` function for
the paired value instead of a `Send + Clone` value.  The returned type
doesn't have to be `Send` nor `Sync` nor anything else, which makes it
feasible for things like a thread-local RNG.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -130,7 +130,7 @@ pub use self::take::Take;
 mod map;
 pub use self::map::Map;
 mod map_with;
-pub use self::map_with::MapWith;
+pub use self::map_with::{MapWith, MapInit};
 mod zip;
 pub use self::zip::Zip;
 mod zip_eq;
@@ -375,6 +375,42 @@ pub trait ParallelIterator: Sized + Send {
         self.map_with(init, op).for_each(|()| ())
     }
 
+    /// Executes `OP` on a value returned by `init` with each item produced by
+    /// the iterator, in parallel.
+    ///
+    /// The `init` function will be called only as needed for a value to be
+    /// paired with the group of items in each rayon job.  There is no
+    /// constraint on that returned type at all!
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// extern crate rand;
+    /// extern crate rayon;
+    ///
+    /// use rand::Rng;
+    /// use rayon::prelude::*;
+    ///
+    /// let mut v = vec![0u8; 1_000_000];
+    ///
+    /// v.par_chunks_mut(1000)
+    ///     .for_each_init(
+    ///         || rand::thread_rng(),
+    ///         |rng, chunk| rng.fill(chunk),
+    ///     );
+    ///
+    /// // There's a remote chance that this will fail...
+    /// for i in 0u8..=255 {
+    ///     assert!(v.contains(&i));
+    /// }
+    /// ```
+    fn for_each_init<OP, INIT, T>(self, init: INIT, op: OP)
+        where OP: Fn(&mut T, Self::Item) + Sync + Send,
+              INIT: Fn() -> T + Sync + Send,
+    {
+        self.map_init(init, op).for_each(|()| ())
+    }
+
     /// Executes a fallible `OP` on each item produced by the iterator, in parallel.
     ///
     /// If the `OP` returns `Result::Err` or `Option::None`, we will attempt to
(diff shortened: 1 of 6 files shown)


src/iter/mod.rs
```diff
@@ -375,6 +375,42 @@ pub trait ParallelIterator: Sized + Send {
...
+    /// extern crate rayon;
+    ///
+    /// use rand::Rng;
+    /// use rayon::prelude::*;
+    ///
+    /// let mut v = vec![0u8; 1_000_000];
+    ///
+    /// v.par_chunks_mut(1000)
+    ///     .for_each_init(
+    ///         || rand::thread_rng(),
```
nikomatsakis: This seems like a pretty useful API -- and in fact I recently had a use-case where I would have wanted this -- but it annoys me that it feels like this is not *quite* able to model the `_with` variants.  You almost can, because you can invoke

```rust
let value = ...;
foo.for_each_init(
    || value.clone(),
    |value, chunk| ...,
);
```

but it's not *quite* right because you get an extra clone. But to do better would seem to require a more complex protocol -- something mirroring the "split -> fold" pattern of the producer/consumers, I guess. i.e., you would have an initial "seed" value then can be "split" into two seeds and eventually converted into the "actual" value that gets given to the callback.

It feels like we're "dancing around" such an API here and I'm wondering if we should think about whether we can "cut to the chase" (and how to ensure it's ergonomic in common cases).

Let's call it `for_each_seed` for now. I imagine it *might* look something like this:

```rust
trait ParallelSeed {
    type Sprout;
    fn split(&self) -> Self;
    fn sprout(self) -> Self::Sprout;
}

fn for_each_seed<S: ParallelSeed>(
    self,
    seed: S,
    op: impl FnMut(&mut S::Sprout, T),
)
```

Then we could offer some types that implement this trait. For example, `for_each_seed(rayon::Cloned(v), ...)` would be equivalent to `for_each_with(...)`:

```rust
pub struct Cloned<T: Send + Clone>(pub T);

impl ParallelSeed for Cloned<T> {
    type Sprout = T;
    fn split(&self) -> Self { Cloned { value: self.value.clone() } }
    fn sprout(self) -> T { self.value }
}
```

However, I realize now that my variant to model `_init` doesn't *quite* work unless we require the "init" closure to be clone -- this is because the value that is being tracked across threads here is not the *seed* but rather a reference to the seed:

```rust
pub struct PerThread<F>(pub F);

impl<O> ParallelSeed for PerThread<F> 
where F: Fn() -> O
{
    type Sprout = O
    fn split(&self) -> Self {
        // how to implement this?
    }
    fn sprout(self) -> O {
       (self.0)()
    }
}
```

Anyway, do you think it's worth thinking about this general direction?

Maybe not worth holding up this API on that basis.

cuviper: > but it annoys me that it feels like this is not _quite_ able to model the `_with` variants. You almost can, because you can invoke
> 
> ```rust
> let value = ...;
> foo.for_each_init(
>     || value.clone(),
>     |value, chunk| ...,
> );
> ```
> 
> but it's not _quite_ right because you get an extra clone.

This approximation requires `Sync` for the `value`, which the `_with` variants specifically avoid.  You did solve this in your `ParallelSeed for Cloned` though.

> Let's call it `for_each_seed` for now. I imagine it _might_ look something like this:
> 
> ```rust
> trait ParallelSeed {
>     type Sprout;
>     fn split(&self) -> Self;
>     fn sprout(self) -> Self::Sprout;
> }
> 
> fn for_each_seed<S: ParallelSeed>(
>     self,
>     seed: S,
>     op: impl FnMut(&mut S::Sprout, T),
> )
> ```

It does feel like this could be a common underpinning of `_with` and `_init`, but I'm not sure what _new_ examples there would be.  Do you have extension ideas that don't already fit in these two current flavors?

> However, I realize now that my variant to model `_init` doesn't _quite_ work unless we require the "init" closure to be clone -- this is because the value that is being tracked across threads here is not the _seed_ but rather a reference to the seed:

Right, we've generally pushed that indirection internally, like `Map` into `MapProducer` or `MapConsumer`.  We would need another public layer to do that for `ParallelSeed`.

nikomatsakis: > Do you have extension ideas that don't already fit in these two current flavors?

A good question =) 

In answer, I don't have a concrete use case in mind, no. I imagine it would be something where you have some state per-thread state (like this RNG) that you wish to share, but the data you need to construct it from is not `Sync`. Perhaps that is too obscure to be concerned with.

In any case, I don't see the harm in having these two variants. *Even if* we later grew a more flexible variant similar to what I suggested, it will likely be less ergonomic, since it has more knobs =)

That said, I am contemplating whether we can offer some kind of "builder-like" interface that would permit one to compose your own "splitting, sprouting, execution"-like operations. Have to think about it. It'd be nice to be able to stop adding new methods for new patterns (or at least to implement them in terms of some other method).

(Recently, I did encounter a desire to have `map_with` except that I didn't want to invoke `clone` to "split across threads" but rather a custom method. Of course I can implement that with a newtype and the existing `map_with`.)

PR #604, opened 2018-10-09
impl ParallelExtend for tuple pairs

### LINKED ISSUE #600: Implement ParallelExtend for tuples
I am currently working on a project , that needs to recursively split the input data to several categories.
Using ParallelIterator::partition_map i can split the input data into 2, but this is not recursive, 
If for example I return Either\<Either\<A,B>,C> from my iterator, I am unable to partition the data into 
((Vec\<A>,Vec\<B>),Vec\<C>).

 So far I have been unable to implement this on my own, and would appreciate help.

My idea was to implement this trait for tuples, something like this

```

impl<R, S, T> ParallelExtend<T> for (R, S)
    where R: ParallelExtend<T>,
               S: ParallelExtend<T>
    {
        fn par_extend<I>(&mut self, par_iter: I) where I: IntoParallelIterator<Item=Either<R, S>> {
    }
}

```

I need this functionality to partition the input data into several containers  with heterogenous types, and do it as fast as possible.

cuviper:
`ParallelExtend<(A, B)>` and `ParallelExtend<Either<L, R>>` for tuples
behave like `unzip` and `partition_map` respectively.  These allow the
possibility of nested `unzip` and `partition_map` operations, filling
into more than just two collections.  For instance, `(A, (B, C))` items
can be unzipped into `(Vec<A>, (Vec<B>, Vec<C>))`.

Fixes #600.

### DIFF
--- a/src/iter/unzip.rs
+++ b/src/iter/unzip.rs
@@ -24,7 +24,7 @@ trait UnzipOp<T>: Sync + Send {
     }
 }
 
-/// Run an unzip-like operation into `ParallelExtend` collections.
+/// Run an unzip-like operation into default `ParallelExtend` collections.
 fn execute<I, OP, FromA, FromB>(pi: I, op: OP) -> (FromA, FromB)
     where I: ParallelIterator,
           OP: UnzipOp<I::Item>,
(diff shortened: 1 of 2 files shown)


src/iter/unzip.rs
```diff
@@ -386,3 +395,57 @@ impl<A, B, RA, RB> Reducer<(A, B)> for UnzipReducer<RA, RB>
         (self.left.reduce(left.0, right.0), self.right.reduce(left.1, right.1))
     }
 }
+
+
+impl<A, B, FromA, FromB> ParallelExtend<(A, B)> for (FromA, FromB)
```
cuviper: FWIW, the standard library does not have an `Extend` implementation like this, though I think it could.  The value there would not be as high though, as there's significantly more magic involved in the parallel implementation.

src/iter/unzip.rs
```diff
@@ -386,3 +395,57 @@ impl<A, B, RA, RB> Reducer<(A, B)> for UnzipReducer<RA, RB>
...
+{
+    fn par_extend<I>(&mut self, pi: I)
+    where
+        I: IntoParallelIterator<Item = (A, B)>,
+    {
+        execute_into(&mut self.0, &mut self.1, pi.into_par_iter(), Unzip);
+    }
+}
+
+impl<L, R, A, B> ParallelExtend<Either<L, R>> for (A, B)
```
cuviper: I think coherence rules would make this impossible for `Extend` since tuples aren't fundamental -- the `either` crate can't implement the trait for a generic tuple type.  So this is a unique possibility for `ParallelExtend`!

cuviper: Hmm, Rust 1.41's "re-rebalance coherence" would allow this now... :thinking:

PR #605, opened 2018-10-11
Use Travis CI instead of AppVeyor for Windows testing

cuviper:


cuviper:
bors try

cuviper:
bors try

cuviper:
bors try

cuviper:
bors try

cuviper:
bors try

cuviper:
All four windows-msvc builds failed with:

      = note: LINK : fatal error LNK1181: cannot open input file 'advapi32.lib'

Two of the windows-gnu builds passed, but the other two look like they failed to even install the toolchain properly, with no output indicating any reason.

cuviper:
Let's see if Travis has improved anything...

bors try

cuviper:
Yay, it passed!

And importantly, the overall CI run took much less time.  Looking over the [AppVeyor build history](https://ci.appveyor.com/project/cuviper/rayon/history), while these "Try" builds are broken since I removed appveyor.yml, note that the "Merge" builds usually take at least 2 hours.  The [full run on Travis](https://travis-ci.org/rayon-rs/rayon/builds/448996856) took only 44 minutes of wall time, for all of Linux, macOS, and Windows!

(I'm only testing Windows stable and nightly here, which is a little unfair to AppVeyor which has been testing beta too, but that alone is not enough to make up the difference.)

cuviper:
As of #731, we're using GitHub workflows.

PR #617, opened 2018-12-28
Handle the last chunk correctly

yegeun542:
Minor change in ChunksMutProducer. `split_at` function currently panics when the last chunk is shorter than the other chunks.


cuviper:
Presumably this fix is needed for the non-mut `ChunksProducer` too, no?

Did you encounter this issue in a real use case?  Splitting on the end index is unusual, although it is allowed -- we have `tests/producer_split_at.rs` which is supposed to stress-test this.  Can you expand the chunks tests therein to include a case with inexact chunks?

cuviper:
I pushed additional changes myself.

PR #618, opened 2018-12-28
cargo fmt

cuviper:


rayon-demo/src/find/mod.rs
```diff
@@ -72,12 +73,10 @@ macro_rules! make_tests {
             #[bench]
             fn parallel_find_common(b: &mut Bencher) {
                 b.iter(|| {
-                    assert!(
-                        HAYSTACK
-                            .par_iter()
-                            .find_any(|&&x| x[0] % 1000 == 999)
-                            .is_some()
-                    )
+                    assert!(HAYSTACK
```
stjepang: I'm surprised by this line. Perhaps it's a bug in rustfmt?

cuviper: Not sure -- stable (1.31) and nightly are consistent, but 1.30's rustfmt did it the old way.

PR #623, opened 2019-01-13
Filtermap test use sum

seanchen1991:
The `filter_map` test collects the parallel iterator using `sum`, but it collects the sequential iterator using `fold`. These should probably be brought to parity with each other, unless there's some rationale with using `fold` on the sequential iterator that I'm not seeing. 

Also a few typo fixes in READMEs that I corrected locally a while ago. If it's preferable that I submit a separate PR for those, I'm happy to do so.


src/iter/test.rs
```diff
@@ -2104,4 +2104,4 @@ fn check_update() {
     v.par_iter_mut().update(|v| v.push(0)).for_each(|_| ());
 
     assert_eq!(v, vec![vec![1, 0], vec![3, 2, 1, 0]]);
-}
+}
```
cuviper: I don't know why your editor removed the final newline, but let's revert that please.

cuviper: Hmm, it doesn't appear to show in the inline review, but the full github diff displays this change with "No newline at end of file", which is what I want to revert.

cuviper:
Thanks!

bors r+

Issue #607, opened 2018-10-25
Add a parallel equivalent to Iterator::find_map


cuviper:
`Iterator::find_map` is now stable in Rust 1.30:

```rust
fn find_map<B, F>(&mut self, f: F) -> Option<B>
where
    F: FnMut(Self::Item) -> Option<B>, 
```
> Applies function to the elements of iterator and returns the first non-none result.
> 
> `iter.find_map(f)` is equivalent to `iter.filter_map(f).next()`.

For `Iterator::find`, we have variants `ParallelIterator::{find_any,find_first,find_last}`, so the user can be precise whether they care about the order.  We should probably do similar for `find_map`, with some bikeshedding required... `find_map_any`?  `find_any_map`?

cuviper:
I think in our case, the easiest implementation would be something like:
`iter.map(f).find_any(Option::is_some).map(Option::unwrap)`

cuviper:
The actual implementation in `Iterator::find_map` is:

```rust
    fn find_map<B, F>(&mut self, mut f: F) -> Option<B> where
        Self: Sized,
        F: FnMut(Self::Item) -> Option<B>,
    {
        self.try_for_each(move |x| {
            match f(x) {
                Some(x) => LoopState::Break(x),
                None => LoopState::Continue(()),
            }
        }).break_value()
    }
```

We could do similar if we invent/copy a similar internal `LoopState` type, but that only works for the `any` variant.  We'd still need something ordered for `first`/`last` variants.

cuviper:
`iter.filter_map(f).find_any(|_| true)` should work too, closer to what the docs suggest.

seanchen1991:
I'm taking a stab at implementing this. @cuviper I was thinking of implementing `find_map` on top of `find_first`. With `find_any` it would seem like it would kind of defeat the purpose of what the sequential version of `find_map` tries to accomplish according to the docs. 

So basically `iter.filter_map(f).find_first(|x| x.is_some())`.

Does that sound reasonable?

seanchen1991:
Just submitted #627 in reference to this issue.

cuviper:
Done in #627.

PR #627, opened 2019-01-20
Add `find_map` with `any`, `first`, and `last` variants

seanchen1991:
PR opened in response to #607.


cuviper:
FYI, GitHub doesn't always send notifications when you push new commits, so it helps to leave a comment. I also amended your commit with `cargo fmt`.

bors r+

Issue #510, opened 2018-01-20
Implement `par_exact_chunks/_mut`; slice chunks with exact size.


Kerollmops:
As new methods as been added to the slice type. I propose the add of these new methods too.
These iterators seems to be really useful to llvm and its optimizations.

Here is the [tracking issue for exact_chunks/_mut; slice chunks with exact size](https://github.com/rust-lang/rust/issues/47115).

cuviper:
(I labelled this `parity-std::iter`, but that's really `std::slice` -- maybe we need to use more labels...)

That addition sounds fine to me, but I'd prefer to wait for it to stabilize in `std` before we mimic it.

cuviper:
Help is welcome, if this is a feature that interests you!

sdroege:
Might also be interesting to have the `rchunks_*` iterators available from `rayon` in addition to this to have more API-equivalence between `std` and `rayon`

cuviper:
I just merged #629 for `par_chunks_exact`/`_mut` and filed #734 for `rchunks`.

PR #629, opened 2019-01-25
Implement chunks_exact and chunks_exact_mut

zesterer:
Follow-up PR from #510. `chunks_exact` (and its mutable counterpart) are not yet implemented on `ParallelIterator`, only `ParallelSlice`.

### DIFF
--- a/src/slice/mod.rs
+++ b/src/slice/mod.rs
@@ -105,9 +105,10 @@ pub trait ParallelSlice<T: Sync> {
     /// ```
     fn par_chunks_exact(&self, chunk_size: usize) -> ChunksExact<'_, T> {
         assert!(chunk_size != 0, "chunk_size must not be zero");
-        let rem = self.as_parallel_slice().len() % chunk_size;
-        let len = self.as_parallel_slice().len() - rem;
-        let (fst, snd) = self.as_parallel_slice().split_at(len);
+        let slice = self.as_parallel_slice();
+        let rem = slice.len() % chunk_size;
+        let len = slice.len() - rem;
+        let (fst, snd) = slice.split_at(len);
         ChunksExact {
             chunk_size: chunk_size,
             slice: fst,
@@ -193,9 +194,10 @@ pub trait ParallelSliceMut<T: Send> {
     /// ```
     fn par_chunks_exact_mut(&mut self, chunk_size: usize) -> ChunksExactMut<'_, T> {
         assert!(chunk_size != 0, "chunk_size must not be zero");
-        let rem = self.as_parallel_slice_mut().len() % chunk_size;
-        let len = self.as_parallel_slice_mut().len() - rem;
-        let (fst, snd) = self.as_parallel_slice_mut().split_at_mut(len);
+        let slice = self.as_parallel_slice_mut();
+        let rem = slice.len() % chunk_size;
+        let len = slice.len() - rem;
+        let (fst, snd) = slice.split_at_mut(len);
         ChunksExactMut {
             chunk_size: chunk_size,
             slice: fst,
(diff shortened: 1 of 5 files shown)


src/slice/mod.rs
```diff
@@ -557,6 +611,67 @@ impl<'data, T: Sync + 'data> IndexedParallelIterator for Chunks<'data, T> {
...
+impl<'data, T: Sync> ChunksExact<'data, T> {
+    /// Return the remainder of the original slice that is not going to be
+    /// returned by the iterator. The returned slice has at most `chunk_size-1`
+    /// elements.
+    pub fn remainder(&self) -> &'data [T] {
+        self.rem
+    }
+}
+
+impl<'data, T: Sync> Clone for ChunksExact<'data, T> {
```
sdroege: Isn't this exactly what `#[derive(Clone)]` would generate?

src/slice/mod.rs
```diff
@@ -557,6 +611,67 @@ impl<'data, T: Sync + 'data> IndexedParallelIterator for Chunks<'data, T> {
...
+
+    fn len(&self) -> usize {
+        self.slice.len() / self.chunk_size
+    }
+
+    fn with_producer<CB>(self, callback: CB) -> CB::Output
+    where
+        CB: ProducerCallback<Self::Item>,
+    {
+        callback.callback(ChunksProducer {
```
sdroege: Might be worth checking in the output of the compiler if all the size checks are optimized away here (i.e. that every chunk has exactly `chunk_size` elements and not potentially fewer). Otherwise a `ChunksExactProducer` might be needed.

src/slice/mod.rs
zesterer: Yes. I have no idea why I wrote this, I must have been half asleep at the time.

cuviper: I think `#[derive(Clone)]` would also force `T: Clone`, which is not needed to clone the slice.

sdroege: I think you're both right, sorry :)

src/slice/mod.rs
cuviper: @zesterer Have you been able to check this?

zesterer: Apologies, I've relatively busy over the past few weeks. I'll have time to pick this up again within a few days.

seijikun:
Hey guys, is there any news on this?
I'm currently depending on @zesterer's branch with my crate ([kmeans](https://github.com/seijikun/kmean-rs)), because its already delivering a speedup by a factor of ~4x in its current form for me (as compared to `par_bridge()`).

I tried having a look at the produced asm code, but couldn't figure out if the `chunk-size`-check was still in there.

cuviper:
@seijikun I'm not surprised that it would perform better than `par_bridge`, but have you compared to the inexact `par_chunks` / `par_chunks_mut`?

seijikun:
@cuviper Ah, totally forgot about that. I wasn't able to use it when I decided to use this branch (because of the remainder). Refactored the code to manually pad the data (so that there is no other-sized remainder).
I now barely see a difference between `par_chunks()` and `par_chunks_exact()`.

cuviper:
That's a good point -- the remainder API has value regardless of size-check optimizations.

src/slice/mod.rs
```diff
@@ -775,6 +890,61 @@ impl<'data, T: Send + 'data> IndexedParallelIterator for ChunksMut<'data, T> {
...
+    chunk_size: usize,
+    slice: &'data mut [T],
+    rem: &'data mut [T],
+}
+
+impl<'data, T: Send> ChunksExactMut<'data, T> {
+    /// Return the remainder of the original slice that is not going to be
+    /// returned by the iterator. The returned slice has at most `chunk_size-1`
+    /// elements.
+    pub fn into_remainder(self) -> &'data mut [T] {
```
cuviper: This is awkward to consume `self`, because it means the rest of the `ParallelIterator` can't be used at all. With `Iterator` you can at least pull the items out first without consuming `self`, with `next`, `by_ref`, etc., but not so in rayon. All of the parallel methods consume `self`, so you can't do that *and* get the remainder here.

Maybe we can keep `into_remainder` for `std` parity, but also add something non-consuming like:

```rust
    pub fn remainder(&mut self) -> &mut [T] {
        self.rem                // ^ lifetime borrowed from self, not 'data
    }

    pub fn take_remainder(&mut self) -> &'data mut [T] {
        mem::replace(&mut self.rem, &mut [])
    }
```

Coder-256:
Is this still being worked on?

src/slice/mod.rs
cuviper: I've rebased the PR on master, which now has a 1.31 baseline, so I also added the custom producers that ultimately use `std::iter::ChunksExact`[`Mut`]. That _should_ punt the performance question to `std`...

cuviper:
I think this is ready now -- thanks for the PR, @zesterer!

bors r+

Issue #612, opened 2018-11-14
Establish a process to maintain code formatting


cuviper:
#611 is doing a one-time cleanup with `rustfmt`, but we should establish some kind of process to make sure we keep it tidy.

This is also being discussed for Rust itself, but there's no best practice established yet:
https://internals.rust-lang.org/t/running-rustfmt-on-rust-lang-rust-and-other-rust-lang-repositories/8732

cuviper:
#731 added a formatting check to CI.

PR #631, opened 2019-02-03
Specialize more `consume_iter`s

huonw:
This specializes the remaining "obvious" `consume_iter`s for the second part of #465. Benchmarks are included in each commit (the first specializes the types that can just defer to a `std` API, the second two require a bit more work); summary (on a 6 core/12 thread machine):

- on the included benchmarks:
  - `find` is up to 1.5x faster (`find::size1::parallel_find_middle`)
  - `collect` is up to 4x faster (`vec_collect::vec_i_filtered::with_collect`)
- using basic program that does something like `(0..std::u32::MAX).into_par_iter().<OPERATION>.map(test::black_box).count()`
  - with no operation (i.e. just testing `map`/`count`): 8x faster
  - with `.map(Some).while_some()`: 3x faster
  - with `.intersperse(1)`: 10x faster

The remaining types without a specialization are:

```
$ rg --files-with-matches 'fn consume' src/iter | xargs rg --files-without-match 'fn consume_iter'
src/iter/filter_map.rs
src/iter/unzip.rs
src/iter/try_fold.rs
src/iter/try_reduce_with.rs
src/iter/try_reduce.rs
src/iter/filter.rs
src/iter/flat_map.rs
src/iter/find_first_last/mod.rs
src/iter/collect/consumer.rs
```

(I've started a WIP branch for the `try_*` types at https://github.com/huonw/rayon/tree/try_fold; diff to this PR: https://github.com/huonw/rayon/compare/iter-opts...try_fold.)

### DIFF
--- a/src/iter/cloned.rs
+++ b/src/iter/cloned.rs
@@ -205,6 +205,14 @@ where
         }
     }
 
+    fn consume_iter<I>(mut self, iter: I) -> Self
+    where
+        I: IntoIterator<Item = &'a T>
+    {
+        self.base = self.base.consume_iter(iter.into_iter().cloned());
+        self
+    }
+
     fn complete(self) -> F::Result {
         self.base.complete()
     }
--- a/src/iter/cloned.rs
+++ b/src/iter/cloned.rs
@@ -207,7 +207,7 @@ where
 
     fn consume_iter<I>(mut self, iter: I) -> Self
     where
-        I: IntoIterator<Item = &'a T>
+        I: IntoIterator<Item = &'a T>,
     {
         self.base = self.base.consume_iter(iter.into_iter().cloned());
         self
--- a/src/iter/filter.rs
+++ b/src/iter/filter.rs
@@ -136,14 +136,9 @@ where
         }
     }
 
-    fn consume_iter<I>(mut self, iter: I) -> Self
-    where
-        I: IntoIterator<Item = T>
-    {
-        self.base = self.base.consume_iter(iter.into_iter().filter(self.filter_op));
-        self
-    }
-
+    // This cannot easily specialize `consume_iter` to be better than
+    // the default, because that requires checking `self.base.full()`
+    // during a call to `self.base.consume_iter()`. (#632)
 
     fn complete(self) -> Self::Result {
         self.base.complete()
--- a/src/iter/filter.rs
+++ b/src/iter/filter.rs
@@ -136,6 +136,15 @@ where
         }
     }
 
+    fn consume_iter<I>(mut self, iter: I) -> Self
+    where
+        I: IntoIterator<Item = T>
+    {
+        self.base = self.base.consume_iter(iter.into_iter().filter(self.filter_op));
+        self
+    }
+
+
     fn complete(self) -> Self::Result {
         self.base.complete()
     }
--- a/src/iter/filter_map.rs
+++ b/src/iter/filter_map.rs
@@ -139,13 +139,10 @@ where
             self
         }
     }
-    fn consume_iter<I>(mut self, iter: I) -> Self
-    where
-        I: IntoIterator<Item = T>
-    {
-        self.base = self.base.consume_iter(iter.into_iter().filter_map(self.filter_op));
-        self
-    }
+
+    // This cannot easily specialize `consume_iter` to be better than
+    // the default, because that requires checking `self.base.full()`
+    // during a call to `self.base.consume_iter()`. (#632)
 
     fn complete(self) -> C::Result {
         self.base.complete()
--- a/src/iter/filter_map.rs
+++ b/src/iter/filter_map.rs
@@ -139,6 +139,13 @@ where
             self
         }
     }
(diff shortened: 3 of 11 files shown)


huonw:
Hi @cuviper and/or @nikomatsakis, I'd love to see this progress, so a review would be appreciated. Thanks!

cuviper:
> 1.5x, 4x, 8x, 3x, 10x

Woah! I expected only modest gains available here, not entire multiples. Thanks for doing this!

I think we need to be a little more careful about checking `full` flags though, for those that may process an arbitrary number of items before anything actually reaches the base consumer. This applies to `filter`, `filter_map`, and `fold`, as well as `find_any` needing to check its global state. In `while_some` you do have a load of its global state, which is good, but it also needs to consult whether its base is full.

For a tangible (though exaggerated) example of this, I tried adding this in tests/octillion.rs:

```rust
#[test]
fn find_any_octillion() {
    let x = octillion().find_any(|&x| x > OCTILLION / 2);
    assert!(x.is_some());
}
```

On the master branch, this takes only a couple milliseconds, even in debug mode, as thread creation itself is the only real work. The higher splits will instantly find a match, short-circuiting the lower splits. (Assuming at least two threads -- if we really add this test, we should guarantee that like some of the other tests in that file do.)

On your branch, the lower splits continue through their entire iterator, without looking up to see that the shared `found` state was set. I think this can be solved using a `take_while`, and similarly in `filter` etc. We'll give up some performance, sadly, but that seems necessary for short-circuiting to work correctly. Hopefully codegen can at least inline and const-prop the cases where `full` is always false.

huonw:
That's a good point! Fixing it for `find` and `fold` should be easy enough, but I can't quite see how to handle it for `filter`/`filter_map`: `self.base` is moved into the `consume_iter` call, and thus there's no way to call its `full` in the iterator adaptor chain. I may have to revert them.

cuviper:
> but I can't quite see how to handle it for `filter`/`filter_map`: `self.base` is moved into the `consume_iter` call, and thus there's no way to call its `full` in the iterator adaptor chain.

That's annoying. You may be able to get by with `take_while().filter().next()`, pulling one item at a time to pass to the consumer. I guess the performance compared to plain `consume` will depend on how many items the `filter` is actually filtering out. If you do revert, please put a comment in there so we remember these considerations.

> I don't think that concern applies to `while_some` because it isn't consuming an arbitrary number of elements for each output,

Ah, yes, you're right. It does need to load its own global flag as you're doing, since the base consumer won't check that, but the base can otherwise take care of itself.

cuviper:
> Should I also open an issue along the lines of "`Folder` API doesn't allow for internal iteration optimisation of `filter`-like adapters", to remember it for a potential (rayon 2?) API change/iteration/improvement?

Sure, although I'm not sure what that change would look like, so if you have ideas please sketch them out. You'd still have the same problem if `consume_iter` took `&mut self`, for instance. Or the folder could have a method to return some associated `type Full` object, but this couldn't reference the folder's own state.

huonw:
I had the same thought about using an associated type, and I think not being able to reference the folder's state is an ok limitation: most things either are never full or need to communicate with other threads to check full-ness anyway. I opened https://github.com/rayon-rs/rayon/issues/632.

huonw:
I added a commit that:

- reverts the change for `filter` and `filter_map`
- adds `take_while` to `fold` and `find_any`
- extends the octillion tests to cover the above

Unsurprisingly, this makes a program like the following slightly slower compared to the direct `consume_iter` without a check, but it's still faster than the non-specialized form.

```rust
#![feature(test)]

extern crate test;
extern crate rayon;
use rayon::prelude::*;

fn main() {
    let count = (0..std::u32::MAX)
        .into_par_iter()
        .map(test::black_box)
        .find_any(|_| test::black_box(false));
    println!("{:?}", count);
}
```

```
$ hyperfine ./find-original ./find-no-check ./find-check
Benchmark #1: ./find-original
  Time (mean ± σ):     627.6 ms ±  25.7 ms    [User: 7.130 s, System: 0.014 s]
  Range (min … max):   588.4 ms … 656.4 ms    10 runs

Benchmark #2: ./find-no-check
  Time (mean ± σ):     481.5 ms ±  10.8 ms    [User: 5.415 s, System: 0.013 s]
  Range (min … max):   468.9 ms … 498.2 ms    10 runs

Benchmark #3: ./find-check
  Time (mean ± σ):     562.3 ms ±  11.8 ms    [User: 6.363 s, System: 0.013 s]
  Range (min … max):   542.5 ms … 578.2 ms    10 runs
```

(find-original = without specialization; find-no-check = custom `consume_iter` without `take_while`; find-check = this commit)

huonw:
Done (formatting isn't checked on CI?).

Let me know if you'd like the commits squashed, for a clean history.

cuviper:
> Done (formatting isn't checked on CI?).

No, we need to figure that out... #612 

> Let me know if you'd like the commits squashed, for a clean history.

This is clean enough for me, and there are important details in your "check full" commit.

bors r+

PR #635, opened 2019-02-14
Allow the deprecated ATOMIC_USIZE_INIT

cuviper:
Nightly 1.34 suggests the const `AtomicUsize::new()` instead, but we
still want to support older compilers, so just allow it for now.


cuviper:
This is a trivial change...

bors r+

PR #640, opened 2019-02-28
Reexport current_thread_index function

### LINKED ISSUE #639: Getting thread number
To avoid reallocating a big at each turn inside my parallel iterator, I'd like to instantiate a big `Vec<Vec<_>>` outside so I can then just send the corresponding `Vec` into my function. However, it doesn't seem like I can get the current thread number since it requires to have a `ThreadPool` object and there is no way to have one. Maybe I'm missing something though...

I see two ways of having access to this information:
 * Adding a function `current_thread_index` which doesn't need to be run on a `ThreadPool`
 * Providing a closure which pass the `ThreadPool` object (or directly the thread number)

What do you think of this?

GuillaumeGomez:
Fixes #639.

### DIFF
--- a/rayon-core/tests/stack_overflow_crash.rs
+++ b/rayon-core/tests/stack_overflow_crash.rs
@@ -15,6 +15,19 @@ fn force_stack_overflow(depth: u32) {
     }
 }
 
+#[cfg(unix)]
+fn overflow_code() -> Option<i32> {
+    None
+}
+
+#[cfg(windows)]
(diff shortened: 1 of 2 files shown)


cuviper:
Thanks!

bors r+

cuviper:
Looks like `overflowing_literals` are now denied. I've hopefully fixed this one.

bors r+

GuillaumeGomez:
Ah? Interesting! Well, we'll see. :)

PR #615, opened 2018-12-12
Implement RFC #1: FIFO spawns

cuviper:
This implements rayon-rs/rfcs#1, adding `spawn_fifo`, `scope_fifo`, and `ScopeFifo`, and deprecating the `breadth_first` flag.

Fixes #590.
Closes #601.

PR #641, opened 2019-03-02
Require rayon-core ^1.4.1

kornelski:
I'm trying to make my crates work with `cargo -Z minimal-versions` to ensure that versions specified in my Cargo.toml are actual versions that work with my crate.

The problem is, it's not true for Rayon's dependencies: rayon 1.0.3 uses rayon-core 1.4.0, which uses rand 0.3.12, which uses advapi32-sys 0.1.2, which uses winapi 0.0.1, which isn't compatible with Rust 1.0.

The problem has been fixed in rayon-core 1.4.1, so that should be the minimal version used by rayon.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ members = ["rayon-demo", "rayon-core", "rayon-futures"]
 exclude = ["ci"]
 
 [dependencies]
-rayon-core = { version = "1.4", path = "rayon-core" }
+rayon-core = { version = "1.4.1", path = "rayon-core" }
 crossbeam-deque = "0.2.0"
 
 # This is a public dependency!


cuviper:
bors r+

I'm approving this, because it's harmless, but I do feel some objection to this in principle. The semantic requirement is correct at 1.4, and it's not our fault (nor yours) that deeper indirect dependencies got their minimum wrong. IMO it's problematic that `-Z minimal-versions` makes you worry about dependencies outside your control -- it might be better if it only pinned the minimum of your workspace's dependencies.

Practically speaking, we're going to need a synchronized version bump of rayon and rayon-core before we publish anyway, because there are new re-exports from #615.

Issue #638, opened 2019-02-23
Parallel iterators should just fail early in case of panic instead of continuing to issue more workers


pedrocr:
When using a parallel iterator if a panic happens in one of the threads that doesn't stop more threads from being launched. This is a surprising behavior since in the equivalent sequential code the loop will terminate early because of the panic. This can lead to DoS in code like this:

```rust
extern crate rayon;
use self::rayon::prelude::*;

fn main() {
    let width  = 10;
    let height = 10;

    let mut out: Vec<u16> = vec![0u16; width*height];
    out.par_chunks_mut(width).enumerate().for_each(|(row, _line)| {
        panic!("panic at row {}", row);
    });
}
```

When `height` is very large it takes forever to issue panics for all the lines. I hit this when fuzzing a binary image format where a well crafted file would take advantage of this to force the code to go through the slow panic processing path a few million times and cause a DoS. The bug is introduced by the parallel iterator as the sequential code would bail at the first panic.

pedrocr:
This on the other hand works as expected:

```rust
extern crate rayon;
use self::rayon::prelude::*;
use std::panic;

fn main() {
    let width  = 10;
    let height = 10;

    let mut out: Vec<u16> = vec![0u16; width*height];
    out.par_chunks_mut(width).enumerate().try_for_each(|(row, _line)| {
        panic::catch_unwind(|| panic!("panic at row {}", row))
    }).ok();
}
```

So the mechanism to abort early already exists for the `try_for_each` variant, it's just not used for panics in `for_each`.

cuviper:
At the low level in `rayon-core`, where panics are caught and propagated, it doesn't know about parallel iterators or what jobs are related. It can only see a connection in that iterators use `join`, but we have explicitly [documented its panic behavior](https://docs.rs/rayon-core/1.4.1/rayon_core/fn.join.html#panics) like so:

> No matter what happens, both closures will always be executed. If a single closure panics, whether it be the first or second closure, that panic will be propagated and hence join() will panic with the same panic value. If both closures panic, join() will panic with the panic value from the first closure.

The implementation of `try_for_each` maintains a shared `AtomicBool` across all threads to track its completion state. I think that's too much synchronization to force on `for_each` and all other parallel iterators.

[...]

PR #644, opened 2019-03-15
Add ParallelIterator::panic_fuse()

cuviper:
This wraps an iterator with a fuse in case of panics, to halt all
threads as soon as possible. While we always propagate panics anyway,
that doesn't usually synchronize across threads, so iterators may end up
processing many more items before the panic is realized. With
`panic_fuse()`, a shared `AtomicBool` is checked before processing each
item, and set for any panic unwinding in its path.

### DIFF
--- a/src/iter/panic_fuse.rs
+++ b/src/iter/panic_fuse.rs
@@ -1,0 +1,338 @@
+use super::plumbing::*;
+use super::*;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::thread;
+
+/// `PanicFuse` is an adaptor that wraps an iterator with a fuse in case
+/// of panics, to halt all threads as soon as possible.
+///
+/// This struct is created by the [`panic_fuse()`] method on [`ParallelIterator`]
+///
+/// [`panic_fuse()`]: trait.ParallelIterator.html#method.panic_fuse
+/// [`ParallelIterator`]: trait.ParallelIterator.html
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+#[derive(Debug, Clone)]
+pub struct PanicFuse<I: ParallelIterator> {
+    base: I,
+}
+
+/// Helper that sets a bool to `true` if dropped while unwinding.
+#[derive(Clone)]
+struct Fuse<'a>(&'a AtomicBool);
+
+impl<'a> Drop for Fuse<'a> {
+    #[inline]
+    fn drop(&mut self) {
+        if thread::panicking() {
+            self.0.store(true, Ordering::Relaxed);
+        }
+    }
+}
+
+impl<'a> Fuse<'a> {
+    #[inline]
+    fn panicked(&self) -> bool {
+        self.0.load(Ordering::Relaxed)
+    }
+}
+
+/// Create a new `PanicFuse` iterator.
+///
+/// NB: a free fn because it is NOT part of the end-user API.
+pub fn new<I>(base: I) -> PanicFuse<I>
+where
+    I: ParallelIterator,
+{
+    PanicFuse { base: base }
+}
+
+impl<I> ParallelIterator for PanicFuse<I>
+where
+    I: ParallelIterator,
+{
+    type Item = I::Item;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+    where
+        C: UnindexedConsumer<Self::Item>,
+    {
+        let panicked = AtomicBool::new(false);
+        let consumer1 = PanicFuseConsumer {
+            base: consumer,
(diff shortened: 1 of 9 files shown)


cuviper:
FWIW, I think we could also do this implicitly at a coarse level without too much overhead. In the `bridge` functions where they check `consumer.full()`, they could also check for pending panics with the same `Fuse` I've used here. This would give us at least *some* short-circuiting behavior, without inhibiting optimizations on any inner serial loops. This fine-grained `panic_fuse()` may still be useful for the really paranoid users though.

cc #638 @pedrocr

pedrocr:
Thanks, I'll benchmark this with my code. The other test showed this as not being a performance regression at all (slight improvement actually). If that continues wouldn't it make sense to just make it default?

cuviper:
The added check will inhibit some optimizations, though it might not matter for your workload.

```rust
#![feature(test)]

extern crate rayon;
extern crate test;

use rayon::prelude::*;

const ITER: std::ops::Range<u64> = 0..100_000_000;

#[bench]
fn sum_raw(b: &mut test::Bencher) {
    b.iter(|| ITER.into_par_iter().sum::<u64>());
}

#[bench]
fn sum_fused(b: &mut test::Bencher) {
    b.iter(|| ITER.into_par_iter().panic_fuse().sum::<u64>());
}
```

```text
test sum_fused ... bench:  27,540,693 ns/iter (+/- 1,598,486)
test sum_raw   ... bench:      35,565 ns/iter (+/- 9,968)
```

If I increase the size of this range, `sum_fused` increases linearly, while `sum_raw` barely budges. It turns out that LLVM knows how to reduce such sums to an expression -- `∑ 1..=n = n*(n+1)/2`, and sub-ranges are simple from that. Most of the time spent in `sum_raw` is just parallel overhead, whereas in serial this would probably optimize to a constant.

So that's a ridiculous example, but it still shows that the fuse gets in the way. A more meaningful case might involve optimizations like auto-vectorization (SIMD).

src/iter/panic_fuse.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    fn opt_len(&self) -> Option<usize> {
+        self.base.opt_len()
+    }
+}
+
+impl<I> IndexedParallelIterator for PanicFuse<I>
+where
+    I: IndexedParallelIterator,
```
cuviper: I'm not sure we should really have this, nor `opt_len()`. It's _safe_, but consumers like `collect_vec()` are going to fail their own assertions if they come up short. If nothing else, that means we may have competing panics.

That's also a problem for the idea of implicitly fusing in the `bridge` functions.

pedrocr:
I wouldn't expect it would be possible to do those optimizations across a `.join()`. But if this is the case then indeed there's no way to make this default. I'll test for my workload then.

cuviper:
It doesn't have to optimize *across* join. At the bottom of all of the recursive joins, you just have a handful of `m..n` sums, which can each be computed in constant time. Adding a conditional break on the fuse disrupts that, making it linear again.

pedrocr:
Ah, that explains it then. In my code the body of the loop is fully independent from each other. There's nothing to do after join.

pedrocr:
For my code this is about just as good as the manual `try_for_each` version:

http://scratch.corujas.net/rayon-iterator-panic-bench/

For some reason it also has a bit of a performance gain.

src/iter/panic_fuse.rs
nikomatsakis: Interesting point. But this is true also for the original, right? i.e., if you do `par_iter().map(|x| panic!()).collect()`, it'll have the same problem, no?

cuviper: The difference in yours is that individual `CollectConsumer` splits will either be fully completed, or they'll be panic-unwound without getting a chance to `complete()` and notice that they're short. Same for the overall `Collect::complete()`, it will just be unwound.

But with `panic_fuse()`, when one consumer panics, the other remaining consumers will return normally, and then `CollectFolder::complete()` has a chance to notice the shortfall and raise a panic of its own.

I guess we could remove the "too few" assertion in `CollectFolder::complete()` and just leave it to the overall `Collect::complete()` to check the total. That last one should still be skipped by the fuse's panic unwinding.

cuviper: > I guess we could remove the "too few" assertion in `CollectFolder::complete()` and just leave it to the overall `Collect::complete()` to check the total. That last one should still be skipped by the fuse's panic unwinding.

I've implemented this, with new tests to ensure we get the panic we want.

cuviper:
bors r=nikomatsakis

PR #647, opened 2019-03-26
impl ParallelIterator for RangeInclusive

### LINKED ISSUE #572: Add support for RangeInclusive
Rust 1.26 stabilizes [`RangeInclusive`](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html) but none of its methods or fields. However, I believe you could hack it with `next()` and  `next_back()` / `last()` (yuck).

cuviper:
All the grunt work is forwarded to `range::Iter`, with fixups for the
end case as needed.

Fixes #572.

### DIFF
--- a/src/range_inclusive.rs
+++ b/src/range_inclusive.rs
@@ -1,0 +1,203 @@
+//! Parallel iterator types for [inclusive ranges][std::range],
+//! the type for values created by `a..=b` expressions
+//!
+//! You will rarely need to interact with this module directly unless you have
+//! need to name one of the iterator types.
+//!
+//! ```
+//! use rayon::prelude::*;
+//!
+//! let r = (0..=100u64).into_par_iter()
+//!                     .sum();
+//!
+//! // compare result with sequential calculation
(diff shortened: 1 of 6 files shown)


cuviper:
I've rebased this for 1.26 after #656 -- in particular, we don't need build-script detection anymore.

src/range_inclusive.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+macro_rules! convert {
+    ( $self:ident . $method:ident ( $( $arg:expr ),* ) ) => {
+        if let Some((start, end)) = $self.bounds() {
+            if let Some(end) = end.checked_add(1) {
+                (start..end).into_par_iter().$method($( $arg ),*)
+            } else {
+                (start..end).into_par_iter().chain(once(end)).$method($( $arg ),*)
```
nikomatsakis: cute

sunjay:
Any chance of releasing a new version of rayon so this feature can be used?

cuviper:
#661 is the release PR.

PR #653, opened 2019-04-10
Remove superfluous use of atomics in Collect

nox:
There is no need to track the global writes with an atomic usize,
just make the CollectConsumer fold and reduce to the number of writes.

### DIFF
--- a/src/iter/collect/mod.rs
+++ b/src/iter/collect/mod.rs
@@ -1,7 +1,6 @@
 use super::{IndexedParallelIterator, IntoParallelIterator, ParallelExtend, ParallelIterator};
 use std::collections::LinkedList;
 use std::slice;
-use std::sync::atomic::{AtomicUsize, Ordering};
 
 mod consumer;
 use self::consumer::CollectConsumer;
@@ -19,8 +18,10 @@ where
 {
     v.truncate(0); // clear any old data
     let mut collect = Collect::new(v, pi.len());
-    pi.drive(collect.as_consumer());
-    collect.complete();
+    let writes = pi.drive(collect.as_consumer());
+    unsafe {
+        collect.complete(writes);
+    }
 }
 
 /// Collects the results of the iterator into the specified vector.
@@ -40,8 +41,10 @@ where
     T: Send,
 {
     let mut collect = Collect::new(v, len);
-    pi.drive_unindexed(collect.as_consumer());
-    collect.complete();
+    let writes = pi.drive_unindexed(collect.as_consumer());
+    unsafe {
+        collect.complete(writes);
+    }
 }
 
 /// Unzips the results of the exact iterator into the specified vectors.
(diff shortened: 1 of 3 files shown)


src/iter/collect/mod.rs
```diff
@@ -19,8 +18,10 @@ where
 {
     v.truncate(0); // clear any old data
     let mut collect = Collect::new(v, pi.len());
-    pi.drive(collect.as_consumer());
-    collect.complete();
+    let writes = pi.drive(collect.as_consumer());
```
cuviper: My only hesitance is that tracking `writes` correctly is very important for safety, but this PR moves that value through the control of a generic unknown third party, `pi`.

I'm not sure how bad that is. `pi.drive` has to be generic in the consumer type, which means it's generic in the `C::Result` type too, so it doesn't *know* that we're expecting a `usize`. Thus it can't directly misbehave by feeding us bad values. I think the worst *innocent* bug would be neglecting to reduce the results from two consumer splits, and then we'd just panic for not having enough writes.

A nefarious `pi` might be able to cheat in other ways, like `unsafe` code, or maybe even safe specialization on the `C::Result` type. But I doubt this could be done on accident, and we probably don't need to worry about whether someone actively shoots their own foot.

So... maybe it's fine, but I wanted to raise the point and give it some thought.

cuviper:
Did you do any benchmarking? The atomics are only updated upon the folder `complete()`, so I expect this won't have much impact on performance with or without, but maybe on larger machines. In rayon-demo, the `vec_collect::vec_i::*` benchmarks should exercise the indexed collect.

nox:
I'll check that out. I didn't do any benchmarking, I just don't see the need to use atomics when we don't need them. The fact that `Collect` was using atomics when it didn't need to made me quite worried that `rayon_croissant` was just unfeasible because it made me wonder if my understanding of the `Folder` and `Reducer` traits was actually correct.

src/iter/collect/mod.rs
nikomatsakis: So, this is admittedly a largely theoretical concern, but I think it's a real one, and I'm inclined not to make this change for this reason.

As you note, it's certainly possible for `pi` to have bugs and to return the wrong value to us, whether that be by failing to reduce properly or by using specialization. 

I agree it seems unlikely but those unlikely circumstances are exactly the ones that take the longest to track down when they happen. =) 

I could imagine trying to declare the traits to be unsafe, but it'd be a real shame. I remember the bad old days when rayon had unsafe all over the place: getting rid of that was always a nice benefit.

If there is a performance concern, I suspect we could create a "tree" of atomic integers and recoup similar benefits that way, without requiring as must trust.

cuviper:
OK, I'm going to close this on the safety concern. I was just making changes in this area for `panic_fuse()`, so I also added a comment of justification for the atomic, commit <https://github.com/rayon-rs/rayon/pull/644/commits/7d99416f7dc9eeecbbcf6f14b7cbbc46e99881c7>.

src/iter/collect/mod.rs
nox: I'm confused as to what can be specialised to make bad things happen here. Is it about specialising with `usize` consumer results? If so, can't we just make `Collect` use a private wrapper around `usize`? You need to be able to name the things for which you want to specialise, right?

PR #656, opened 2019-04-26
Rust 1.26 and assorted cleanups

### LINKED ISSUE #586: Consider bumping crossbeam-deque

cuviper:
In anticipation of [RFC 3](https://github.com/rayon-rs/rfcs/pull/3), giving us a policy to soon raise our minimum to Rust 1.26, I started this branch to make the conversion. After the primary motivation of upgrading crossbeam, I also looked for ways to use newer language features.

Throughout all of this, the public API should be completely unaffected.

Closes #586 for newer crossbeam, even though we're not quite at the *newest*.


cuviper:
We merged RFC 3, and @nikomatsakis said, "I'm good with merging #656", so...

bors r=nikomatsakis

cuviper:
bors r=nikomatsakis

PR #661, opened 2019-06-05
Release rayon 1.1.0 / rayon-core 1.5.0

cuviper:


RELEASES.md
```diff
@@ -1,4 +1,48 @@
-# Release rayon 1.0.3
+# Release rayon 1.1.0 / rayon-core 1.5.0 (2019-06-05)
```
cuviper: I'll check this date before we actually merge and publish.

cuviper:
Oh, I almost forgot -- we should pick a logo along with the release!
https://users.rust-lang.org/t/ideas-for-a-rayon-logo/26323

We don't need a long-term/formal commitment here. Let's just get *something* in place for the README and the @rayon-rs org picture. I'd prefer one of the square/roundish choices so it fits as an avatar.

RELEASES.md
nikomatsakis:
```suggestion
# Release rayon 1.1.0 / rayon-core 1.5.0 (2019-06-12)
```

PR #664, opened 2019-06-24
Update crossbeam-deque to 0.7

Eijebong:


cuviper:
I also have this branch ready: https://github.com/rayon-rs/rayon/compare/master...cuviper:crossbeam-deque-0.7

But we don't want to make this change until allowed by our minimum rustc policy in [RFC 3](https://github.com/rayon-rs/rfcs/blob/master/accepted/rfc0003-minimum-rustc.md), for at least one year of Rust support. `crossbeam-deque 0.7` requires [Rust 1.28 (2018-08-02)](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1280-2018-08-02).

Eijebong:
Oh, did that because I couldn't find any open PR with it. I'll close mine then and wait for august.

PR #665, opened 2019-06-25
Add ParallelIterator::copied()

cuviper:


cuviper:
Per usual, we should wait until `std` actually stabilizes `Iterator::copied` (Rust 1.36) before we merge our mimic -- even though this one is trivial and unlikely to change.

cuviper:
It landed in 1.36, so we should be good.

bors r=nikomatsakis

Issue #666, opened 2019-06-28
ThreadPool::install and allocation overhead


ralfbiedert:
## Background

We are using rayon within a library that is called from game engines in the main loop, at about 90 FPS, depending on the device. From the game engine we receive a number of objects which we transform with some math-heavy calculations. 

We have used rayon for some time now, and from a performance / usability perspective it hits a sweep spot where we can simplify our code a lot, and see a good speed up over a single threaded approach.

However, as we are getting into more demanding environments we need to control for allocation, and want to be more-or-less allocation free after initialization. 

## Problem

I started analyzing our allocation pattern and noticed that `ThreadPool::install` does about 2 allocations per invocation, at about 10 - 20 bytes each (after the 1st round):

```
=== Allocation Statistics ===
"()" -- Events: 0, Bytes: 0
"Box::new(1)" -- Events: 1, Bytes: 4
"pool = ThreadPoolBuilder::new().num_threads(4).build()" -- Events: 65, Bytes: 4942
"pool.install(|| {}) -- 1st time" -- Events: 23, Bytes: 5608
"pool.install(|| {}) -- 2nd time" -- Events: 2, Bytes: 24
"pool.install(|| {}) -- next 1000 times" -- Events: 2032, Bytes: 48064
"drop(pool)" -- Events: 0, Bytes: 0
```
([Source](https://github.com/ralfbiedert/rust_issues/tree/rayon_allocation))

Ideally, `ThreadPool::install` should be allocation free eventually, and / or there should be an alternative API that allow for allocation free processing. In the other thread #614 a heapless(?) `StackJob` was mentioned, but I couldn't find anything in the docs.

cuviper:
`StackJob` isn't in the docs because it's an internal detail, but anyway this is already what `install` uses. So we're avoiding direct allocation, but there still may be something in the `SegQueue` which we use to send the job into the thread pool.

cc @stjepang -- is it expected that `SegQueue` would allocate every time? Is there a better alternative for this use case? e.g. Maybe the `crossbeam_deque::Injector` when we update to 0.7...

ralfbiedert:
I just did some digging, this is what I found out so far:

- `ThreadPool::install` invokes `Registry::in_worker`
- There the check `worker_thread.is_null()` constantly fails, so it invokes `self.in_worker_cold(op)`.
- That in turn does a `LockLatch::new()`, which causes the following allocations: 

```
"Mutex::new" -- Events: 1, Bytes: 16
[...]

PR #668, opened 2019-07-02
Reusing LockLatch to reduce allocation 

ralfbiedert:
This addresses the first half of #666.

I kept the `LockLatch` and added a `reset()` method because that had overall fewer lines of code. If you want I can still change it to a separate `ReusableLockLatch` type.

### DIFF
--- a/rayon-core/src/job.rs
+++ b/rayon-core/src/job.rs
@@ -64,7 +64,7 @@ impl JobRef {
 /// executes it need not free any heap data, the cleanup occurs when
 /// the stack frame is later popped.  The function parameter indicates
 /// `true` if the job was stolen -- executed on a different thread.
-pub(super) struct StackJob<'a, L, F, R>
+pub(super) struct StackJob<'a, L: 'a, F, R>
 where
     L: Latch + Sync,
     F: FnOnce(bool) -> R + Send,
--- a/rayon-core/src/job.rs
+++ b/rayon-core/src/job.rs
@@ -64,24 +64,24 @@ impl JobRef {
 /// executes it need not free any heap data, the cleanup occurs when
 /// the stack frame is later popped.  The function parameter indicates
 /// `true` if the job was stolen -- executed on a different thread.
-pub(super) struct StackJob<L, F, R>
+pub(super) struct StackJob<'a, L, F, R>
 where
     L: Latch + Sync,
     F: FnOnce(bool) -> R + Send,
     R: Send,
 {
-    pub(super) latch: L,
+    pub(super) latch: &'a mut L,
     func: UnsafeCell<Option<F>>,
     result: UnsafeCell<JobResult<R>>,
 }
 
-impl<L, F, R> StackJob<L, F, R>
+impl<'a, L, F, R> StackJob<'a, L, F, R>
 where
     L: Latch + Sync,
     F: FnOnce(bool) -> R + Send,
     R: Send,
 {
-    pub(super) fn new(func: F, latch: L) -> StackJob<L, F, R> {
+    pub(super) fn new(func: F, latch: &'a mut L) -> StackJob<'a, L, F, R> {
         StackJob {
             latch,
             func: UnsafeCell::new(Some(func)),
@@ -102,7 +102,7 @@ where
     }
 }
 
-impl<L, F, R> Job for StackJob<L, F, R>
+impl<'a, L, F, R> Job for StackJob<'a, L, F, R>
 where
     L: Latch + Sync,
     F: FnOnce(bool) -> R + Send,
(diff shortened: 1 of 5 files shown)


rayon-core/src/job.rs
```diff
@@ -64,24 +64,24 @@ impl JobRef {
 /// executes it need not free any heap data, the cleanup occurs when
 /// the stack frame is later popped.  The function parameter indicates
 /// `true` if the job was stolen -- executed on a different thread.
-pub(super) struct StackJob<L, F, R>
+pub(super) struct StackJob<'a, L: 'a, F, R>
```
nikomatsakis: instead of adding `'a`, can we implement `Latch` for `&'a mut L` where `L: Latch`?

nikomatsakis: or, probably,, just `&L`

nikomatsakis: (Why are we using `&mut L` in the first place?)

ralfbiedert:
> The bigger question, which I raised on gitter, is whether we should just use the parking_lot crate, which offers locks w/o allocation as I understand it. @cuviper and I were discussing the pros/cons, which (I think) come down to the con of taking another dependency vs the pro of simpler code.

From a user perspective, the number of dependencies is observable. Unless `parking_lot` adds another measurable benefit, it's probably just an implementation detail. 

I can't judge how much adding it would change the existing code. 

Given how widely used `rayon` is, I think the bar for lowering an observable metric without providing another benefit should be high. If it's just about replacing the few lines of this PR its probably not worth it.

Regardless of the implementation used, as a user and contributor I'd prefer having a unit test that verifies this implementation is "according to spec" (i.e., does not allocate), but that might be another PR / issue.

rayon-core/src/job.rs
ralfbiedert: This should be addressed now.

rayon-core/src/join/mod.rs
```diff
@@ -120,13 +120,12 @@ where
...
+        let latch = SpinLatch::new();
+
         // Create virtual wrapper for task b; this all has to be
         // done here so that the stack frame can keep it all live
         // long enough.
-        let job_b = StackJob::new(
-            |migrated| oper_b(FnContext::new(migrated)),
-            SpinLatch::new(),
-        );
+        let job_b = StackJob::new(|migrated| oper_b(FnContext::new(migrated)), &latch);
```
cuviper: Why did you make this one a reference? A `SpinLatch` is just an `AtomicBool`, so it seems all you're doing is adding needless pointer indirection in this `StackJob`.

rayon-core/src/registry.rs
```diff
@@ -611,6 +620,8 @@ pub(super) struct WorkerThread {
 // for a RefCell<T> etc.
 thread_local! {
     static WORKER_THREAD_STATE: Cell<*const WorkerThread> = Cell::new(ptr::null());
+
+    static LOCK_LATCH: RefCell<LockLatch> = RefCell::new(LockLatch::new());
```
cuviper: Why does this need a `RefCell`? From the outside it's only used immutably (no `borrow_mut`), and its inner mutex provides the only mutability we need.

Also, this thread-local doesn't need global scope -- let's move it into `worker_cold` where it's used. It will still be static, shared across multiple calls to that function.

rayon-core/src/registry.rs
```diff
@@ -479,26 +480,34 @@ impl Registry {
...
+
+            // This thread isn't a member of *any* thread pool, so just block.
+            debug_assert!(WorkerThread::current().is_null());
+            let job = StackJob::new(
+                |injected| {
+                    let worker_thread = WorkerThread::current();
+                    assert!(injected && !worker_thread.is_null());
+                    op(&*worker_thread, true)
+                },
+                latch.deref(),
```
cuviper: If you lose the `RefCell`, this will be a `&LockLatch` already.

rayon-core/src/registry.rs
```diff
@@ -479,26 +480,34 @@ impl Registry {
...
+            );
+            self.inject(&[job.as_job_ref()]);
+            job.latch.wait();
+            job.latch.reset(); // Makes sure we can use the same latch again next time.
```
cuviper: Can we make this a unified `wait_and_reset` instead? That way it doesn't have to be re-locked for the reset.

rayon-core/src/registry.rs
```diff
@@ -479,26 +480,34 @@ impl Registry {
             }
         }
     }
-
+    
+
```
cuviper: Please run through `cargo fmt` -- it should clean up odd changes like this.

cuviper:
FWIW, I reverted `in_worker_cross` to a non-reference latch too. The rest looks good, thanks!

bors r+

PR #670, opened 2019-07-02
Removed remaining allocations in ThreadPool::install.

ralfbiedert:
This removes the 2nd allocation source as discussed in #666, feedback welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ target
 *~
 TAGS
 *.bk
+.idea
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -53,6 +53,7 @@ mod private;
 mod job;
 mod join;
 mod latch;
+mod queue;
 mod registry;
 mod scope;
 mod sleep;
--- a/rayon-core/src/queue.rs
+++ b/rayon-core/src/queue.rs
@@ -4,9 +4,9 @@ use crossbeam_queue::{ArrayQueue, PopError, SegQueue};
 ///
 /// This structure is useful if you want a possibly unbounded queue, and are willing to trade
 /// strict ordering for predictable allocations up to a certain point. In particular, if you
-/// load fewer than `cap` elements, this structure will not allocate.
+/// load fewer than `cap` elements, this structure will not allocate during `push()` or `pop()`.
 ///
-/// Also, up to a load of `cap` the structure will behave like a regular, but pre-allocated queue.
+/// Also, up to a load of `cap` the structure will behave like a regular, pre-allocated queue.
 /// Beyond that, the ordering of elements might be random until the load falls below `cap` again.
 pub(super) struct AlmostQueue<T> {
     unbounded: SegQueue<T>,
--- a/rayon-core/src/queue.rs
+++ b/rayon-core/src/queue.rs
@@ -1,45 +1,49 @@
 use crossbeam_queue::{ArrayQueue, PopError, SegQueue};
 
 /// A data exchange structure that tries to be a queue and almost succeeds.
-///
-/// This structure is useful if you want a possibly unbounded queue, and are willing to trade
-/// strict ordering for predictable allocations up to a certain point. In particular, if you
-/// load fewer than `cap` elements, this structure will not allocate during `push()` or `pop()`.
-///
-/// Also, up to a load of `cap` the structure will behave like a regular, pre-allocated queue.
-/// Beyond that, the ordering of elements might be random until the load falls below `cap` again.
-pub(super) struct AlmostQueue<T> {
+pub(crate) struct AlmostQueue<T> {
     unbounded: SegQueue<T>,
     bounded: ArrayQueue<T>,
 }
 
 impl<T> AlmostQueue<T> {
-    pub(super) fn new(cap: usize) -> Self {
+    pub(crate) fn new(cap: usize) -> Self {
         Self {
             unbounded: SegQueue::new(),
             bounded: ArrayQueue::new(cap),
         }
     }
-
-    pub(super) fn push(&self, t: T) {
-        if let Err(x) = self.bounded.push(t) {
-            // At this point we know we failed to insert into the bounded queue. In the meantime
-            // (between the Err and the next line) the bounded one might have become empty again,
-            // but we insert into the unbounded one anyway.
-            self.unbounded.push(x.0)
+    
+    pub(crate) fn push(&self, t: T) {
+        // If unbounded has any elements, it always takes over receiving elements.
+        if !self.unbounded.is_empty() {
+            // Once we entered here, bounded might have become empty again, and a newer element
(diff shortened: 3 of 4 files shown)


cuviper:
You noted https://github.com/rayon-rs/rayon/issues/666#issuecomment-507733350:

> Due to it's simplicity it loses the FIFO property past a certain point (if you insert more than `cap` elements ordering might be mixed up at the tranisition zone if multiple threads `push` / `pop` at the same time).

So if I push enough jobs to overflow the `ArrayQueue` into the `SegQueue`, then as soon as the threadpool does pop, the next pushed job gets to jump the ahead of the `SegQueue`. This race will continue as long as you're still pushing more jobs, so the overflowed jobs might be delayed indefinitely.

> However, if I understand it correctly the`injected_jobs` inside the `Registry` don't really mind if that happens, because a job might get injected "randomly" anyway, e.g., from another task.

We don't need to worry about the order of jobs from separate contexts, but I think we would like to preserve FIFO ordering of jobs from a given thread. [Rayon RFC 1](https://github.com/rayon-rs/rfcs/blob/master/accepted/rfc0001-scope-scheduling.md) went into some of the queue ordering that we try to maintain. There's always going to be some uncertainty with randomized work-stealing within the pool, but I'd like to do our best to be *somewhat* predictable.

Maybe this could be mitigated by only trying to push to the `ArrayQueue` when the `SegQueue` is empty. But at some point, these extra checks will be more trouble than it would have been to just allow the allocations. I'm more inclined to wait and see if they can improve in crossbeam-rs/crossbeam#398.

ralfbiedert:
> So if I push enough jobs to overflow the `ArrayQueue` into the `SegQueue`, then as soon as the threadpool does pop, the next pushed job gets to jump the ahead of the `SegQueue`. This race will continue as long as you're still pushing more jobs, so the overflowed jobs might be delayed indefinitely.


Yes. I thought about that, but considered it a "pathological" case:

If the number of elements is lower than `cap` it doesn't happen and everything is in order.

Once the number of elements passes and stays larger than cap, anything that gets inserted is also in the right order. 

The only unlucky elements are the ones that get inserted into the queue right when it overflows _and_ when another element got interrupted failing the bounded check and moving unbounded. 

That unlucky element takes one for the team and will stay around until the unbounded queue is empty again.


> We don't need to worry about the order of jobs from separate contexts, but I think we would like to preserve FIFO ordering of jobs from a given thread. [Rayon RFC 1](https://github.com/rayon-rs/rfcs/blob/master/accepted/rfc0001-scope-scheduling.md) went into some of the queue ordering that we try to maintain. There's always going to be some uncertainty with randomized work-stealing within the pool, but I'd like to do our best to be _somewhat_ predictable.

Ah, wasn't aware of the RFCs.

> Maybe this could be mitigated by only trying to push to the `ArrayQueue` when the `SegQueue` is empty. But at some point, these extra checks will be more trouble than it would have been to just allow the allocations. I'm more inclined to wait and see if they can improve in [crossbeam-rs/crossbeam#398](https://github.com/crossbeam-rs/crossbeam/issues/398).

I agree that adding more logic is probably not worth it. 

Feel free to close if you think `SegQueue` will fix eventually it. 

However, if it won't fix, at least I would value not having allocations higher than the added "worst case" this introduces. Maybe the queue / scheduler could then be user configured or injected depending on the use case?

ralfbiedert:
It's not a stack, it's not a queue, it's a schque ... In all seriousness my brain went the other direction while I was writing that. Please disregard. Will get back tomorrow.

ralfbiedert:
Let me try this again ... I updated the PR to behave more like a real queue. With respect to a global clock and the time a producer called `push()` it still isn't _perfect_ (although I was wondering in a mpmc scenario what the actual ground truth for ordering is), however, it should behave much more FIFO-ish.

To verify, I also [created a test case](https://gist.github.com/ralfbiedert/e9fef930239ceaa79dc6c3b034afcbf7), where I simulated queue behavior in about 1000 different constellations along these parameters:

```rust
let queue_pre_allocation_size = 2 + rand::random::<usize>() % 30;
let total_threads= 1 + rand::random::<usize>() % 31;
let percent_pushers= 0.2f32 + rand::random::<f32>() / 2f32;
let popper_sleep = 50 + rand::random::<u32>() % 10_000;
let pusher_sleep = 50 + rand::random::<u32>() % 10_000;
let num_push = 1 + rand::random::<usize>() % 7;
let num_pop = 1 + rand::random::<usize>() % 7;
```

I then measured how, according to an atomic number producer, and a shared observation pipeline, how much observed ranks would differ between the queue in this PR, and the `SeqQueue`. Apparently not much, if at all:

Untitled

With that said, having `SegQueue` fixed would be nice. However if that doens't happen I think the above could be a rather simple alternative. My thought was to keep this PR open until `SegQueue` is fixed, there is a better alternative (or until this is merged). I'm also inclined to use this in our code base in the meantime because it seems to work.

However, I'd love to hear if you still think this isn't a good idea or violates any other rayon spec that's not covered by the existing unit tests.

zayenz:
> ```rust
> let queue_pre_allocation_size = 2 + rand::random::<usize>() % 30;
> let total_threads= 1 + rand::random::<usize>() % 31;
> let percent_pushers= 0.2f32 + rand::random::<f32>() / 2f32;
> let popper_sleep = 50 + rand::random::<u32>() % 10_000;
> let pusher_sleep = 50 + rand::random::<u32>() % 10_000;
> let num_push = 1 + rand::random::<usize>() % 7;
> let num_pop = 1 + rand::random::<usize>() % 7;
> ```

As a drive-by comment, consider using `gen_range` instead, since using modulus to get a range of random numbers has several issues (only uses the low-order bits that may have less entropy for some generators, introduces non-uniformity due to full range not being a multiple of the modulus value).

An example (not tested in your particular case), would be to replace with 
```rust
let mut rng = rand::thread_rng(); 
let queue_pre_allocation_size: usize = 2 + rng.gen_range(0, 30);
let total_threads: usize = 1 + rng.gen_range(0, 31);
let percent_pushers: f32 = 0.2 + rng.gen_range(0.0, 0.5);
let popper_sleep: usize = 50 + rng.gen_range(0, 10_000);
let pusher_sleep: usize = 50 + rng.gen_range(0, 10_000);
let num_push: usize = 1 + rng.gen_range(0, 7);
let num_pop: usize = 1 + rng.gen_range(0, 7);
```

The reason is that when running randomized tests, the subtle biases from modding and the potential low-entropy of low-order bits (depending on actual source of randomness) can be a confounding factor that can make reasoning harder. Also, it might not be important at all for your specific case.

ralfbiedert:
Thanks for the comment! 

Yes, I should have used `gen_range()` as good practice. 

I don't think that in this particular case though it will have much impact, as I was doing an A/B test and mostly even used the same numbers for both cases. There might still be a skew in the data, but that would have affected the shape of a pretty random search space to start with.

PR #672, opened 2019-07-05
Update dev/demo dependencies

cuviper:
- rand 0.7
- rand_xorshift 0.2
- glium 0.25

The only code churn is that `Rng::sample_iter` now takes `self` by value
instead of `&mut self`. So in some places we now had an unused mut, and
in others we need an explicit `&mut` to reach `&mut R: Rng` instead.


cuviper:
Well that's annoying -- on Rust 1.26:

```
$ cargo build
    Updating registry `https://github.com/rust-lang/crates.io-index`
error: no matching package named `getrandom_package` found
location searched: registry `https://github.com/rust-lang/crates.io-index`
required by package `rand v0.7.0`
    ... which is depended on by `rayon-core v1.5.0 (file:///home/travis/build/rayon-rs/rayon/rayon-core)`
    ... which is depended on by `rayon v1.1.0 (file:///home/travis/build/rayon-rs/rayon)`
The command "cargo build" exited with 101.
```

Their use of dependency renaming (Rust 1.31) means it won't resolve for older Rust at all, even when we're trying to check a plain build without dev-dependencies.

So, this is blocked on our minimum Rust.

cuviper:
We could still update glium, but TBH that was kind of an afterthought in this PR.

Issue #671, opened 2019-07-05
hitting type_length_limit with relatively few operations


winstonewert:
I'm running into the type_length_limit when using rayon.

As a simplified instance:

```
    Vec::<Result<(),()>>::new()
        .into_par_iter()
        .map(|x| x)
        .map(|x| x)
        .map(|x| x)
        .map(|x| x)
        .map(|x| x)
        .map(|x| x)
        .collect::<Result<(), ()>>();
```

It looks like there is some sort of exponential growth in type names for each additional operation. 

Initially I just increased the type length limit, but this seems to result in very slow compiles.

cuviper:
Have you manually set a lower limit to begin with? The default is 1048576 (= 2<sup>20</sup> = 1M), and your simplified instance is accepted on the playground without trouble:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e5b29846cab15818ee5d8c1e9922927e

I gave a talk a PDXRust recently, and at one point I noted a similar situation of really long symbols. I didn't actually exceed `type_length_limit` with that one either, "only" ~133K, but it probably wouldn't take much to tip it over the edge.
https://gist.github.com/cuviper/99a38550061f1e3513dfe797639ba3ef

winstonewert:
No, I haven't set a lower limit. After looking around a bit, the issue is that I'm using a different version of Rayon than the playground.

Playground uses:

```
rayon = "=1.0.3"
rayon-core = "=1.4.1"

```
I'm on
```
rayon = "1.1.0"
```

winstonewert:
Bisected to this commit:

https://github.com/rayon-rs/rayon/commit/dd7053074ae9b26f4ceebb99ff6ba41f43cfcf40

cuviper:
Oh, I guess playground doesn't update too often. Indeed that fails for me locally with the latest rayon.

```
error: reached the type-length limit while instantiating `<rayon::vec::SliceDrain<std::res...(), std::result::Result<(), !>>>`
     |
     = note: consider adding a `#![type_length_limit="1400241"]` attribute to your crate
```

I'm not certain what affected this, but I suspect it was the map `consume_iter` change in #631, as this would add additional caller and type depth.

winstonewert:
Also, how did you get the full typename in your case? That mgiht be helpful for the purpose of identifying where I'm inadvertently bloating my types.

cuviper:
[...]

PR #673, opened 2019-07-05
Reduce the genericity of many closures

cuviper:
The `take_while` closure only needs to be generic in `T`, along with a
captured `&AtomicBool`. When we write the closure directly, it carries
all the type baggage of `WhileSomeFolder<'f, C>::consumer_iter<I>`,
where the monomorphized type may explode. Instead, we can move this
closure to a standalone function for reduced genericity.

### DIFF
--- a/rayon-core/src/join/mod.rs
+++ b/rayon-core/src/join/mod.rs
@@ -98,7 +98,12 @@ where
     RA: Send,
     RB: Send,
 {
-    join_context(|_| oper_a(), |_| oper_b())
+    #[inline]
+    fn call<R>(f: impl FnOnce() -> R) -> impl FnOnce(FnContext) -> R {
+        move |_| f()
+    }
+
+    join_context(call(oper_a), call(oper_b))
 }
 
 /// Identical to `join`, except that the closures have a parameter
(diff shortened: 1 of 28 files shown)


cuviper:
This fixes the type-length-limit of the specific example in #671, but I'm going to audit for similar cases before we call that fixed. So far, this only takes the max type in that example from ~1.4M characters to ~700K, sneaking under the compiler's default 1M limit, but that's still way too close for comfort.

cuviper:
OK, I think we're good now. I swept through pretty much everything, isolating closures wherever I saw there was a type parameter that could be avoided. This is not meant to be an aversion to closures entirely, just where they are implicitly overly generic.

rayon-core/src/join/mod.rs
```diff
@@ -98,7 +98,12 @@ where
     RA: Send,
     RB: Send,
 {
-    join_context(|_| oper_a(), |_| oper_b())
+    #[inline]
+    fn call<R>(f: impl FnOnce() -> R) -> impl FnOnce(FnContext) -> R {
+        move |_| f()
+    }
+
+    join_context(call(oper_a), call(oper_b))
```
cuviper: To call out one example -- before, both of those closure wrappers depended on all 4 parameters, `<A, B, RA, RB>`, which meant that the `join_context` type also carried that duplication. Now the `call` wrappers only depend on their respective types, `<A, RA>` or `<B, RB>`.

nikomatsakis:
I'd be happy to r+ this, but @cuviper did you want to wait until the corresponding PR on rust-lang landed?

cuviper:
I was ready to wait, but we don't necessarily need to. I was more concerned whether upstream rust thought this approach was reasonable, and since it was approved, that's good enough here.

Looks like I have a conflict to resolve here though -- will fix.

cuviper:
The rust PR is now merged, so the waiting question is now moot. :smile: 

bors r=nikomatsakis

PR #684, opened 2019-08-24
Add env vars for ROUNDS_UNTIL_SLEEPY and ROUNDS_UNTIL_ASLEEP

aloucks:
Running rayon in a rather tight loop can cause very high cpu and power consumption that is not proportional to the amount of work being performed. As documented in the sleep module, this is caused by the work stealing algorithm and workers spinning in a hot loop while looking for work.

This commit adds two environment variables that serve as configuration knobs and allows the user to adjust how aggressive the workers behave.

* `RAYON_ROUNDS_UNTIL_SLEEPY`
* `RAYON_ROUNDS_UNTIL_ASLEEP`

### DIFF
--- a/rayon-core/src/sleep/mod.rs
+++ b/rayon-core/src/sleep/mod.rs
@@ -2,6 +2,7 @@
 //! for an overview.
 
 use log::Event::*;
+use std::env;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Condvar, Mutex};
 use std::thread;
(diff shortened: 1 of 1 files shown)


wagnerf42:
just to be sure I get it. a round (bounded by the 32 and 64 values) is a steal request ?
what is the duration before sleep ? I cannot really see how this could be a problem. you have a sequential loop on really tiny unbalanced parallel computations ?

nikomatsakis:
OK. I don't think we'd want to expose those details, but I do think it might be reasonable to let users supply some kind of configuration to the thread-pool when its created. Probably a more abstract one. (Of course we've long wanted to revisit our "sleep mechanism" in any case.)

nikomatsakis:
The sleeping code has been completely rewritten by now, so I think we can close this PR. @aloucks have you had a chance to try the new scheduler by the way? Hopefully it is doing a better job when it comes to CPU performance.

PR #691, opened 2019-09-06
refactoring: have latches tickle registries directly

nikomatsakis:
As a start to improving our sleep system, this PR does a refactoring to how tickle works. We used to tickle the registry after each job executed: the idea was that executing a job would set some latch, and we needed to wake up any threads that might be blocked, waiting on that latch. However, this was a bit overapproximated: for example, the latch might've been a `LockLatch`, which indicates a thread from outside the pool, in which case there is no need to tickle threads at all. 

In this PR, the latches themselves track the registry that they must tickle when they are set. In the case of a `Countdown` latch, they are given it from the outside; this is because countdown latches are sometimes owned by the registry itself, so it would be impossible for them to have a reference to the registry.

I'd like to work towards a scenario where we know not only the *registry* that must be awoken but the exact helper thread. This could avoid needless wakeups. But this refactoring is as far as I got for now.

### DIFF
--- a/rayon-core/src/join/mod.rs
+++ b/rayon-core/src/join/mod.rs
@@ -138,7 +138,7 @@ where
         // Create virtual wrapper for task b; this all has to be
         // done here so that the stack frame can keep it all live
         // long enough.
-        let job_b = StackJob::new(call_b(oper_b), SpinLatch::new());
+        let job_b = StackJob::new(call_b(oper_b), SpinLatch::new(worker_thread.registry()));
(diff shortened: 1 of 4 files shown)


ktaeleman:
Tried this patch and it does seem to reduce our cpu load immensely. Thanks!

cuviper:
Is this still relevant / worth reviewing on its own, in relation to rayon-rs/rfcs#5?

nikomatsakis:
closing in favor of https://github.com/rayon-rs/rayon/pull/746

PR #692, opened 2019-09-06
Link the iterator traits from top of the docs

### LINKED ISSUE #687: docs: Link to the ParallelIterator from the intro?
When I open the https://docs.rs/rayon/1.2.0/rayon/, there's a lot of links to various parts of the documentation. However, I myself usually want to look at the methods of `ParallelIterator` at https://docs.rs/rayon/1.2.0/rayon/iter/trait.ParallelIterator.html, which is not linked from there.

As the main trait of crate, would it make sense to have it linked directly somewhere at the top, maybe instead of the `iter` submodule (which one is seldom the thing users really need to interact with directly)?

I'm not sending a PR, because I want to ask about the exact location and wording first 😇.

vorner:
As they contain the often looked up methods, have them readily
available.

Closes #687.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,7 +18,7 @@
 //!   typically the most efficient.
 //!   - [Parallel iterators][iter module] make it easy to convert a sequential iterator to
 //!     execute in parallel.
-//!     - The [`ParallelIterator`] trait defines general methods for for all parallel iterators.
+//!     - The [`ParallelIterator`] trait defines general methods for all parallel iterators.
 //!     - The [`IndexedParallelIterator`] trait adds methods for iterators that support random
 //!       access.
 //!   - The [`par_sort`] method sorts `&mut [T]` slices (or vectors) in parallel.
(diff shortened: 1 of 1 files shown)


nikomatsakis:
bors r+

Thanks @vorner!

src/lib.rs
```diff
@@ -18,6 +18,9 @@
 //!   typically the most efficient.
 //!   - [Parallel iterators][iter module] make it easy to convert a sequential iterator to
 //!     execute in parallel.
+//!     - The [`ParallelIterator`] trait defines general methods for for all parallel iterators.
```
cuviper: I'm going to sneak in a quick typo fix:

```suggestion
//!     - The [`ParallelIterator`] trait defines general methods for all parallel iterators.
```

cuviper: Seems your PR is not open to maintainer pushes. I might just push this after bors is done.

cuviper:
bors r=nikomatsakis

src/lib.rs
vorner: The check-box for maintainers to push is checked, though. Nevertheless, thanks, I've accepted the suggestion. But it seems that have canceled bors :-(.

cuviper: Maybe it's just the web UI that doesn't work for that? Anyway, thank for applying, and I've reapproved the PR.

PR #698, opened 2019-09-18
Issue 697

jClaireCodesStuff:
Wrote a test case.  It demonstrates that the master sends `!Send` results, and fails to compile after my changes.  The existing 12 tests pass.  

Removed `unsafe impl` of `Sync` and `Send` for `ScopeFuture`.  Replaced it with:
 - `Send` bounds for the return types of the user's `Future`
 - a newtype wrapping `ScopeHandle` with unsafe `impl Send`

Now the only other type in the crate with `unsafe impl` is `ScopeFutureWrapped`.

I need help transforming the test into one that expects a compiler error.

I did not investigate whether sending `ScopeHandle` is sound.


cuviper:
I do think you're right that this needs `Send`, but we also plan to just deprecate and remove the rayon-futures crate, since it's based on old futures. New integration will probably be added directly to `rayon-core`, something like #679.

This PR constitutes a breaking change, though it's to fix soundness, which we've allowed without a semver bump before. Plus rayon-futures requires the nasty `--cfg=rayon_unstable` anyway. So I'm thinking of pushing your changes to a new branch, along with publishing a new release that pins the compatible rayon-core minor version and marks rayon-futures deprecated. Then I will remove it for the new rayon-core minor version coming to the master branch.

Does this sound OK?

cuviper:
I published rayon-futures 0.1.1 with this change.

Issue #628, opened 2019-01-23
Rayon `ParallelIterator` not object safe 


HarkonenBade:
Is there a particular reason why the Rayon `ParallelIterator` trait requires `Self: Sized`? As I am attempting to box an in progress parallel iteration and this results in the construction of a trait object being impossible.

nikomatsakis:
We decided that it wasn't worth the limitations to make things object safe pretty early on. I can't even remember all the places that we took advantage of that anymore, but I thnk there are many, and it would be a breaking change to change it now. So I'm going to close the issue as "not going to fix". =)

However, if you *do* want a trait object, you may be able to create your own trait that offers just the methods you want, and implement it for your own newtype:

```rust
trait MyTrait { ... }

struct SomeParallelIterator<T: ParallelIterator> { t: T }

impl<T> MyTrait for SomeParallelIterator<T> { ... }
```

then you can have `Box<dyn MyTrait>`, presumably.

HarkonenBade:
hmm, not sure if that will work for my scenario, given I have a trait that applies something to an iterator as a transform and I wanted to make the equivalent for ParallelIterator:

```rust
trait IntoMap<S, X> {
  fn into_map<I: IntoIterator<Item = S>>(self, i: I) -> Box<dyn Iterator<Item=X>>; 
}
```

I'm forced to use Box for the output as you cannot use impl Trait in trait methods.

Qqwy:
I am encountering a similar problem. I am trying to parallelise the following code:

```rust
fn fancy<Input, Output, Item>(input: Input, depth: usize) -> Output
where
    Input : IntoIterator<Item = Item>,
    Output : std::iter::FromIterator<Item>,
    Item : Copy + std::ops::Add<Output = Item>,
{
    let mut iter: Box<dyn Iterator<Item = Item>> = Box::new(input.into_iter());
    for _ in 0..depth {
        iter = Box::new(iter.flat_map(|x| vec![x, x + x].into_iter()));
    }
    iter.collect()
}
```

Which would become something like:

```rust
use rayon::prelude::*;
fn fancy2<Input, Output, Item>(input: Input, depth: usize) -> Output
where
    Input : IntoParallelIterator<Item = Item>,
    Output : FromParallelIterator<Item>,
    Item : Copy + std::ops::Add<Output = Item> + Send,
{
    let mut iter: Box<dyn rayon::iter::ParallelIterator<Item = Item>> = Box::new(input.into_par_iter());
[...]

PR #702, opened 2019-10-22
Adds info about (lack of) object safety to `iter` module documentation.

Qqwy:
Because I spent quite a while yesterday and today tracking this problem down, I thought it might be an improvement to the documentation to mention that Rayon's ParallelIterator and friends are not object-safe (and why).

(c.f. #628)

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -74,7 +74,7 @@
 //! `None`/`Err` values will exit early.
 //!
 //! A note about object safety: It is currently _not_ possible to wrap
-//! a ParallelIterator (or any trait that depends on it) using a
+//! a `ParallelIterator` (or any trait that depends on it) using a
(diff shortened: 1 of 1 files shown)


src/iter/mod.rs
```diff
@@ -72,6 +72,12 @@
 //! to mirror the unstable `std::ops::Try` with implementations for `Option` and
 //! `Result`, where `Some`/`Ok` values will let those iterators continue, but
 //! `None`/`Err` values will exit early.
+//!
+//! A note about object safety: It is currently _not_ possible to wrap
+//! a ParallelIterator (or any trait that depends on it) using a
```
```suggestion
//! a `ParallelIterator` (or any trait that depends on it) using a
```

Qqwy:
@cuviper Thank you for being so precise! :slightly_smiling_face: 

The formatting change has been applied.

cuviper:
Thanks!

bors r+

PR #703, opened 2019-10-30
increase version of num_cpus to support RustHermit

stlankes:
We are developing the unikernel RustyHermit ([https://github.com/hermitcore/libhermit-rs](https://github.com/hermitcore/libhermit-rs)), where the kernel is written in Rust and is already part of the Rust Standard Library. With this pull request, we want to increate the version number of num_cpus, which already supports RustyHermit.

### DIFF
--- a/rayon-core/Cargo.toml
+++ b/rayon-core/Cargo.toml
@@ -15,7 +15,7 @@ categories = ["concurrency"]
 
 # Some dependencies may not be their latest version, in order to support older rustc.
 [dependencies]
-num_cpus = "1.2"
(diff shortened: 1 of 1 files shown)


cuviper:
Versions 1.2 and 1.11 are semver compatible, so folks can already `cargo update` to use the newer dependency. Cargo will also use the newest compatible version for new projects. So is there much reason to increase our *minimum* version?

(I don't really have a reason against either, except for avoiding needless churn.)

stlankes:
Ah, yes, you are right. I close the pull request.

PR #704, opened 2019-11-07
Update crossbeam to avoid multi version dependency

dnaka91:
Just did a `cargo update` on my project today morning and noticed that with the release of **crossbeam-epoch 0.8.0**, rayon starts to depend on 2 versions of **crossbeam-utils**.

Before:
```
 rayon-core v1.6.0
 ├── crossbeam-deque v0.7.2
 │   ├── crossbeam-epoch v0.8.0
 │   │   └── crossbeam-utils v0.7.0 <-- latest
 │   └── crossbeam-utils v0.7.0     <-- latest
 ├── crossbeam-queue v0.1.2
 │   └── crossbeam-utils v0.6.6     <-- outdated
 └── crossbeam-utils v0.6.6         <-- outdated
```

Here Rust will compile 2 versions of the same dependency.

After:
```
rayon-core v1.6.0
├── crossbeam-deque v0.7.2
│   ├── crossbeam-epoch v0.8.0
│   │   └── crossbeam-utils v0.7.0 <-- latest
│   └── crossbeam-utils v0.7.0     <-- latest
├── crossbeam-queue v0.2.0
│   └── crossbeam-utils v0.7.0     <-- latest
└── crossbeam-utils v0.7.0         <-- latest
```

Here all references are up to date and Rust will only compile one version of `crossbeam-utils`.

PR #705, opened 2019-11-07
Update compat-Cargo.lock

matklad:
I've noticed that https://github.com/rayon-rs/rayon/pull/704 uses old compat lockfile, which is strange, given that it changes Cargo.lock.

What happens here is that cargo writes new compat lockfile during CI.

That means that CI might break randomly, if rayon's dependency is published with the same major version, but different MSRV

With this PR, CI will break deterministically if you change Cargo.toml and don't update compat-Cargo.lock.

Not sure if this actually makes sense :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
       os: linux
       #if: everything!
       before_script: cp ci/compat-Cargo.lock ./Cargo.lock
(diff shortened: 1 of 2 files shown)


cuviper:
Good catch, thank you!

bors r+

PR #711, opened 2019-12-05
impl IntoParallelIterator for tuples => MultiZip

### LINKED ISSUE #567: Support for Zipping multiple array
I am currently using `rayon` iterator for zipping two arrays. I need to zip multiple arrays and a combination of rayon and `itertools` will definitely help.  for now, I had to switch to `itertools` which is sequential.

See the similar request on [itertools](https://github.com/bluss/rust-itertools/issues/271)

cuviper:
This is implemented for tuples up to arity 12, much like the standard
library's trait implementations.

- For `(a, b, ...)`, it calls `into_par_iter()` on each member.
- For `&(a, b, ...)`, it calls `par_iter()` on each member.
- For `&mut (a, b, ...)`, it calls `par_iter_mut()` on each member.

The resulting `MultiZip` iterator returns a tuple of the zipped items
from each input iterator. Internally, it is implemented with macros that
forward to a series of regular `zip`s, mapping to a flattened tuple.

Closes #567.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -141,6 +141,8 @@ mod zip;
 pub use self::zip::Zip;
 mod zip_eq;
 pub use self::zip_eq::ZipEq;
+mod multizip;
+pub use self::multizip::MultiZip;
 mod interleave;
 pub use self::interleave::Interleave;
 mod interleave_shortest;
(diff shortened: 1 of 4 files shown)


nikomatsakis:
My obvious hesitation is that this goes beyond the surface area of libstd. But I think it makes a lot of sense to have `IntoParallelIterator` on tuples and zipping is the obvious meaning.

cuviper:
It seems windows-gnu is broken: https://github.com/rust-lang/rust/issues/67408

cuviper:
bors retry

PR #712, opened 2019-12-10
Rust 2018

c410-f3r:
Converts code-base to Rust 2018 and bumps minimum version to 1.31.

Brought to you by `rustfix`


c410-f3r:
Not sure about `compat-Cargo.lock`

cuviper:
```
error[E0658]: imports can only refer to extern crate names passed with `--extern` on stable channel (see issue #53130)
  --> rayon-core/src/lib.rs:60:9
   |
45 | mod join;
   | --------- not an extern crate passed with `--extern`
...
60 | pub use join::{join, join_context};
   |         ^^^^
   |
note: this import refers to the module defined here
  --> rayon-core/src/lib.rs:45:1
   |
45 | mod join;
   | ^^^^^^^^^
```

Looks like [1.32 uniform paths](https://blog.rust-lang.org/2019/01/17/Rust-1.32.0.html#final-module-improvements) -- you should probably use `rustfix`/`cargo fix` from 1.31.

We'll need to increase the minor versions of the crates for our [minimum rust policy](https://github.com/rayon-rs/rfcs/blob/master/accepted/rfc0003-minimum-rustc.md).
This will also unblock #707.

c410-f3r:
@cuviper Paths fixed

It is not possible to `cargo test` or `cargo build -all` with 1.31 because `docopt` requires 1.32.

examples/cpu_monitor.rs
```diff
@@ -1,10 +1,8 @@
-extern crate docopt;
-extern crate rayon;
 #[macro_use]
 extern crate serde_derive;
-extern crate serde;
 
 use docopt::Docopt;
+use rayon;
```
cuviper: I know rustfix does this, but I think this `use` is not actually needed, as `rayon` should be in the crate prelude. There are a few cases of this, but I won't comment them all. Not a _big_ deal, just something that could be cleaned up.

cuviper:
I cut back Windows CI with #714, so let's see how this goes...

bors r+

Tom1380:
Any plans to update to Rust 2021?

cuviper:
@Tom1380 no specific plans, but we try to support versions of Rust for at least a year:
https://github.com/rayon-rs/rfcs/blob/master/accepted/rfc0003-minimum-rustc.md

I'm not aware of any motivating factor to press for that update. Crates are interoperable across editions.

PR #717, opened 2019-12-22
Change wording, less "High level for low level"

matthiasbeyer:
This patch changes the wording of the documentation, because right now it is "There's a highlevel overview over the low level stuff", which is ridiculous.

Rather use "general overview" for less funny wording.


cuviper:
I don't think it's so ridiculous, but your change is fine too, so I'll take it. Thanks!

bors r+

Issue #494, opened 2017-12-17
Implement step_by


hherman1:
I believe this should be parallelizable? The iterator is currently in nightly rust:

https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.step_by

cuviper:
It should be possible, yes.  I think I'd prefer to wait for `Iterator::step_by` to stabilize though, so we can make sure to match its signature.

lu-zero:
The iterator is now in stable since a while :)

gbip:
This would be a very useful feature :smile:

emmanuelantony2000:
Can i try to implement it?

cuviper:
Go for it!

FlyingCanoe:
I have created a pull request adding this future, but It depend on std StepBy being a double ended iterator. This trait as bean stabilise in rustc 1.38. Does bumping the minimal rustc version is acceptable for this future?

cuviper:
I don't think we should bump the MSRV for this feature. We can either implement our own version of the `StepBy` iterator, or add a build script which detects support to set a `cfg` (so it wouldn't be available at all earlier). Use [autocfg](https://crates.io/crates/autocfg) if you go for a build script, where you can probe an expression like `(0..10).step_by(2).rev()`.

Coder-256:
I think that this #726 was supposed to close this. :)

cuviper:
Yes, thanks!

PR #726, opened 2020-02-19
add step by parallel iterator.

FlyingCanoe:
this is a pull request for the #494. It add a step_by iterator for `IndexedParallelIterator` producing a iterator that is also indexed parallel iterator. It is Implemented in pull mode.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -151,6 +151,9 @@ mod intersperse;
 pub use self::intersperse::Intersperse;
 mod update;
 pub use self::update::Update;
+mod step_by;
+#[cfg(step_by)]
+pub use self::step_by::StepBy;
 
 mod noop;
 mod rev;
@@ -2424,6 +2427,30 @@ pub trait IndexedParallelIterator: ParallelIterator {
     /// ```
     fn enumerate(self) -> Enumerate<Self> {
         Enumerate::new(self)
+    }
+
+    /// Creates an iterator that steps by the given amount
+    ///
+    /// # Examples
+    ///
+    /// ```
+    ///use rayon::prelude::*;
+    ///
+    /// let range = (3..10);
+    /// let result: Vec<i32> = range
+    ///    .into_par_iter()
+    ///    .step_by(3)
+    ///    .collect();
+    ///
+    /// assert_eq!(result, [3, 6, 9])
+    /// ```
+    ///
+    /// # Compatibility
+    ///
+    /// This method is only available on Rust 1.38 or greater.
+    #[cfg(step_by)]
+    fn step_by(self, step: usize) -> StepBy<Self> {
+        StepBy::new(self, step)
     }
 
     /// Creates an iterator that skips the first `n` elements.
--- a/src/iter/step_by.rs
+++ b/src/iter/step_by.rs
@@ -1,0 +1,144 @@
+#![cfg(step_by)]
+use std::cmp::min;
+
+use super::plumbing::*;
+use super::*;
+use crate::math::div_round_up;
+use std::iter;
+use std::usize;
+
+/// `StepBy` is an iterator that skips `n` elements between each yield, where `n` is the given step.
+/// This struct is created by the [`step_by()`] method on [`IndexedParallelIterator`]
+///
+/// [`step_by()`]: trait.IndexedParallelIterator.html#method.step_by
+/// [`IndexedParallelIterator`]: trait.IndexedParallelIterator.html
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+#[derive(Debug, Clone)]
+pub struct StepBy<I: IndexedParallelIterator> {
+    base: I,
+    step: usize,
+}
+
+impl<I> StepBy<I>
+where
+    I: IndexedParallelIterator,
+{
+    /// Create a new `StepBy` iterator.
+    pub(super) fn new(base: I, step: usize) -> Self {
+        StepBy { base, step }
+    }
+}
+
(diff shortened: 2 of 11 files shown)


FlyingCanoe:
forget to run cargo test

FlyingCanoe:
It should be ready by now.

src/iter/step_by.rs
```diff
@@ -0,0 +1,118 @@
...
+                base: left,
+                step: self.step,
+            },
+            StepByProducer {
+                base: right,
+                step: self.step,
+            },
+        )
+    }
+}
```
cuviper: This should also define `min_len` and `max_len` -- again the same math as `src/iter/chunks.rs`.

FlyingCanoe:
I have tow question

- Do there are requirement for a minimal rustc version in which test can build. It seam that some of the dev dependency do have a MSRV superior to 1.31.0.

- Can I safely squash some of my commit together, the GitHub doc about pull request say not to force push as it can corrupt pr. while other resource say that as long as your are the only one working in that branch it is safe.

thank in advance.

cuviper:
> Do there are requirement for a minimal rustc version in which test can build. It seam that some of the dev dependency do have a MSRV superior to 1.31.0.

We run tests on the current stable, beta, and nightly.

For MSRV we only check the plain build. If you look at `.travis.yml` you can see that this uses a locked set of dependencies too, per `ci/compat-Cargo.lock`. You'll need to update that since you're adding `autocfg`. Copy that to `Cargo.lock`, run `cargo build` to let cargo update it, then copy it back and check the difference.

> Can I safely squash some of my commit together,

Sure -- I would generally prefer to merge a "clean" history, though I don't always insist on that. The ideal would be a series of commits that's functional at each step. For a feature like this, that may just be a single commit.

FlyingCanoe:
It should be ready now.

because the cargo.lock stuff is black magic to me and that it may require further explanation, I decide to make it a separate commit.

It did consider to make the build script it own commit, but it would make the not fully functional (kind of), as oppose to the cargo.lock update where it just ci complaining even if the lib is functional.

thank you for your time

src/iter/mod.rs
```diff
@@ -2426,6 +2429,29 @@ pub trait IndexedParallelIterator: ParallelIterator {
         Enumerate::new(self)
     }
 
+
+    /// Creates a iterator that step by the given amount
```
```suggestion
    /// Creates an iterator that steps by the given amount
```

src/iter/mod.rs
```diff
@@ -2426,6 +2429,29 @@ pub trait IndexedParallelIterator: ParallelIterator {
...
+    /// let range = (3..10);
+    /// let result: Vec<i32> = range
+    ///    .into_par_iter()
+    ///    .step_by(3)
+    ///    .collect();
+    ///
+    /// assert_eq!(result, [3, 6, 9])
+    /// ```
+    /// # Compatibility
+    /// this only work on rust 1.38 or superior
```
```suggestion
    ///
    /// # Compatibility
    ///
    /// This method is only available on Rust 1.38 or greater.
```

src/iter/step_by.rs
```diff
@@ -0,0 +1,131 @@
+use super::plumbing::*;
+use super::*;
+use crate::math::div_round_up;
+use std::iter;
+use std::usize;
+
+/// `StepBy` is an iterator that skip `n` element between each yield. where `n` is the given step
```
cuviper:
```suggestion
/// `StepBy` is an iterator that skips `n` elements between each yield, where `n` is the given step.
```

src/iter/step_by.rs
```diff
@@ -0,0 +1,131 @@
+use super::plumbing::*;
```
cuviper: Rather than repeating `#[cfg(step_by)]`, you can add a single `#!` attribute here to cover the whole file:

```suggestion
#![cfg(step_by)]

use super::plumbing::*;
```

src/iter/step_by.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Debug, Clone)]
+pub struct StepBy<I: IndexedParallelIterator> {
+    base: I,
+    step: usize,
+}
+
+#[cfg(step_by)]
+impl<I> StepBy<I>
+    where
+        I: IndexedParallelIterator,
```
cuviper: The indentation is odd here and elsewhere -- please run `cargo fmt` to clean it up.

src/iter/step_by.rs
```diff
@@ -0,0 +1,131 @@
...
+{
+    type Item = P::Item;
+    type IntoIter = iter::StepBy<P::IntoIter>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.base.into_iter().step_by(self.step)
+    }
+
+    fn split_at(self, index: usize) -> (Self, Self) {
+        let (left, right) = self.base.split_at(index * self.step);
```
cuviper: I think this will fail if the step does not evenly divide the length, because `index * self.step` will be greater than the total length. You can reference `ChunksProducer::split_at` to see how to limit this.

tests/producer_split_at.rs
```diff
@@ -224,6 +224,12 @@ fn enumerate() {
     check(&v, || (0..10).into_par_iter().enumerate());
 }
 
+#[test]
+fn step_by() {
+    let v: Vec<_> = (0..10).step_by(2).collect();
```
cuviper: Can you also add a case where the step does not evenly divide the total length? I think this will expose the issue I noted in `split_at`.

FlyingCanoe:
I think it should be ready now.

PR #728, opened 2020-03-06
Use GitHub workflows for CI

cuviper:

PR #731, opened 2020-03-11
Use GitHub workflows for CI

cuviper:


cuviper:
Same as #728, but trying a local branch to see if bors likes that better than a fork.

bors try

cuviper:
bors try

PR #732, opened 2020-03-11
Modernize Error for ThreadPoolBuildError

cuviper:
- Add `source` to return the inner `io::Error`
- Remove `description`, soft-deprecated since Rust 1.27.


cuviper:
> * Remove `description`, soft-deprecated since Rust 1.27.

@nikomatsakis the only real review question in the PR is whether you're okay with removing this -- callers will get the trait-default message, `"description() is deprecated; use Display"`. We could instead keep our message and `#[allow(deprecated)]` to avoid the warning in Rust 1.42+.

cuviper:
I softened this to keep our `Error::description` around, just masking the deprecation.

bors r+

PR #737, opened 2020-03-16
Use structopt instead of docopt

lu-zero:
It might or might not be nicer to use. (this got prompt by docopt somehow crashing rustc nightly on power9)


cuviper:
I think if we change this, it should be done for `rayon-demo` too for consistency. Direct `clap` is another option, since `structopt` is just a fancy wrapper on that.

Or since this is just a tiny example program, I think it would be fine with manual arg parsing, and get rid of the dev-dependency in rayon altogether.

> (this got prompt by docopt somehow crashing rustc nightly on power9)

ref: https://github.com/rust-lang/rust/issues/70041

lu-zero:
I prefer using structopt (or gumdrop if we want to keep the build times smaller) since it makes the code more compact, not distracting from the important parts and produces informative `--help` that do not go out of sync easily.

I can convert rayon demo today if there is consensus for it.

cuviper:
> I can convert rayon demo today if there is consensus for it.

I don't have a strong opinion on _what_ we use, especially since this is only for examples and demos. The fact that `docopt` declares itself unmaintained is reason enough to find something else. If you're willing to do the rest of the conversion, `structopt` should be fine.

nikomatsakis:
For the record I'm fine with structopt :)

PR #749, opened 2020-04-13
Fix link in doc comment

adrian5:


cuviper:
Thanks, good catch!

I don't know what's going on with CI -- all three checks failed entirely at first, no output. I reran them, and then two passed fine, but the formatting check again failed with no output. I'll let this sit for a bit in case they're having infrastructure issues, then I'll try again.

PR #750, opened 2020-04-16
Align doc comments to use indicative mood for functions

adrian5:
I'll admit it feels a bit pedantic.

Looking at some of the biggies from `std::`, the indicative mood appears to dominate, even for simple `set`ters, `get`ters and `new`s. It also finds ample use within rayon, so this PR is not a total inversion of the present state, as glancing at the diff may suggest!

I've left a few functions such as `registry::wait_until` and `::steal` in imperative mood as I felt uncertain about changing them.


cuviper:
Thanks!

bors r+

PR #759, opened 2020-04-29
In collect to vector, drop any partial result on unwinding

bluss:
When collecting into a vector, track initialized values for correct drop on unwind

The collect folder produces a CollectResult<'c, T>,
This is a "proxy owner" of the slice of elements that the folder has
written to, and if unwinding happens under way, the elements will
be dropped.

Rewrite the Collect methods so that it is used like this:
```rust
Collect::new(&mut vector, len).with_consumer(|consumer| ... );
```
Where the closure must run the collect consumer and return a
CollectResult of the promised length, to prove that the whole vector
has been written to.

The Collect can assert that the final CollectResult is covering the
target slice. The CollectResult borrows from the target slice, and the
lifetime parameter ensures that the collect caller can't subsitute or
fake the collect result.

The previous atomic used for counting writes is now removed, since
we trust the collected result instead. The producers under way
could misbehave, but in that case the CollectResult will not
end up correct.

## Motivation

Even though it is accepted that drops may not run in corner cases, it should
not be routine that we leak on unwind. A robust system should be recoverable
to the best of its ability. The consumer result model maps well to tracking
initialized parts of the target, so the complexity cost is rather low for this change.

## Not addressed in this PR

Current best practices for uninitialized values want us to use raw pointers,
raw slices or `MaybeUninit<T>`; existing parts of the code have not been
updated for this; MaybeUninit (Rust 1.36) can't be used at this point
and raw slices can't be created directly from raw pointers (Rust 1.42).


src/iter/collect/consumer.rs
```diff
@@ -70,44 +112,74 @@ impl<'c, T: Send + 'c> Consumer<T> for CollectConsumer<'c, T> {
...
+        if self.ptr == self.end {
+            panic!("too many values pushed to consumer");
+        }
         unsafe {
+            let head = self.ptr;
+            self.ptr = self.ptr.add(1);
             ptr::write(head, item);
         }
 
         self.local_writes += 1;
```
bluss, line 130: `ptr` and `local_writes` are counting the same sequence in parallel, so that can be updated

src/iter/collect/consumer.rs
```diff
@@ -70,44 +112,74 @@ impl<'c, T: Send + 'c> Consumer<T> for CollectConsumer<'c, T> {
...
         // but `Collect::complete()` will assert the global write count.
 
         // track total values written
         self.global_writes
             .fetch_add(self.local_writes, Ordering::Relaxed);
+        self.complete = true;
+        PartialCollect {
+            start: self.start,
+            len: self.local_writes,
+        }
```
bluss, line 145: There's a problem here if `local_writes` is indeed not covering the whole chunk that the folder was given, then we can't merge this PartialCollect with its right-hand side neighbord.

src/iter/collect/consumer.rs
```diff
@@ -35,12 +41,41 @@ impl<'c, T: Send + 'c> CollectConsumer<'c, T> {
     }
 }
 
+/// Partial collection result
+///
+/// This is a proxy owner of the elements in a (raw) slice; when PartialCollect drops, the elements
+/// will be dropped, unless it is defused using `defuse_complete`.
+#[must_use]
+pub(crate) struct PartialCollect<T> {
```
bluss, line 49: should be `pub(super)`

bluss:
Second draft posted. With more lifetime parameters so that the borrowing is correct. After working with the tests, I think the types guide usage of the collect consumer well, and it's impossible to fool it into creating a non-empty Vec without the whole area having been written to.

There are some inelegant parts, around the verify* function.

src/iter/collect/consumer.rs
```diff
@@ -13,54 +12,94 @@ pub(super) struct CollectConsumer<'c, T: Send> {
...
 }
 
 pub(super) struct CollectFolder<'c, T: Send> {
-    global_writes: &'c AtomicUsize,
     local_writes: usize,
+    complete: bool,
 
     /// An iterator over the *uninitialized* target memory.
-    target: slice::IterMut<'c, T>,
+    // Original start of iteration
```
bluss, line 25:
```suggestion
    // start of slice
```

src/iter/collect/consumer.rs
```diff
@@ -13,54 +12,94 @@ pub(super) struct CollectConsumer<'c, T: Send> {
...
+    start: *mut T,
+    // total length of iteration
```
bluss, line 27:
```suggestion
    // length of slice
```

src/iter/collect/mod.rs
```diff
@@ -91,32 +92,44 @@ impl<'c, T: Send + 'c> Collect<'c, T> {
         let start = self.vec.len();
         let mut slice = &mut self.vec[start..];
         slice = unsafe { slice::from_raw_parts_mut(slice.as_mut_ptr(), self.len) };
-        CollectConsumer::new(&self.writes, slice)
+        CollectConsumer::new(slice)
     }
 
     /// Updates the final vector length.
-    fn complete(self) {
+    fn complete(self, _result: VerifiedCollectResult) {
```
bluss, line 99: Yes, the VerifiedCollectResult is not "tagged" so if you have two Collect, you can mix them up.

src/iter/collect/consumer.rs
```diff
@@ -13,54 +12,94 @@ pub(super) struct CollectConsumer<'c, T: Send> {
     /// certain amount of trust that the producer driving this will behave
     /// itself. Since this count is important to the safety of marking the
     /// memory initialized (`Vec::set_len`), we choose to keep it internal.
```
bluss, line 14: This comment explains the previous qualms with this solution, yet I can't see anymore that it would be possible to "fake" the reduced CollectResult value.

bluss:
(Sorry for the noise). Third draft posted. Using a scoping function on struct Collect solves the previously not so pretty ends. Now, filling up the collect consumer and checking the result are integrated steps, so that it's much easier to use the API correctly.

src/iter/collect/mod.rs
```diff
@@ -57,60 +54,70 @@ where
...
-    right.complete();
+    Collect::new(left, len).with_consumer(|left_consumer| {
+        let mut left_result = None;
+        Collect::new(right, len).with_consumer(|right_consumer| {
+            let (left_r, right_r) = unzip_indexed(pi, left_consumer, right_consumer);
+            left_result = Some(left_r);
+            right_r
+        });
+        left_result.unwrap()
+    });
```
bluss, line 65: The left and right parts cannot be mixed up. First, because they correspond to type parameters `A` and `B`, but even with that difference removed, the borrow checker will stop the user from mixing up left and right results.

cuviper:
I feel like if I just sit back, you're going to converge on an ideal implementation all by yourself!

src/iter/collect/consumer.rs
```diff
@@ -1,66 +1,97 @@
...
+/// CollectResult represents an initialized part of the target slice.
+///
+/// This is a proxy owner of the elements in the slice; when it drops,
+/// the elements will be dropped, unless its ownership is released before then.
+#[must_use]
+pub(super) struct CollectResult<'c, T> {
+    start: *mut T,
+    len: usize,
+    invariant_lifetime: PhantomData<&'c mut &'c mut [T]>,
+}
```
bluss, line 39: One could wonder, since this is the initialized part, if we can't use `&mut [T]` here. But strictly following unsafe coding guidelines, we'd need a separate raw pointer to write through from the folder, since we can't use the slice's pointer to write to the element one past the end. Then we'd have to juggle two pointers, one to use for writing in the folder and one to use for representing the proxy ownership. This could maybe be ok, if the scopes of use are nesting correctly, but harder to reason about?

bluss:
I want to stop and not make further changes (unless we decide something in review). Remake of the uninit memory handling would be interesting, but we can do the best job if we have access to MaybeUninit, I think.

cuviper:
Interesting -- I was worried what the more-involved folding would do to performance, but at least on my Ryzen 3800X, it's faster! I guess removing that shared atomic helps, but I'm surprised it's so much.

```
$ cargo benchcmp bench.master bench.pr759
 name                                              bench.master ns/iter  bench.pr759 ns/iter  diff ns/iter  diff %  speedup
 vec_collect::vec_i::with_collect                  771,504               726,315                   -45,189  -5.86%   x 1.06
 vec_collect::vec_i::with_collect_into_vec         780,705               720,810                   -59,895  -7.67%   x 1.08
 vec_collect::vec_i::with_collect_into_vec_reused  448,375               407,077                   -41,298  -9.21%   x 1.10
```

bluss:
This PR was my training grounds for doing the same for ndarray.

The n-dim case is pretty similar, because we arrange for the output array to be split only in such a way that each chunk is contiguous like a slice.

Instead of matching producer to consumer, both the input arrays and the output is handled in the producer.

We make an internal parallel iterator that produces the split parts as the iterator item. That means we can use standard rayon `.reduce()` to do the equivalent job of the collect reducer here, to combine the partial results. But it doesn't have the rigorous lifetime parameter use as this PR has.

The code is WIP and a current link is: https://github.com/rust-ndarray/ndarray/blob/24f0b20c11f36e7e56a51399e73c77d879710107/src/parallel/impl_par_methods.rs#L85-L130

bluss:
Both the atomic and the folds would have a larger impact in the scenarios when we have a lot of rayon threads, and split the work much more than needed. Maybe the impact is much less noticeable with less threads?

src/iter/collect/consumer.rs
```diff
@@ -108,6 +136,26 @@ impl<'c, T: Send + 'c> UnindexedConsumer<T> for CollectConsumer<'c, T> {
...
+impl<'c, T> Reducer<CollectResult<'c, T>> for CollectReducer {
+    fn reduce(
+        self,
+        mut left: CollectResult<'c, T>,
+        right: CollectResult<'c, T>,
+    ) -> CollectResult<'c, T> {
+        // Merge if the CollectResults are adjacent and in left to right order
+        // else: drop the right piece now and total length will end up short in the end,
+        // when the correctness of the collected result is asserted.
+        if left.start.wrapping_add(left.len) == right.start {
```
cuviper, line 156: Why did you choose `wrapping_add` here?

Also, I worry a little that this can't track ZSTs properly, but I guess the aggregate result will still drop items if it comes up short of the final length, right?

tests/producer_split_at.rs
```diff
@@ -342,3 +342,58 @@ fn zip() {
     check(&v[..5], || (0..5).into_par_iter().zip(10..20));
     check(&v[..5], || (0..10).into_par_iter().zip(10..15));
 }
+
+#[test]
+fn collect_drop_on_unwind() {
```
cuviper, line 347: This `producer_split_at.rs` has a collection of similar structured tests of producers, which doesn't seem to fit your test. Can you move it to its own file?

cuviper:
> (In my tests I was surprised to find that a > million element array was partly split down into single element parts; those weren't the most important chunks being processed, but still.)

Yeah, the adaptive splitting can get really aggressive in the tail end of computation. I usually suggest folks just add `with_min_len`, but maybe we should figure out an automatic limit, because this often surprises people.

src/iter/collect/consumer.rs
bluss: `wrapping_add` was chosen just since it's a safe method, and it seems to do the job, it's more or less an arithmetic check. It's not used for wrapping, but because the requirements compared with `add` are somewhat relaxed. 

Right, we don't track ZST the same way.. we can't see if they are adjacent or not, and just add up the lengths; so we could totally be fooled into merging "non-adjacent" chunks. We add up the length and will drop the right number of them in the end.

src/iter/collect/consumer.rs
```diff
@@ -1,66 +1,97 @@
...
+    }
+}
+
+/// CollectResult represents an initialized part of the target slice.
+///
+/// This is a proxy owner of the elements in the slice; when it drops,
+/// the elements will be dropped, unless its ownership is released before then.
+#[must_use]
+pub(super) struct CollectResult<'c, T> {
+    start: *mut T,
```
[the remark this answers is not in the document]
bluss: I didn't. NonNull doesn't have any useful effect for us here, I think, so I'd skip it

bluss:
I've moved the test. I'm unsure of the benefit from `NonNull` in this case, is it wanted for code "style" reasons?

bluss:
@cuviper Oh on adaptive splitting, I have only been surprised, not have it bother me in performance yet (but it could be a reason we have to prefer a lower thread count for rayon, actually). `with_min_len` does not apply to ndarray's unordered producers, unfortunately. (For some producers, it makes sense to configure them in terms of maximal number of splits instead.)

cuviper:
> I'm unsure of the benefit from `NonNull` in this case, is it wanted for code "style" reasons?

I like the self-documentation, and the chance for niche layout, but off-hand I don't know where it would matter in this case. Not a big deal.

Looks good, thanks!

bors r+

PR #761, opened 2020-05-07
Replace "cfg(test)" with "cfg(doctest)" for readme testing

GuillaumeGomez:
Rustdoc now passes "doctest" when running in test mode.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,7 +82,7 @@
 //!
 //! [faq]: https://github.com/rayon-rs/rayon/blob/master/FAQ.md
 
-#[cfg(test)]
+#[cfg(doctest)]
 doc_comment::doctest!("../README.md");
 
 #[macro_use]


cuviper:
Thanks!

bors r+

src/lib.rs
```diff
@@ -82,7 +82,7 @@
 //!
 //! [faq]: https://github.com/rayon-rs/rayon/blob/master/FAQ.md
 
-#[cfg(test)]
+#[cfg(doctest)]
```
taiki-e, line 85: This seems to cause compile to fail on Rust 1.38 and Rust 1.39:

```console
$ cargo +1.39.0 check
error[E0658]: `cfg(doctest)` is experimental and subject to change
  --> src/lib.rs:85:7
   |
85 | #[cfg(doctest)]
   |       ^^^^^^^
   |
   = note: for more information, see https://github.com/rust-lang/rust/issues/62210
```

GuillaumeGomez: It's been implemented in the 1.40 version.

cuviper: Hmm, it would be OK if it were just ineffective before 1.40, but an error is actively harmful. I don't see a lint name to `allow` for this, so I think we'll have to revert it for now.

Alternatively, maybe we can just move this to `tests/`? Then it wouldn't affect the compilation of the main library, and we only run tests in CI on stable/beta/nightly.

GuillaumeGomez: Absolutely! Do you want me to send a PR to fix it?

cuviper: If you can, that would be great, thanks!

GuillaumeGomez: I just thought: if moved in another file, then we'll have to run `rustdoc --test` "by hand" (because `rustdoc` isn't run on the `tests` folder. And my knowledge of github-actions is non-existent. So basically, someone needs to copy the two lines of `doctest` into a file on its own and then run `rustdoc --test` on it.

cuviper: Is there any downside to going back to `cfg(test)`?

GuillaumeGomez: It doesn't get tested since only `rustdoc --test` is testing it, and `cfg(test)` isn't set, preventing `rustdoc` to "see" this and to parse the readme file.

cuviper: So it was never doing anything before this PR? Or did `rustdoc` formerly set `cfg(test)`?

GuillaumeGomez: It did for a few months then we changed it (and this is why I sent the PR in the first place). To test it, add a failing example in the readme then run `cargo test --doc`. :)

cuviper: I moved it to rayon-demo in #763.

PR #763, opened 2020-05-21
Move doctest to the rayon-demo library

cuviper:
We've used `doc_comment::doctest!` to push the README through rustdoc
testing, but its `#[cfg(doctest)]` only works on 1.40+, and causes a
hard error on 1.38 and 1.39. Now we move this into rayon-demo so it
doesn't hurt compatibility of the main library.


cuviper:
ref: https://github.com/rayon-rs/rayon/pull/761#discussion_r428639158

bors r+

GuillaumeGomez:
:+1:

PR #766, opened 2020-06-13
Fix wrong examples of find_map variants

picoHz:
All variants return `first_number` and `find_map_any` is not used 😝


cuviper:
Thanks!

bors r+

PR #771, opened 2020-06-18
impl IntoParallelIterator for Range<char>

### LINKED ISSUE #770: par_iter for Range*<char>
With https://github.com/rust-lang/rust/pull/72413, ranges of `char` are now iterable (in 1.45, releasing 2020-07-16). Rayon should support iterating them in parallel just like normal continuous integer ranges.

[I've done experimentation previously](https://github.com/CAD97/mileage/blob/master/src/range/par_iter.rs), which suggested that `chain(chars before surrogate range, chars after surrogate range)` is the optimal `ParallelIterator` for `Range*<char>`.

I've very little experience with implementing parallel iterators, however, so that could be untrue.

CAD97:
Closes #770

Implements both unindexed and indexed parallel iteration for `Range<char>` and `RangeInclusive<char>`.

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -225,6 +225,27 @@ unindexed_range_impl! {u128, u128}
 unindexed_range_impl! {i128, u128}
 
 // char is special because of the surrogate range hole
+macro_rules! convert_char {
+    ( $self:ident . $method:ident ( $( $arg:expr ),* ) ) => {{
+        let start = $self.range.start as u32;
+        let end = $self.range.end as u32;
+        if start < 0xD800 && 0xE000 < end {
+            // chain the before and after surrogate range fragments
+            (start..0xD800)
+                .into_par_iter()
+                .chain(0xE000..end)
+                .map(|codepoint| unsafe { char::from_u32_unchecked(codepoint) })
+                .$method($( $arg ),*)
+        } else {
+            // no surrogate range to worry about
+            (start..end)
+                .into_par_iter()
+                .map(|codepoint| unsafe { char::from_u32_unchecked(codepoint) })
+                .$method($( $arg ),*)
+        }
+    }};
+}
+
 impl ParallelIterator for Iter<char> {
     type Item = char;
 
@@ -232,7 +253,7 @@ impl ParallelIterator for Iter<char> {
     where
         C: UnindexedConsumer<Self::Item>,
     {
-        self.drive(consumer)
+        convert_char!(self.drive(consumer))
     }
 
     fn opt_len(&self) -> Option<usize> {
@@ -246,22 +267,7 @@ impl IndexedParallelIterator for Iter<char> {
     where
         C: Consumer<Self::Item>,
     {
-        let start = self.range.start as u32;
-        let end = self.range.end as u32;
-        if start < 0xD800 && 0xE000 < end {
-            // chain the before and after surrogate range fragments
-            (start..0xD800)
-                .into_par_iter()
-                .chain(0xE000..end)
-                .map(|codepoint| unsafe { char::from_u32_unchecked(codepoint) })
-                .drive(consumer)
-        } else {
-            // no surrogate range to worry about
-            (start..end)
-                .into_par_iter()
-                .map(|codepoint| unsafe { char::from_u32_unchecked(codepoint) })
-                .drive(consumer)
-        }
+        convert_char!(self.drive(consumer))
     }
 
     fn len(&self) -> usize {
@@ -283,22 +289,7 @@ impl IndexedParallelIterator for Iter<char> {
     where
         CB: ProducerCallback<Self::Item>,
     {
-        let start = self.range.start as u32;
-        let end = self.range.end as u32;
-        if start < 0xD800 && 0xE000 < end {
-            // chain the before and after surrogate range fragments
-            (start..0xD800)
-                .into_par_iter()
-                .chain(0xE000..end)
-                .map(|codepoint| unsafe { char::from_u32_unchecked(codepoint) })
-                .with_producer(callback)
-        } else {
-            // no surrogate range to worry about
-            (start..end)
-                .into_par_iter()
-                .map(|codepoint| unsafe { char::from_u32_unchecked(codepoint) })
-                .with_producer(callback)
-        }
+        convert_char!(self.with_producer(callback))
     }
 }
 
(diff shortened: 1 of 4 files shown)


cuviper:
I'm not so concerned about 16-bit `usize` for rayon -- the only such target built into rust is `msp430-none-elf`, which has `core` only.

src/range.rs
```diff
@@ -223,6 +224,94 @@ unindexed_range_impl! {i64, u64}
...
+        } else {
+            0
+        }
+    }
+
+    fn with_producer<CB>(self, callback: CB) -> CB::Output
+    where
+        CB: ProducerCallback<Self::Item>,
+    {
+        callback.callback(IterProducer { range: self.range })
```
cuviper, line 287: You can use the same trick as `drive` here, then not bother with `IterProducer<char>` at all. That will also make the MSRV of useful `Range<char>` moot.

CAD97:
I've updated the PR to just be for `Range<char>` currently, which works already. The tests only fail because I added a smoke test for correctness across the surrogate range that tests against the std impl, which isn't stable until 1.45.

`RangeInclusive` cannot be implemented in the same manner, as `RangeInclusive<u32>` is not `ExactSizeIterator`, so the chain isn't `IndexedParallelIterator`.

cuviper:
> `RangeInclusive` cannot be implemented in the same manner, as `RangeInclusive<u32>` is not `ExactSizeIterator`, so the chain isn't `IndexedParallelIterator`. (I'm working on doing so only using `Range<u32>` right now.)

Yeah, `end_char as u32 + 1` will never overflow, so `Range<u32>` should be fine.

CAD97:
Alright, this should be correctly functional now. (Just that the tests require rustc 1.45 which is currently beta.)

src/range_inclusive.rs
```diff
@@ -147,6 +148,93 @@ parallel_range_impl! {i64}
...
+}
+
+// Range<u32> is broken on 16 bit platforms, may as well benefit from it
+impl IndexedParallelIterator for Iter<char> {
+    // Split at the surrogate range first if we're allowed to
+    fn drive<C>(self, consumer: C) -> C::Result
+    where
+        C: Consumer<Self::Item>,
+    {
+        if let Some((start, end)) = self.bounds() {
```
cuviper, line 174: Hmm, `bounds()` won't work until Rust 1.45...

FWIW, I was also looking at that function in godbolt, and it works a lot better for `char` this way:

```rust
    if range.size_hint().0 > 0 {
        Some((*range.start(), *range.end()))
    } else {
        None
    }
```
https://rust.godbolt.org/z/8FrRRW

But it does require that we rely on `size_hint()` to be correct (like `TrustedLen`). I guess this code shouldn't be a hotspot anyway, so it probably doesn't matter.

CAD97:
The tests before 56ef498 are more obviously correct (as they compare the rayon and std implementations), but said commit compares them against a provided vector...

but the impl requires `RangeInclusive<char>: Iterator` anyway so that we can get the bounds and whether the iterator is exhausted.

So `RangeInclusive<char>: IntoParallelIterator` seems to need to be gated behind `RangeInclusive<char>: Iterator`, unlike the half-open case.

cuviper:
> So `RangeInclusive<char>: IntoParallelIterator` seems to need to be gated behind `RangeInclusive<char>: Iterator`, unlike the half-open case.

You can add a test to the build script to get a `cfg`.

CAD97:
> You can add a test to the build script to get a `cfg`.

I got that in before you suggested it 🙃

src/range_inclusive.rs
CAD97: The correct way to do it would be to use `.is_empty()` if/when it stabilizes.

That can be optimized in a followup if necessary/desired, I suppose. I don't think it would be _incorrect_ to use `size_hint` here, so long as there's a note about relying on `TrustedLen`-like semantics for correctness (but hopefully not safety?).

CAD97: Side note: oof, there's a unwrap/expect panic still in there. I guess I need to go tweak the impls again to see if I can remove that (as it _should_ be completely unreachable, and if it isn't, that's definitely a bug.)

cuviper:
For the other `RangeInclusive` impls we have a `convert!` macro which hides the code duplication -- maybe you could add a `convert_char!` for your case? You only have two places each, but it's still fairly complicated code.

src/range_inclusive.rs
CAD97: Yep, the generated asm is a _lot_ better when I add in the missed `_unchecked` to `next_back`: https://rust.godbolt.org/z/9dviLq

We already have it for `next` because I spotted the missed optimization out of the unreachable panicking code, but for some reason I didn't think to check reverse iteration for the same thing.

cuviper: > I don't think it would be _incorrect_ to use `size_hint` here, so long as there's a note about relying on `TrustedLen`-like semantics for correctness (but hopefully not safety?).

There's no safety problem, just `start()` and `end()` are _unspecified_ when empty.

CAD97:
Added a macro to deduplicate; I think this is what you were suggesting?

cuviper:
And since I can't help tinkering, here's another option for `bounds` without `Iterator`:

```rust
    let start = *range.start();
    let end = *range.end();
    if start <= end && (start..=end) == *range {
        Some((start, end))
    } else {
        None
    }
```

... assuming `Eq` captures the private exhausted state, I think this would be OK. It actually looks better for `char` than for `i32`! https://rust.godbolt.org/z/TT2DBm

CAD97:
> And since I can't help tinkering, here's another option for `bounds` without `Iterator`:

Here's a comparison of a bunch of the proposed options, with `next_back` fixed to optimize out the unreachable overflow check: https://rust.godbolt.org/z/F25_Ej

TL;DR: it looks to my untrained eyes that any of the proposed options are roughly equivalent. (Note for `_eq2`: `Some` no longer implies there is iteration remaining, but could construct "reverse ranges" when given a reverse range. This will, however, still produce a correctly empty iterator.) I believe the `char` case looks better because of how niches interact with the return value (and thus the difference should disappear when inlined).

CAD97:
The commit just pushed uses `start <= end && range == (start..=end)` to check that using the bounds is still correct. (The `start <= end` check, though redundant for the `char` case, is requisite for the `.chain(once(end))` case. This wasn't caught by tests when I implemented it -- needed test? `u32::MAX..=0` would yield `u32::MAX` without the `start <= end` check.)

cuviper:
I think we're good for this now, thanks!

bors r+

PR #774, opened 2020-06-23
Note that collect_into_vec is in a different trait

SuperFluffy:
I couldn't find `collect_into_vec` until I noticed that it's not in `ParallelIterator`, but in `IndexedParallelIterator`.


cuviper:
Looks good, thanks!

@bors r+

PR #786, opened 2020-08-16
Add flat_map_iter and flatten_iter

cuviper:
The existing `flat_map` and `flatten` adaptors expect items that
implement `IntoParallelIterator`, and they each go through a new round
of parallel task splitting. This often creates smaller task granularity
than people intended, especially if they're doing something simple on
the flattened iterator like `collect()` or `sum()`.

The new `flat_map_iter` and `flatten_iter` expect `IntoIterator` items,
which are simply flattened directly within the current parallel task.
The iterator doesn't even need to be thread safe, because it never
leaves the thread where it is created, only its items.

### DIFF
--- a/src/iter/flatten.rs
+++ b/src/iter/flatten.rs
@@ -1,9 +1,8 @@
 use super::plumbing::*;
 use super::*;
 
-/// `Flatten` turns each element to an iterator, then flattens these iterators
-/// together. This struct is created by the [`flatten()`] method on
-/// [`ParallelIterator`].
+/// `Flatten` turns each element to a parallel iterator, then flattens these iterators
+/// together. This struct is created by the [`flatten()`] method on [`ParallelIterator`].
 ///
 /// [`flatten()`]: trait.ParallelIterator.html#method.flatten
 /// [`ParallelIterator`]: trait.ParallelIterator.html
@@ -13,10 +12,10 @@ pub struct Flatten<I: ParallelIterator> {
     base: I,
 }
 
-impl<I, PI> Flatten<I>
+impl<I> Flatten<I>
 where
-    I: ParallelIterator<Item = PI>,
-    PI: IntoParallelIterator + Send,
+    I: ParallelIterator,
+    I::Item: IntoParallelIterator,
 {
     /// Creates a new `Flatten` iterator.
     pub(super) fn new(base: I) -> Self {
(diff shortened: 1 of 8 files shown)


src/iter/flatten.rs
```diff
@@ -13,32 +12,129 @@ pub struct Flatten<I: ParallelIterator> {
...
+    }
+
+    fn complete(self) -> Self::Result {
+        match self.previous {
+            Some(previous) => previous,
+            None => self.base.into_folder().complete(),
+        }
+    }
 
-        self.base.flat_map(id).drive_unindexed(consumer)
```
cuviper, line 42: I also expanded `flatten` with its own consumer and folder instead of leaning on `flat_map(id)`, both as a micro-optimization and to better mirror the new `flatten_iter` implementation.

nikomatsakis (approved):
This looks good to me! I wonder if we should have made this the primary `flat_map` but can't change that now I suppose. (Or at least it's not worth a 2.x release for)

cuviper:
Yeah, if we could go back, I think I'd call it `par_flat_map` for nested parallelism. I suppose we could add that and deprecate `flat_map`, but then this PR should get a more symmetric name -- maybe `ser_flat_map`? `seq_`? We don't have a good convention yet, because serial is usually the unprefixed case.

nikomatsakis:
Hmm, yeah, the idea of deprecating `flat_map` is appealing, to help steer people towards this new method. I don't love the name `seq_flat_map` but I do agree that having a symmetric naming convention between `par_flat_map` and `seq_flat_map` makes sense.

That said, I feel like `seq_flat_map` could be confusing, as it could suggest that the entire parallel iterator is somehow made sequential, rather than simply that the items from a single iterator are processed together...hmm.

I wonder if `flat_map_par` and `flat_map_seq` is better? (i.e., suffix?) I guess that it's kind of inconsistent with `.par_iter` instead of `.iter`. 

Another idea: `flat_map_par_iter` and `flat_map_iter`?

nikomatsakis:
Oh, I just ignored `ser_flat_map`, but that's because it's kind of unclear to me what `ser` stands for -- serialized? serde? :)

nikomatsakis:
Options:

| parallel | sequential | 
| --- | --- |
| `par_flat_map` | `seq_flat_map` |
| `par_flat_map` | `ser_flat_map` |
| `flat_map_par_iter` | `flat_map_iter` |

I think of these I prefer the first one. I think I favor deprecating the existing methods and recommending the user choose between `seq_flat_map` and `par_flat_map`

cuviper:
These should also be compared with `flatten`:

| parallel | sequential | 
| --- | --- |
| `par_flatten` | `seq_flatten` |
| `par_flatten` | `ser_flatten` |
| `flatten_par_iter` | `flatten_iter` |

I agree we can strike "ser" -- it sounds like the Game of Thrones honorific here, "Ser Flatten"!

I kind of like the suffix for both `flat_map` and `flatten`, as it reads like a verb-object, to flatten an iterator or flatten a parallel iterator.

nikomatsakis:
One other thing I am wondering is whether there is some way to make `par_flat_iter` more efficient, but that would obviously require some deeper edits. =)

nikomatsakis:
Discussed in gitter today. We decided to keep the PR as is. Our reasoning:

* Deprecating the existing method is a bit of a blunt hammer -- we may also eventually be able to make the "to parallel iterator" perform better.
* The "verb-object" name makes sense ("flat map an iterator", "flatten an iterator"), and even if you don't consider performance, it may be important for type reasons to have the option to flat-map a non-parallel-iterator.
* We retain the option of deprecating and adding an explicit name later.

src/iter/mod.rs
```diff
@@ -843,7 +852,47 @@ pub trait ParallelIterator: Sized + Send {
         FlatMap::new(self, map_op)
     }
 
-    /// An adaptor that flattens iterable `Item`s into one large iterator
+    /// Applies `map_op` to each item of this iterator to get nested serial iterators,
+    /// producing a new parallel iterator that flattens these back into one.
+    ///
+    /// Each nested iterator will be flattened local to the thread where the mapping takes
```
nikomatsakis, line 858: Maybe add a section like

### `flat_map` vs `flat_map_iter`

These two methods are similar but behave slightly differently. With `flat_map`, each of the nested iterators must be a parallel iterator, and ultimately we will parallelize all the nested items independently. With `flat_map_iter`, each nested iterator can be a sequential iterator, and we only parallelize *between iterators* (but all items produced by any iterator are processed sequentially).

If the iterators that are being produced tend to have relatively few items, or a consistent number of items, `flat_map_iter` will likely perform better, as it has less parallelization overhead. If you really need to be able to parallelize between the individual items (and not just between iterators), then you should choose `flat_map`.

nikomatsakis (approved):
docs look great!

PR #787, opened 2020-08-16
Add traits ParallelDrainRange and ParallelDrainFull

cuviper:
These define methods of the same name, so only one is expected to be
implemented for any given collection type. Collections that support
ranges can always call `par_drain(..)` for a full drain.

```rust
pub trait ParallelDrainRange<Idx = usize> {
    type Iter: ParallelIterator<Item = Self::Item>;
    type Item: Send;
    fn par_drain<R: RangeBounds<Idx>>(self, range: R) -> Self::Iter;
}

pub trait ParallelDrainFull {
    type Iter: ParallelIterator<Item = Self::Item>;
    type Item: Send;
    fn par_drain(self) -> Self::Iter;
}
```

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -1,3 +1,5 @@
+use std::ops::{Bound, Range, RangeBounds};
+
 /// Divide `n` by `divisor`, and round up to the nearest integer
 /// if not evenly divisable.
 #[inline]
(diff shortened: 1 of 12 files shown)


cuviper:
> These define methods of the same name,

I did this so these traits can be seen as extension traits mirroring `par_drain` to each collection's inherent `drain`, but some take an argument (like `ParallelDrainRange`) and some always drain everything (like `ParallelDrainFull`).

src/math.rs
```diff
@@ -10,6 +12,30 @@ pub(super) fn div_round_up(n: usize, divisor: usize) -> usize {
...
+    let start = match range.start_bound() {
+        Bound::Unbounded => 0,
+        Bound::Included(&i) if i <= len => i,
+        Bound::Excluded(&i) if i < len => i + 1,
+        bound => panic!("range start {:?} should be <= length {}", bound, len),
+    };
+    let end = match range.end_bound() {
+        Bound::Unbounded => len,
+        Bound::Excluded(&i) if i <= len => i,
+        Bound::Included(&i) if i < len => i + 1,
```
nikomatsakis, line 26: should we worry about overflow? or at least do `check_add(1).unwrap()`?

nikomatsakis: (I feel like if we generally assume no overflow, then we should've made `1..2` and `1..=1` represented by the same data structures...oh well...)

nikomatsakis (approved):
I left one nit but overall this PR reads great, good idea to support `drain`

src/math.rs
cuviper: With `i < len`, `i + 1` will never overflow.

nikomatsakis: OK, I see, so we *are* doing checked addition, so to speak.

nikomatsakis (approved):

src/math.rs
cuviper: Right, we're checking the bounds of the slice, and that implicitly proves the `MAX` bound too.

PR #788, opened 2020-08-17
Add a parallel iterator of matching positions

cuviper:
This is the plural of the `position` methods, where `positions` finds
_all_ matching items and returns their indices in a new iterator. This
matches the API of `Itertools::positions`, but in parallel.


nikomatsakis (changes requested):
The code looks right, but I don't really see any tests with non-trivial predicates, can we add some?

cuviper:
@nikomatsakis I don't have non-trivial unit tests, but I think the doc-tests should suffice.

nikomatsakis:
Agreed!

PR #789, opened 2020-08-17
Add array-based chunks and windows from slices

cuviper:
```rust
pub trait ParallelSlice<T: Sync> {
    fn par_array_windows<const N: usize>(&self) -> ArrayWindows<'_, T, N>;
    fn par_array_chunks<const N: usize>(&self) -> ArrayChunks<'_, T, N>;
}
pub trait ParallelSliceMut<T: Send> {
    fn par_array_chunks_mut<const N: usize>(&mut self) -> ArrayChunksMut<'_, T, N>;
}
```


src/slice/mod.rs
```diff
@@ -67,6 +67,22 @@ pub trait ParallelSlice<T: Sync> {
...
+    /// length `N`. The windows overlap.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    /// let windows: Vec<_> = [1, 2, 3].par_array_windows().collect();
+    /// assert_eq!(vec![&[1, 2], &[2, 3]], windows);
+    /// ```
+    fn par_array_windows<const N: usize>(&self) -> ArrayWindows<'_, T, N> {
```
nikomatsakis, line 80: clearly we're not going to land with `const N: usize` in here, right?

src/slice/mod.rs
```diff
@@ -115,6 +131,37 @@ pub trait ParallelSlice<T: Sync> {
...
+        assert_ne!(N, 0);
+        let slice = self.as_parallel_slice();
+        let len = slice.len() / N;
+        let (fst, snd) = slice.split_at(len * N);
+        // SAFETY: We cast a slice of `len * N` elements into
+        // a slice of `len` many `N` elements chunks.
+        let array_slice = unsafe {
+            let ptr = fst.as_ptr() as *const [T; N];
+            ::std::slice::from_raw_parts(ptr, len)
+        };
```
nikomatsakis, line 159: Oof, this is .. clever. I mean it makes sense, but for some reason it was hurting my head, I think because I was thinking of this as a `&[&[T]]` but of course that's not what we have here, it's a `&[[T;N]]`.

nikomatsakis (approved):

src/slice/mod.rs
cuviper: Pending stable `min_const_generics`, yes!

Otherwise, how would you do this for arrays? We already have `par_windows` to get slices.

src/slice/mod.rs
cuviper: FWIW, we did it the same way in the standard library [`array_chunks`](https://github.com/rust-lang/rust/pull/74373/files#diff-e8ccaf64ce21f955ccebef33b52158631493a6f0966815a2ebc142d7cd2b5e06R873-R881) and [`array_chunks_mut`](https://github.com/rust-lang/rust/pull/75021/files#diff-e8ccaf64ce21f955ccebef33b52158631493a6f0966815a2ebc142d7cd2b5e06R1067-R1077).

src/slice/mod.rs
cuviper: This is now `build.rs` detected.

cuviper:
My only hesitation is to wait for `std` stabilization: rust-lang/rust#74985 and rust-lang/rust#75027

pittengermd:
Any update on when this pull request might be merged? I would love to use this for video format conversion.

cuviper:
I'm still hoping `std` will settle its API, so we can be consistent in how we replicate it here.

rezat-c:
> My only hesitation is to wait for `std` stabilization: [rust-lang/rust#74985](https://github.com/rust-lang/rust/issues/74985) and [rust-lang/rust#75027](https://github.com/rust-lang/rust/issues/75027)

[rust-lang/rust#74985](https://github.com/rust-lang/rust/issues/74985) and [rust-lang/rust#75027](https://github.com/rust-lang/rust/issues/75027) were closed.

cuviper:
> > My only hesitation is to wait for `std` stabilization: [rust-lang/rust#74985](https://github.com/rust-lang/rust/issues/74985) and [rust-lang/rust#75027](https://github.com/rust-lang/rust/issues/75027)
> 
> [rust-lang/rust#74985](https://github.com/rust-lang/rust/issues/74985) and [rust-lang/rust#75027](https://github.com/rust-lang/rust/issues/75027) were closed.

Right, but the first one (chunks) was closed as not planned, removed in rust-lang/rust#143289. I think the intention is that you can use `slice::as_chunks` or `as_chunks_mut` first, then iterate as normal. This ought to work with parallel iterators too, so maybe that's how we should leave it.

Rust 1.94.0 did stabilize `array_windows` though!

cuviper:
I published `par_array_windows` in rayon 1.12.0, and I'm going to close on chunks. If the standard library ever reconsiders, so can we. :)

Issue #801, opened 2020-09-17
`partition_map` with a fallible closure - how to shortcircuit?


SuperFluffy:
I am trying to do something like the code below which works fine if I choose to ignore the `Err` coming out of `fallible_calculation`. However, I do not know how to short-circuit the calculation and turn `(Vec<T>, Vec<U>)` into `Result<(Vec<T>, Vec<U>), E>`.

```rust
let res: (Vec<T>, Vec<U>) = us
    .par_iter()
    .partition_map(|u match fallible_calculation(u) {
        Ok(t) if cond(t) => Either::Right(u.clone()),
        Ok(t) => Either::Left(t),
        Err(SpecificError) => Either::Right(u.clone()),
        Err(e) => {},
    })
    .collect()
```

How would I go about properly handling the error case? In the case of the `map` adapter the issue is pretty straight forward, but I don't see how can be done in this case, other than maybe collecting to a `(Vec<T>, Result<Vec<U>, E>>)` instead, which however doesn't seem to short circuit?

cuviper:
> I do not know how to short-circuit the calculation and turn `(Vec<T>, Vec<U>)` into `Result<(Vec<T>, Vec<U>), E>`.

I think you could do this today by mimicking `FromParallelIterator<Result<T, E>> for Result<C, E>`:
https://github.com/rayon-rs/rayon/blob/4a26ac43a7a995870b9b9a737c1d8e3960e0e335/src/result.rs#L88-L132

Your `map` in this case would create `Option<Either<L, R>>`, saving the error aside. Then `.while_some()` to handle short-circuiting, and `.partition_map(identity)` to split your left/right collections.

> maybe collecting to a `(Vec<T>, Result<Vec<U>, E>>)` instead, which however doesn't seem to short circuit?

That's controlled by `full`, which only stops when both sides say so:
https://github.com/rayon-rs/rayon/blob/4a26ac43a7a995870b9b9a737c1d8e3960e0e335/src/iter/unzip.rs#L388-L391

---

A while ago, I was also experimenting with various `FromParallelIterator` for tuples as a followup to #604. That included `FromParallelIterator<Either<L, R>> for (A, B)` as a `partition_map`. I think with that in place, you could map your fallible calculation to `Result<Either<T, U>, E>`, then just `collect` to `Result<(Vec<T>, Vec<U>), E>` as you wanted. Now that you've presented a use case, I'll see if I can dig up that branch and make a pull request.

#721 is somewhat related too.

[...]

PR #802, opened 2020-09-17
impl FromParallelIterator for tuple pairs

cuviper:
Like #604 for `ParallelExtend`, implementing `FromParallelIterator` for tuple pairs opens up new possibilities for nesting `collect`. The possibility of short-circuiting into `Result<(T, U), E>` for #801 is particularly motivating.

- `FromParallelIterator<(A, B)> for (FromA, FromB)` works like `unzip`
- `FromParallelIterator<Either<L, R>> for (A, B)` works like `partition_map`

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -1966,6 +1966,79 @@ pub trait ParallelIterator: Sized + Send {
     ///
     /// assert_eq!(sync_vec, async_vec);
     /// ```
+    ///
+    /// You can collect a pair of collections like [`unzip`](#method.unzip)
+    /// for paired items:
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    ///
+    /// let a = [(0, 1), (1, 2), (2, 3), (3, 4)];
+    /// let (first, second): (Vec<_>, Vec<_>) = a.into_par_iter().collect();
+    ///
+    /// assert_eq!(first, [0, 1, 2, 3]);
+    /// assert_eq!(second, [1, 2, 3, 4]);
+    /// ```
+    ///
+    /// Or like [`partition_map`](#method.partition_map) for `Either` items:
+    ///
+    /// ```
+    /// use rayon::prelude::*;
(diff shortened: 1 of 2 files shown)


nikomatsakis:
@cuviper 

> The possibility of short-circuiting into Result<(T, U), E> for #801 is particularly motivating.

Can you explain this a bit more? Is that potential future work or did I miss something.

cuviper:
It's a composition. `Result<X, E>` implements a short-circuiting `FromParallelIterator` for any `X: FromParallelIterator`. With this PR, that `X` can be `(T, U)`, so we now have a short-circuiting `unzip`/`partition_map` via `collect::<Result<(T, U), E>>()`.

nikomatsakis:
I see.

cuviper:
I finally got around to adding doc-examples for this.

nikomatsakis:
I think this will not get accepted in the stdlib, right?

cuviper:
> I think this will not get accepted in the stdlib, right?

I expect not, but I haven't even proposed it yet. I think given how `IntoIterator` rust-lang/rust#78204 went, it's unlikely that `FromIterator` would be accepted. But I also think it's unlikely that `FromIterator` for tuples would ever be added with _different_ semantics, so I'm not worried about us being incompatible.

And I do think it's still justified here, because it's otherwise really hard to collect multiple things in parallel. With serial `Iterator`, it's trivial to just write a manual `for` loop that unpacks items and stuffs them into however many collections you want. With `ParallelIterator`, collection is much more complicated to do manually, probably best as `fold`+`reduce`, or `try_fold`+`try_reduce` if you want short-circuiting. Yet this PR is a relatively small addition that enables a lot of power easily.

nikomatsakis:
That makes sense.

nikomatsakis (approved):

nikomatsakis:
I'd be r+ on landing this.

cuviper:
bors r=nikomatsakis

PR #808, opened 2020-10-18
Update glium so that rayon-demo runs on Gnome Wayland

mrksu:
Currently, the `nbody` demo fails to open in the Gnome 3.38 Wayland session:

```
rayon-demo]$ cargo run --release -- nbody visualize

[wayland-client error] Attempted to dispatch unknown opcode 0 for wl_shm, aborting.
```

The problem has already been fixed in `glium` 0.28.0. After updating the crate version, the demo runs as expected.


cuviper:
Thanks! I updated the CI lockfile as well to get that passing.

bors r+

PR #811, opened 2020-10-24
Switch par-iter to reentrant mutex to avoid deadlock

untitaker:
Before I invest greatly in tests and having a clean, nice API that does not rely on pinky-promises I just want to get some verification that this approach is safe for fixing #690. It did fix the issue for me.

### DIFF
--- a/src/iter/par_bridge.rs
+++ b/src/iter/par_bridge.rs
@@ -1,12 +1,17 @@
 use crossbeam_deque::{Steal, Stealer, Worker};
 
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::sync::{Mutex, TryLockError};
+use std::sync::{RwLock, TryLockError};
 use std::thread::yield_now;
 
 use crate::current_num_threads;
 use crate::iter::plumbing::{bridge_unindexed, Folder, UnindexedConsumer, UnindexedProducer};
 use crate::iter::ParallelIterator;
+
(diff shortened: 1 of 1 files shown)


src/iter/par_bridge.rs
```diff
@@ -1,13 +1,18 @@
...
 
 use crate::current_num_threads;
 use crate::iter::plumbing::{bridge_unindexed, Folder, UnindexedConsumer, UnindexedProducer};
 use crate::iter::ParallelIterator;
 
+/// A wrapper around rwlock where the user basically pinky-promises to never call read()
+struct ReentrantMutex<T>(RwLock<T>);
+unsafe impl<T> Send for ReentrantMutex<T> where T: Send {}
+unsafe impl<T> Sync for ReentrantMutex<T> where T: Send {}
+
```
untitaker, line 15: this stuff is copypasted from mutex guarantees. I *think* this is safe, as I don't think reentrancy on a `!Sync` T causes problems as long as you're in the same thread?

untitaker: Yeah I think this is safe: https://amanieu.github.io/parking_lot/parking_lot/struct.ReentrantMutex.html

untitaker:
I just realized that rwlock is not actually reentrant. I'm not entirely sure but it definetly fixes the issue for me in practice. In that case perhaps we can still discuss whether adding some sort of reentrant lock here could unblock use of par_bridge with jwalk, but this PR is obsolete.

PR #814, opened 2020-11-17
feat: Simplify `leak` by using `Box::leak`

Hywan:
Thanks for the great project!

I thing `leak` can be rewritten by using `Box::leak`.

Bonus: It removes one `unsafe` block.


Hywan:
bors try

cuviper:
Seems good -- I think the only reason we didn't use `Box::leak` was because it was only added later, in Rust 1.26.

But let's see where we go with #815 first...

PR #815, opened 2020-11-17
feat: Avoid leaking the registry

Hywan:
Hi!

I'm working for Wasmer (https://github.com/wasmerio/wasmer), and we are using `rayon`. Thanks for the awesome project!

We have a memory leak with `rayon`. Valgrind gives us the following trace:

```
==1319== 8 bytes in 1 blocks are definitely lost in loss record 1 of 58
==1319==    at 0x102025635: malloc (in /usr/local/Cellar/valgrind/HEAD-6049595/lib/valgrind/vgpreload_memcheck-amd64-darwin.so)
==1319==    by 0x1001D35D6: alloc::alloc::alloc (alloc.rs:74)
==1319==    by 0x1001D3689: alloc::alloc::Global::alloc_impl (alloc.rs:153)
==1319==    by 0x1001D43FA: <alloc::alloc::Global as core::alloc::AllocRef>::alloc (alloc.rs:203)
==1319==    by 0x1001D3535: alloc::alloc::exchange_malloc (alloc.rs:281)
==1319==    by 0x1001D4B93: rayon_core::util::leak (boxed.rs:175)
==1319==    by 0x1001C2F90: rayon_core::registry::set_global_registry::{{closure}}::{{closure}} (registry.rs:196)
==1319==    by 0x1001EFB78: core::result::Result<T,E>::map (result.rs:508)
==1319==    by 0x1001C2FE8: rayon_core::registry::set_global_registry::{{closure}} (registry.rs:195)
==1319==    by 0x1001DAC6B: std::sync::once::Once::call_once::{{closure}} (once.rs:265)
==1319==    by 0x100BEFDB0: std::sync::once::Once::call_inner (once.rs:421)
==1319==    by 0x1001DABF7: std::sync::once::Once::call_once (once.rs:265)
```

While trying to fix the leak, I noticed a code that could be optimized, hence #814.

But shortly after, I then came with this patch that **actually fixes the memory leak**. An immediate consequence is that `util.rs` is no longer required and #814 is useless. Feel free to close #814 if you consider this patch is a good candidate to land in the main branch of this repository.

# How to reproduce the memory leak

Follow the `README.md` at https://github.com/Hywan/rayon-fix-registry-leak. It contains a usecase you can execute freely.


rayon-core/src/registry.rs
```diff
@@ -159,23 +158,28 @@ pub(super) struct Registry {
...
 /// initialization has not already occurred, use the default
 /// configuration.
-fn global_registry() -> &'static Arc<Registry> {
+fn global_registry() -> Arc<Registry> {
     set_global_registry(|| Registry::new(ThreadPoolBuilder::new()))
-        .or_else(|err| unsafe { THE_REGISTRY.ok_or(err) })
+        .or_else(|err| unsafe {
+            THE_REGISTRY
+                .as_ref()
+                .map(|registry| registry.clone())
```
cuviper, line 172: I think we wanted to avoid the `Arc::clone` in cases that don't actually need an "owned" handle. It's potentially a performance issue to add those atomic count updates, although I don't think this is called in many places.

Another possibility is to replace `THE_REGISTRY` and `THE_REGISTRY_SET` with something like `THE_REGISTRY: once_cell::sync::OnceCell<Arc<Registry>>` to deal with initialization. Those accessors do return a reference too, which can be `&'static T` when coming from a `static` variable.

cuviper:
(meta: I think your description meant to reference #814).

Hywan:
> (meta: I think your description meant to reference #814).

Correct. Sorry for the mistake. Fixed.

rayon-core/src/registry.rs
Hywan: I was hesistant to use `OnceCell` at first: I was afraid to introduce a new dependency. Since it's part of your proposals, I've updated the code to use `OnceCell` accordingly.

Hywan:
Tests are failing with:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', library/test/src/lib.rs:473:32
```

I don't see the link with `OnceCell`. I'm investigating.

Hywan:
I've rebased my attempts into a single commit. I've adopted the same approach than my first commit, however we don't clone the `Arc`. We return a `&'static Arc<…>` by using `Option::as_ref()`.

rayon-core/src/registry.rs
Hywan: `OnceCell` was generated an issue when testing. I've found another solution based on the first one, but without cloning the `Arc`.

cuviper (commented):
I think the resource problem in CI was from unconditionally calling `registry`, creating a whole bunch of threads even if `THE_REGISTRY` is already initialized. I would expect using `get_or_try_init` instead.

rayon-core/src/registry.rs
```diff
@@ -191,15 +190,15 @@ where
...
-            let registry = leak(registry);
-            unsafe {
+        result = registry()
+            .map(|registry: Arc<Registry>| unsafe {
                 THE_REGISTRY = Some(registry);
-            }
-            registry
-        });
+            })
+            .map(|_| unsafe { THE_REGISTRY.as_ref().unwrap() })
```
cuviper, line 199: I think we can fold these two maps together using `THE_REGISTRY.get_or_insert(registry)`, knowing that it's already `None`, and then we don't need an `unwrap`.

Hywan:
> I think the resource problem in CI was from unconditionally calling `registry`, creating a whole bunch of threads even if `THE_REGISTRY` is already initialized.

Yes, correct, but I assumed this behavior could happen in the real life. I prefered to find an alternative.

> I would expect using `get_or_try_init` instead.

`get_or_try_init` does not generate an error when the cell is already initialized. It's required in this case.

cuviper:
> `get_or_try_init` does not generate an error when the cell is already initialized. It's required in this case.

No, but you can set a local boolean whether the closure was actually called.

I'm fine with the new approach that does it without `once_cell` though.

rayon-core/src/registry.rs
Hywan: Indeed!

cuviper:
I discussed this with @nikomatsakis because I wanted to be sure there wasn't some historical reason why this _needed_ to be leaked, but we couldn't come up with anything. It's been a leaked reference ever since it was introduced in [here](https://github.com/rayon-rs/rayon/commit/13110f6415c1020823c26fe7257de191d5970b27#diff-7a8877a46c60729e5e7146831bce1735583200fb3c5cafcb8c8b75071dfa7f29R28), but there was no explanation why it must be that way. So let's go ahead and change that!

bors r+

Hywan:
Thanks :-). It's better to double-check, that's my first patch to `rayon`!

PR #823, opened 2021-01-30
Fix broken doc links in iter module

Qwaz:
The doc links in the iter module appears broken on docs.rs:
https://docs.rs/rayon/1.5.0/rayon/iter/trait.IndexedParallelIterator.html?search=#tymethod.with_producer

This PR updates the links to point to the correct location.


cuviper:
Thanks!

bors r+

PR #831, opened 2021-03-02
added by_blocks method

wagnerf42:
hi,

here is the code for blocks. in draft request like we discussed last time.

### DIFF
--- a/src/iter/blocks.rs
+++ b/src/iter/blocks.rs
@@ -1,0 +1,90 @@
+use super::plumbing::*;
+use super::*;
+
+/// `ByBlocks` is a parallel iterator that consumes itself as a sequence
+/// of parallel blocks.
+///
+/// This struct is created by the [`by_blocks()`] method on [`IndexedParallelIterator`]
+///
+/// [`by_blocks()`]: trait.IndexedParallelIterator.html#method.by_blocks
+/// [`IndexedParallelIterator`]: trait.IndexedParallelIterator.html
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+#[derive(Debug, Clone)]
+pub struct ByBlocks<I, S> {
+    base: I,
+    sizes: S,
+}
+
+impl<I, S> ByBlocks<I, S>
+where
+    I: IndexedParallelIterator,
+    S: IntoIterator<Item = usize>,
+{
+    /// Creates a new `ByBlocks` iterator.
+    pub(super) fn new(base: I, sizes: S) -> Self {
+        ByBlocks { base, sizes }
+    }
+}
+
+struct BlocksCallback<S, C> {
+    sizes: S,
+    consumer: C,
+    len: usize,
+}
+
+impl<T, S, C> ProducerCallback<T> for BlocksCallback<S, C>
+where
+    C: UnindexedConsumer<T>,
+    S: Iterator<Item = usize>,
+{
+    type Output = C::Result;
+    fn callback<P: Producer<Item = T>>(mut self, mut producer: P) -> Self::Output {
+        let mut remaining_len = self.len;
+        let mut consumer = self.consumer;
+        // we need a local variable for the accumulated results
+        // we call the reducer's identity by splitting at 0
+        let (left_consumer, right_consumer, _) = consumer.split_at(0);
+        let mut leftmost_res = left_consumer.into_folder().complete();
+        consumer = right_consumer;
+
+        // now we loop on each block size
+        while remaining_len > 0 && !consumer.full() {
+            // we compute the next block's size
+            let size = self.sizes.next().unwrap_or(std::usize::MAX);
+            let capped_size = remaining_len.min(size);
+            remaining_len -= capped_size;
+
+            // split the producer
+            let (left_producer, right_producer) = producer.split_at(capped_size);
+            producer = right_producer;
+            // split the consumer
+            let (left_consumer, right_consumer, _) = consumer.split_at(capped_size);
+            consumer = right_consumer;
+            leftmost_res = consumer.to_reducer().reduce(
+                leftmost_res,
+                bridge_producer_consumer(capped_size, left_producer, left_consumer),
+            );
+        }
+        leftmost_res
+    }
+}
+
+impl<I, S> ParallelIterator for ByBlocks<I, S>
+where
+    I: IndexedParallelIterator,
+    S: Iterator<Item = usize> + Send,
+{
+    type Item = I::Item;
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+    where
+        C: UnindexedConsumer<Self::Item>,
+    {
+        let len = self.base.len();
+        let callback = BlocksCallback {
+            consumer,
+            sizes: self.sizes,
+            len,
+        };
+        self.base.with_producer(callback)
+    }
+}
--- a/src/iter/blocks.rs
+++ b/src/iter/blocks.rs
@@ -1,27 +1,14 @@
 use super::plumbing::*;
 use super::*;
 
-/// `ByBlocks` is a parallel iterator that consumes itself as a sequence
-/// of parallel blocks.
-///
-/// This struct is created by the [`by_uniform_blocks()`]
-/// and the [`by_doubling_blocks`] methods on [`IndexedParallelIterator`]
-///
-/// [`by_uniform_blocks()`]: trait.IndexedParallelIterator.html#method.by_uniform_blocks
-/// [`by_doubling_blocks()`]: trait.IndexedParallelIterator.html#method.by_doubling_blocks
-/// [`IndexedParallelIterator`]: trait.IndexedParallelIterator.html
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
 #[derive(Debug, Clone)]
-pub struct ByBlocks<I, S> {
+struct ByBlocks<I, S> {
     base: I,
     sizes: S,
 }
 
-impl<I, S> ByBlocks<I, S>
-where
-    I: IndexedParallelIterator,
-    S: IntoIterator<Item = usize>,
-{
+impl<I, S> ByBlocks<I, S> {
     /// Creates a new `ByBlocks` iterator.
     pub(super) fn new(base: I, sizes: S) -> Self {
         ByBlocks { base, sizes }
@@ -40,6 +27,7 @@ where
     S: Iterator<Item = usize>,
 {
     type Output = C::Result;
+
     fn callback<P: Producer<Item = T>>(mut self, mut producer: P) -> Self::Output {
         let mut remaining_len = self.len;
         let mut consumer = self.consumer;
@@ -77,6 +65,7 @@ where
     S: Iterator<Item = usize> + Send,
 {
     type Item = I::Item;
+
     fn drive_unindexed<C>(self, consumer: C) -> C::Result
     where
         C: UnindexedConsumer<Self::Item>,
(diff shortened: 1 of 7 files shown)


src/iter/blocks.rs
```diff
@@ -0,0 +1,90 @@
...
+        // we need a local variable for the accumulated results
+        // we call the reducer's identity by splitting at 0
+        let (left_consumer, right_consumer, _) = consumer.split_at(0);
+        let mut leftmost_res = left_consumer.into_folder().complete();
+        consumer = right_consumer;
+
+        // now we loop on each block size
+        while remaining_len > 0 && !consumer.full() {
+            // we compute the next block's size
+            let size = self.sizes.next().unwrap_or(std::usize::MAX);
```
cuviper, line 53: There's a design choice here when `sizes.next()` returns `None`, with at least three options:

- Consume the remainder as one entire block, which is what you have now.
- Run the remainder for side effects, without feeding it into our consumer. `Skip` does this with its skipped prefix to match the semantics of `std::iter::Skip`.
- Ignore the remainder altogether, basically acting like the suffix of `Take`.

Did you consider this? I think behavior like `Take` might actually be a nice choice.

Another question is whether `size == 0` should be allowed. At the very least, it would be a wasted split here, which could just be considered the user's poor choice. Or we could make that impossible by having `S` produce `NonZeroUsize` items, but that makes it more annoying to write the iterator.

src/iter/mod.rs
```diff
@@ -2169,6 +2176,39 @@ impl<T: ParallelIterator> IntoParallelIterator for T {
...
+    /// all**.
+    /// `by_blocks` on the other hand will start with the leftmost range from 0 to 10, continue
+    /// with 10 to 30, the 30 to 70, 70 to 150, ...
+    /// Each subrange is treated in parallel, while all subranges are treated sequentially.
+    /// We therefore ensure a logarithmic number of blocks (and overhead) while guaranteeing
+    /// we stop at the first block containing the searched data.
+    fn by_blocks<S: IntoIterator<Item = usize>>(self, sizes: S) -> ByBlocks<Self, S> {
+        ByBlocks::new(self, sizes)
+    }
     /// Collects the results of the iterator into the specified
```
cuviper, line 2212: tiny nit: please add a blank line between items like the two functions here, and also between the `type` and `fn` in your trait implementations.

cuviper (commented):
I think this is a nice and powerful abstraction, but I also think we could make it easier for cases we expect to be common. Perhaps there could be a couple helpers like:

```rust
    // double each time, like your example, but this name is so-so
    fn by_doubling_blocks(self, first: usize) -> ByDoublingBlocks<I>; 

    // always use the same size, like `by_blocks(repeat(size))`
    fn by_uniform_blocks(self, size: usize) -> ByUniformBlocks<I>;
```

Those ideas could still use the same `BlocksCallback` internally.

nikomatsakis (approved):

src/iter/blocks.rs
nikomatsakis: Hmm. It seems quite surprising to me to ignore some of the elements. I would definitely not expect that. Ultimately, I think we want to offer users a choice of how to manage this -- another behavior I could imagine would be "go on using the last block size that got returned until we are complete".

I am wondering if we can express these semantics with some composed operators? I'd sort of like to say something like `.by_blocks().skip_remainder()` or something. We could just add inherent methods to the return type to allow users to change back and forth between those modes, no?

cuviper: It seems to me that the user can express most of those semantics in the iterator itself:

- current MAX behavior: `my_iter.chain(iter::once(usize::MAX))`
- repeat behavior: `my_iter.chain(iter::repeat(last))`
- my suggested `Take` behavior: just `my_iter`, and let its `None` be the end
- run remainder without consuming: ???

> We could just add inherent methods to the return type to allow users to change back and forth between those modes, no?

Yeah, that's possible too.

nikomatsakis: another might be asserting that there is no remainder

nikomatsakis: I think at minimum we should document these techniques, and i might prefer a convenient way to express them

wagnerf42:
> I think this is a nice and powerful abstraction, but I also think we could make it easier for cases we expect to be common. Perhaps there could be a couple helpers like:
> 
> ```rust
>     // double each time, like your example, but this name is so-so
>     fn by_doubling_blocks(self, first: usize) -> ByDoublingBlocks<I>; 
> 
>     // always use the same size, like `by_blocks(repeat(size))`
>     fn by_uniform_blocks(self, size: usize) -> ByUniformBlocks<I>;
> ```
> 
> Those ideas could still use the same `BlocksCallback` internally.

hi, sorry for the delay. that's a good idea. it's the only two patterns i was able to come up with for now.

src/iter/blocks.rs
wagnerf42: if we only allow constant sizes and doubling sizes in the public interface then this choice about what to do when the size iterator ends is of private concern.
i used to have both versions, one like an implicit take when it runs out and the other one forcing to consume everything. so there is indeed something to choose here.

i would favor consuming it all since I happened to get some bugs missing some elements.

wagnerf42: also, i have some questions like:
- should i add unit tests ? where ?
- should i add benches ? where ?

nikomatsakis: I would be in favor of only allowing the two options until we know there is demand for something more general. As for unit tests, we typically just add `#[test]` in the modules, I think? Alternatively, you can add tests into the `tests` directory but we don't usually do that. (Is this right, @cuviper?)

For benchmarks, we generally modify the rayon-bench project.

cuviper: For tests, there are a few generic sanity checks to add: `tests/clones.rs`, `tests/debug.rs`, and `src/compile_fail/must_use.rs`. Beyond that, adding unit `#[test]` directly in the module is fine, or break into a `tests` submodule if there are a lot. Doc-test examples are also nice to serve both doc and testing roles.

src/iter/mod.rs
wagnerf42: sorry i just saw this comment. in the meantime i changed the file so i'm not too sure what you are referring to.

wagnerf42:
hi, so I did the changes to by_doubling_blocks and by_uniform_blocks.

i also removed the initial block size argument in by_doubling_blocks to make it more transparent to the user. tell me if you think it is not a good idea.

quick questions: 
- is it ok to return in both cases a ByBlocks or should I get two different types and hide the size iterator ?
- is it ok if i do modifications to rayon_demo/src/find ? find_last and find_first are calling find_any

src/iter/mod.rs
cuviper: I'll add inline suggestions for it.

src/iter/blocks.rs
```diff
@@ -0,0 +1,92 @@
...
+    len: usize,
+}
+
+impl<T, S, C> ProducerCallback<T> for BlocksCallback<S, C>
+where
+    C: UnindexedConsumer<T>,
+    S: Iterator<Item = usize>,
+{
+    type Output = C::Result;
+    fn callback<P: Producer<Item = T>>(mut self, mut producer: P) -> Self::Output {
```
cuviper, line 43:
```suggestion
    type Output = C::Result;

    fn callback<P: Producer<Item = T>>(mut self, mut producer: P) -> Self::Output {
```

src/iter/blocks.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+
+impl<I, S> ParallelIterator for ByBlocks<I, S>
+where
+    I: IndexedParallelIterator,
+    S: Iterator<Item = usize> + Send,
+{
+    type Item = I::Item;
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
```
cuviper, line 80:
```suggestion
    type Item = I::Item;

    fn drive_unindexed<C>(self, consumer: C) -> C::Result
```

src/iter/mod.rs
```diff
@@ -2169,6 +2176,63 @@ impl<T: ParallelIterator> IntoParallelIterator for T {
...
+    fn by_doubling_blocks(
+        self,
+    ) -> ByBlocks<Self, std::iter::Successors<usize, fn(&usize) -> Option<usize>>> {
+        let first = crate::current_num_threads();
+        ByBlocks::new(
+            self,
+            std::iter::successors(Some(first), |s| Some(s.saturating_mul(2))),
+        )
+    }
+    /// Normally, parallel iterators are recursively divided into tasks in parallel.
```
cuviper, line 2217:
```suggestion
    }

    /// Normally, parallel iterators are recursively divided into tasks in parallel.
```

src/iter/mod.rs
```diff
@@ -2169,6 +2176,63 @@ impl<T: ParallelIterator> IntoParallelIterator for T {
...
+    ///     .into_par_iter()
+    ///     .by_uniform_blocks(1_000_000)
+    ///     .fold(Vec::new, |mut v, e| { v.push(e); v})
+    ///     .reduce(Vec::new, |mut v1, mut v2| { v1.append(&mut v2); v1});
+    /// assert_eq!(v, (0u32..10_000_000).collect::<Vec<u32>>());
+    /// ```
+    fn by_uniform_blocks(self, blocks_size: usize) -> ByBlocks<Self, std::iter::Repeat<usize>> {
+        ByBlocks::new(self, std::iter::repeat(blocks_size))
+    }
     /// Collects the results of the iterator into the specified
```
cuviper, line 2236:
```suggestion
    }

    /// Collects the results of the iterator into the specified
```

cuviper (commented):
These are the style nits I mentioned, just adding blank lines.

cuviper:
> * is it ok to return in both cases a ByBlocks or should I get two different types and hide the size iterator ?

I think it would be better to hide those details.

> * is it ok if i do modifications to rayon_demo/src/find ? find_last and find_first are calling find_any

Yes, that's fine -- it's probably a good place to showcase the difference between these approaches.

src/iter/mod.rs
wagnerf42: ok this is taking shape.

however there are some functions i had which are not included here.
i'm not too sure if we should add them at least in this pull request.

src/iter/mod.rs
```diff
@@ -2169,6 +2176,61 @@ impl<T: ParallelIterator> IntoParallelIterator for T {
...
+    /// we stop at the first block containing the searched data.
+    fn by_exponential_blocks(self) -> ExponentialBlocks<Self> {
+        ExponentialBlocks::new(self)
+    }
+
+    /// Normally, parallel iterators are recursively divided into tasks in parallel.
+    /// This adaptor changes the default behavior by splitting the iterator into a **sequence**
+    /// of parallel iterators of given `blocks_size`.
+    /// The main application is to obtain better
+    /// memory locality (especially if the reduce operation re-use folded data).
```
nikomatsakis, line 2218: This is good -- I think it'd be good to make a section like

```
# When to use this method
```

and put this text in there, and maybe point at `by_uniform_blocks` as an alternative. I think users will be unclear when to use it.

nikomatsakis (changes requested):

wagnerf42:
some more things which can be done: 
- add a sequential fold or try_fold method as argument to `by_blocks` (well, in a new method).
this way we could try_fold sequentially the results reduced in each block.
it would actually be needed to have the best manual `find_first` implementation.
- have another iterator size based on timings (like: one block every x duration)

i'm a bit worried about the combinations between block sizes and iterator choices because
we would have 2 or 3 iterator choices times 2-3 block sizes choices so that's up to 9 methods.

wagnerf42:
actually it would not work this way because we would need to know what the return type is and this info is not known until the call to reduce. i think it might still be possible to do it but i would need to modify the `Folder` trait. if we add a `seq_consume` for consuming the blocks results then i think we can add a `seq_fold` method in `ParallelIterator` registering in the consumer what the blocks folding method would be. 

i'll give it a try

src/iter/blocks.rs
```diff
@@ -0,0 +1,147 @@
...
+/// `ExponentialBlocks` is a parallel iterator that consumes itself as a sequence
+/// of parallel blocks of increasing sizes (exponentially).
+///
+/// This struct is created by the [`by_exponential_blocks()`] method on [`IndexedParallelIterator`]
+/// [`by_exponential_blocks()`]: trait.IndexedParallelIterator.html#method.by_exponential_blocks
+/// [`IndexedParallelIterator`]: trait.IndexedParallelIterator.html
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+#[derive(Debug, Clone)]
+pub struct ExponentialBlocks<I>(
+    ByBlocks<I, std::iter::Successors<usize, fn(&usize) -> Option<usize>>>,
```
cuviper, line 92: This only _needs_ to store `I`, and then a `ByBlocks` can be created on the fly in `drive_unindexed`. That will also let it use the direct unnameable closure type, rather than forcing it to a function pointer.

src/iter/blocks.rs
```diff
@@ -0,0 +1,147 @@
...
+
+/// `UniformBlocks` is a parallel iterator that consumes itself as a sequence
+/// of parallel blocks of constant sizes.
+///
+/// This struct is created by the [`by_uniform_blocks()`] method on [`IndexedParallelIterator`]
+/// [`by_uniform_blocks()`]: trait.IndexedParallelIterator.html#method.by_uniform_blocks
+/// [`IndexedParallelIterator`]: trait.IndexedParallelIterator.html
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+#[derive(Debug, Clone)]
+pub struct UniformBlocks<I>(ByBlocks<I, std::iter::Repeat<usize>>);
```
cuviper, line 127: Similarly, this really only needs to store `I` and `blocks_size` at first, but in this case it probably doesn't matter either way.

cuviper:
@wagnerf42 did you ever try the suggestion I had about dropping the `I` parameters? Essentially making them lazier about that, IIRC. I also wonder if you were still exploring your own ideas, since this is still in Draft.

wagnerf42:
> @wagnerf42 did you ever try the suggestion I had about dropping the `I` parameters? Essentially making them lazier about that, IIRC. I also wonder if you were still exploring your own ideas, since this is still in Draft.

well, i'm not too sure about what you mean. sorry it's been a while ago.
this adapter is one i use quite often and i'd rather get it done right.

cuviper:
If you have any comments on what I added, please let me know!

I plan to publish a new release soon, and it would be nice to include this.

PR #834, opened 2021-03-18
Add a pub-in-private trait for Range integers

### LINKED ISSUE #833: IntoParallelIterator for Range breaks type inference
Motivated by [this stack overflow question](https://stackoverflow.com/questions/66666909/why-does-filter-have-different-type-requirements-on-parallel-iterators):

```
use rayon::prelude::*;

fn is_even(n: i64) -> bool {
    n % 2 == 0
}

fn main() {
    let v: Vec<_> = (1..300_000_000)
        .into_par_iter() // works correctly without this line
        .filter(|&x| is_even(x))
        .collect();
}
```

```
error[E0271]: type mismatch resolving `<rayon::range::Iter<i32> as rayon::iter::ParallelIterator>::Item == i64`
  --> src/main.rs:11:10
   |
11 |         .filter(|&x| is_even(x))
   |          ^^^^^^ expected `i32`, found `i64`

error[E0271]: type mismatch resolving `<rayon::range::Iter<i32> as rayon::iter::ParallelIterator>::Item == i64`
  --> src/main.rs:12:10
   |
12 |         .collect();
   |          ^^^^^^^ expected `i32`, found `i64`
   |
   = note: required because of the requirements on the impl of `rayon::iter::ParallelIterator` for `rayon::iter::Filter<rayon::range::Iter<i32>, [closure@src/main.rs:11:17: 11:32]>`
```

My assessment was that this was due to how `ParallelIterator` is defined for `range::Iter` and a fix was *possible*, but I don't know enough about rayon internals to know if there's a significant motivating factor for doing it this way.

cuviper:
Fixes #833 type inference by adding traits that are used as a constraint
directly on integer types for ranges and their `Iter` implementations,
rather than implementing each `Iter<{integer}>` separately. The traits
have to be `pub` since they're used by public `impl<T> ... for Iter<T>`,
but they're hidden in a private module to stay out of our semver API.

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -19,10 +19,11 @@
 use crate::iter::plumbing::*;
 use crate::iter::*;
 use std::char;
+use std::convert::TryFrom;
 use std::ops::Range;
 use std::usize;
 
-/// Parallel iterator over a range, implemented for all integer types.
+/// Parallel iterator over a range, implemented for all integer types and `char`.
 ///
 /// **Note:** The `zip` operation requires `IndexedParallelIterator`
 /// which is not implemented for `u64`, `i64`, `u128`, or `i128`.
@@ -48,6 +49,7 @@ pub struct Iter<T> {
     range: Range<T>,
 }
 
+/// Implemented for ranges of all primitive integer types and `char`.
 impl<T> IntoParallelIterator for Range<T>
 where
     Iter<T>: ParallelIterator,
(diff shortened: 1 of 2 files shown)


cuviper:
I don't love the amount of boilerplate indirection added here, but it does appear to help type inference as desired, so I guess we can accept complexity on our side as an ergonomic win for the user.

cuviper:
@nikomatsakis does this seem reasonable to you?
(Both from rayon's perspective and knowing how rustc works.)

cuviper:
bors r=nikomatsakis

PR #836, opened 2021-03-24
Descendants

wagnerf42:
hi,

this is an equivalent of the `split` function but for tree like patterns.
it solves an issue that was posted recently and i could also use it with some of my code.

tell me if you think it is worth inclusion for you. if so i can add some more tests and benches.


src/iter/descendants.rs
```diff
@@ -0,0 +1,209 @@
...
+///            .chain(r.right.as_ref())
+///            .map(|n| &**n)
+///    })
+///    .map(|node| node.content)
+///    .collect();
+///    v.sort_unstable();
+///    assert_eq!(v, vec![3, 10, 14, 18]);
+///    ```
+///
+pub fn descendants<S, B, I>(root: S, breed: B) -> Descendants<S, B, I>
```
nikomatsakis, line 198: I wonder about a different name, like `iter::walk_tree` or something?

nikomatsakis (approved):
This seems definitely useful to me. I feel like an awful lot of parallel patterns wind up descending over trees.

cuviper:
Do I understand correctly that this makes no guarantees about traversal order?

Of course parallel execution is unordered, but it's still visible in fold and reduce. I think it would be a lot stronger if we could specify the traversal and hold to that. It could even be generic pre/in/post-order if the user provided two closures, or one closure returning two things, so they choose what is logically traversed before or after the current item. Then examples like your tree could `collect` in the desired order, rather than sorting afterward.

I'm sure that would make the implementation harder though...

wagnerf42:
> Do I understand correctly that this makes no guarantees about traversal order?
> 
> Of course parallel execution is unordered, but it's still visible in fold and reduce. I think it would be a lot stronger if we could specify the traversal and hold to that. It could even be generic pre/in/post-order if the user provided two closures, or one closure returning two things, so they choose what is logically traversed before or after the current item. Then examples like your tree could `collect` in the desired order, rather than sorting afterward.
> 
> I'm sure that would make the implementation harder though...

i'll take a look. note that it's not for binary trees but for arbitrary degrees so the only meaningful orders are pre and post

wagnerf42:
hi, so i did two functions with two different orderings.
prefix and postfix (order between children also differs in the two functions)
i benchmarked quite some number of variants and stayed with the fastest ones.
it is actually pretty fast. you can try the tree example.

there is an overhead for the prefix order because breed borrows the state so we must consume the whole iterator before consuming s and then loop again on the same data towards the children.

i'd like to have some feedback at this point if you can.
if that's ok for you then i'll add some tests for arbitrary degrees and the debug+clone tests

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,432 @@
...
+            seen: Vec::new(),
+            breed: &self.breed,
+            phantom: PhantomData,
+        };
+        bridge_unindexed(producer, consumer)
+    }
+}
+
+/// Create a tree like parallel iterator from an initial root state.
+/// Thre `breed` function should take a state and iterate on all of its children states.
```
nikomatsakis, line 107:
```suggestion
/// The `children_of` function should take a node and return an iterator over its child nodes.
```

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,432 @@
...
+            to_explore: once(self.initial_state).collect(),
+            seen: Vec::new(),
+            breed: &self.breed,
+            phantom: PhantomData,
+        };
+        bridge_unindexed(producer, consumer)
+    }
+}
+
+/// Create a tree like parallel iterator from an initial root state.
```
nikomatsakis, line 106:
```suggestion
/// Create a tree-like parallel iterator from an initial root node.
```

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,432 @@
...
+/// Thre `breed` function should take a state and iterate on all of its children states.
+/// The best parallelization is obtained when the tree is balanced
+/// but we should also be able to handle harder cases.
+///
+/// This guarantees a prefix ordering.
+/// Between siblings, last bred child comes first.
```
nikomatsakis, line 112:
```suggestion
/// Between siblings, children are reduced in reverse order -- that is, the children that returned last are reduced first.
```

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,432 @@
...
+        bridge_unindexed(producer, consumer)
+    }
+}
+
+/// Create a tree like parallel iterator from an initial root state.
+/// Thre `breed` function should take a state and iterate on all of its children states.
+/// The best parallelization is obtained when the tree is balanced
+/// but we should also be able to handle harder cases.
+///
+/// This guarantees a prefix ordering.
```
nikomatsakis, line 111:
```suggestion
/// # Ordering
///
/// This function guarantees a prefix ordering. See also [`walk_tree_postfix`], which guarantees a postfix order. If you don't care about ordering, you should use `walk_tree`, which will use whatever is believed to be fastest.
```

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,432 @@
...
+    if v.len() <= 1 {
+        None
+    } else {
+        let n = v.len() / 2;
+        Some(v.split_off(n))
+    }
+}
+
+/// Create a tree like parallel iterator from an initial root state.
+/// Thre `breed` function should take a state and iterate on all of its children states.
```
nikomatsakis, line 337: (as above)

nikomatsakis (changes requested):
This looks good to me. I made some documentation suggestions. I think we should introduce a `walk_tree` that simply calls `walk_tree_postfix`, so that people have something they can use that does not guarantee ordering.

src/iter/walk_tree.rs
wagnerf42: ok it's done

wagnerf42:
thanks for all your suggestions. it definitely is much cleaner now.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,533 @@
...
+/// Create a tree like parallel iterator from an initial root node.
+/// Thre `children_of` function should take a node and iterate on all of its child nodes.
+/// The best parallelization is obtained when the tree is balanced
+/// but we should also be able to handle harder cases.
+///
+/// # Ordering
+///
+/// This function does not guarantee any ordering.
+/// See also [`walk_tree_prefix`] which guarantees a
+/// prefix order and [`walk_tree_postfix`] which guarantees a postfix order.
```
nikomatsakis, line 482: We should note that it attempts to do "whatever is fastest"

nikomatsakis (changes requested):
Looks great! One last nit.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,533 @@
...
+/// The best parallelization is obtained when the tree is balanced
+/// but we should also be able to handle harder cases.
+///
+/// # Ordering
+///
+/// This function guarantees a prefix ordering. See also [`walk_tree_postfix`],
+/// which guarantees a postfix order.
+/// If you don't care about ordering, you should use [`walk_tree`],
+/// which will use whatever is believed to be fastest.
+/// Between siblings, children are reduced in reverse order -- that is, the children that returned last are reduced first.
```
cuviper, line 118: The reverse order seems really bizarre -- is that necessary? As a user, I would expect a "normal" pre-order traversal of a binary tree to first visit the node, then the left children, then the right children.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,533 @@
...
+/// For example a perfect binary tree of 7 nodes will reduced in the following order:
+///
+/// ```text
+///      1
+///     / \
+///    /   \
+///   5     2
+///  / \   / \
+/// 7   6 4   3
+/// ```
```
cuviper, line 128: I was honestly confused at first how this tree was supposed to tell me the order, especially since the numbers are not organized like the values of a normally-ordered binary tree. Then I realized that they *aren't* values at all, rather the sequence indices of the traversal order, but I think this needs clarification somehow.

The other tree examples do have values, not indices, so this is inconsistent. I think it would be more clear if you first showed some tree of values, then a flat list of those values in their visited order.

src/iter/walk_tree.rs
wagnerf42: hi, this is the fastest version. since it is a prefix order we need to consume the father before its children. but we need to borrow the father to iterate on children. this means we first need to extend the stack with the children.
because it's a vec the last pushed will be the first poped and this reverses children's order.

other options are:
- use a more complex struct and some unsafe.
- require the iterator to be double ended
- first collect the iterator into a vec and then extend the stack in the right order

to me the only acceptable option is the third one. it comes with a performance hit though.

if you think it would be better to have the standard order then i can re-implement this version.

wagnerf42: ok, so i did a quick bench (the one in rayon_demo)
this is the current version:
sum: 382ns collect: 683ns
this is the version with the "fixed" ordering:
sum: 705ns collect: 950ns

nikomatsakis: That speed difference seems pretty significant to me

cuviper: > first collect the iterator into a vec and then extend the stack in the right order

You could do this in-place by noting the previous length, extending, then reversing the new tail slice. That's still some added cost, but at least it doesn't require a new allocation.

cuviper: I do care about the speed, but that comes after having useful semantics. Reversed child order seems unusual enough to me that I doubt anyone really *wants* that order, and if they don't care they can use the unordered `walk_tree`.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,540 @@
...
+    S: Send,
+    B: Fn(&S) -> I + Send + Sync,
+    I: IntoIterator<Item = S> + Send,
+{
+    type Item = S;
+    fn split(mut self) -> (Self, Option<Self>) {
+        // explore while front is of size one.
+        while self.to_explore.len() == 1 {
+            let front_node = self.to_explore.pop().unwrap();
+            self.to_explore = (self.breed)(&front_node).into_iter().collect();
```
cuviper, line 244: Likewise, this could `extend`.

src/iter/walk_tree.rs
wagnerf42: in my bench i actually did it without an extra allocation. using a buffer vec and draining / extending.
i'm not sure that the actual order is that bad. user can just reverse their iterator if they really care about siblings ordering.
walk_tree has no guarantee while this has

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,540 @@
...
```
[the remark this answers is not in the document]
wagnerf42: this has nothing to do with the sequential folding. this is the task splitting which requires a new vec

src/iter/walk_tree.rs
wagnerf42: > Likewise, this could `extend`.

same answer here

src/iter/walk_tree.rs
wagnerf42: i did it like this :
```
         self.reorder_buffer.extend((self.breed)(&e));
         self.to_explore.extend(self.reorder_buffer.drain(..).rev());
```

wagnerf42: just to be clear, i do understand there is a choice to be made here.
the overhead is between 10 and 15 nanoseconds per element. this is not zero cost but not excessive.
if you ask me i would favor the current (reversed) order because the algorithm is "natural" this way. user can also reverse their iterators to get to the other order. however i do see why the other option also makes sense.

in the end, choose whatever you want and let me know.

cuviper: Ideally, internal algorithm details should not be so visible in the public API, especially since those details might want to change in the future. I want to aim for what is most "natural" to the user first, without exposing details like "well actually you probably want to reverse this." That's even evident in your `tree_prefix_collect` benchmark where you have an explanatory comment, "large indices to the left, small to the right", being different than the others.

To that end, I think `DoubleEndedIterator` might be reasonable, although you immediately dismissed that. The user can provide the children in their natural order, **matching the visitation order**, but we'll have the flexibility to internally iterate in reverse. Are there realistic use-cases that you think could not be double-ended?

Of course there are plenty of non-DE iterators, but I'm asking for a real data structure that could not operate this way, especially since you say "users can also reverse their iterators." If they can so easily reverse, then DE should be no issue; if they can't, then I think we won't be serving them well with an algorithmically-reversed order.

src/iter/walk_tree.rs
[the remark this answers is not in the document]
cuviper: You have the existing allocation in `self.to_explore`, now empty since the last item was just popped. Why throw that away for a newly collected `Vec`, instead of using the buffer you have?

The split will happen below, regardless. So the difference is "collect a new `Vec`, drop the old one, and split a new one" (one `free` and two `malloc`s), versus "reuse the old `Vec` and split a new one" (one `malloc`).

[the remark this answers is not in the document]
wagnerf42: fair enough, i removed the unneeded allocation

src/iter/walk_tree.rs
wagnerf42: so i did the switch to double ended iterators

wagnerf42:
i'll add a bit more tests with higher degrees

nikomatsakis (approved):
This is looking pretty good to me.

wagnerf42:
do i need to do something more ? i'm in no hurry but just tell me if you want something from me.

nikomatsakis:
I'm satisfied -- @cuviper ?

src/iter/walk_tree.rs
cuviper: I think we now just need to update (or remove) this line about reversal.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,545 @@
...
+///    .collect();
+///    assert_eq!(v, vec![10, 3, 14, 18]);
+///    ```
+///
+pub fn walk_tree_prefix<S, B, I, IT>(root: S, breed: B) -> WalkTreePrefix<S, B, I>
+where
+    S: Send,
+    B: Fn(&S) -> I + Send + Sync,
+    IT: DoubleEndedIterator<Item = S>,
+    I: IntoIterator<Item = S, IntoIter = IT> + Send,
```
cuviper, line 218: I think we can remove `IT` as a public API parameter by constraining `I::IntoIter: Double...` instead.

Should we also require `I::IntoIter: Send`? We consume it immediately now, but might there be a future where we'd want to be more lazy about that?

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,545 @@
...
+///    .map(|node| node.content)
+///    .collect();
+///    assert_eq!(v, vec![3, 18, 14, 10]);
+///    ```
+///
+pub fn walk_tree_postfix<S, B, I>(root: S, breed: B) -> WalkTreePostfix<S, B, I>
+where
+    S: Send,
+    B: Fn(&S) -> I + Send + Sync,
+    I: IntoIterator<Item = S> + Send,
```
cuviper, line 468: Same question about maybe `I::IntoIter: Send`.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,545 @@
...
+            });
+        (self, right)
+    }
+    fn fold_with<F>(mut self, mut folder: F) -> F
+    where
+        F: Folder<Self::Item>,
+    {
+        // start by consuming everything seen
+        for s in self.seen {
+            folder = folder.consume(s);
```
cuviper, line 60: This can use `folder.consume_iter(self.seen)` instead of a manual loop, and one `full()` check afterward.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,545 @@
...
+            seen: Vec::new(),
+            breed: &self.breed,
+            phantom: PhantomData,
+        };
+        bridge_unindexed(producer, consumer)
+    }
+}
+
+/// Create a tree-like prefix parallel iterator from an initial root node.
+/// The `children_of` function should take a node and return an iterator over its child nodes.
```
cuviper, line 111: The current revision is calling this function `breed`.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,545 @@
...
+        // now do all remaining explorations
+        for e in self.to_explore {
+            folder = consume_rec_postfix(&self.breed, e, folder);
+            if folder.full() {
+                return folder;
+            }
+        }
+        // end by consuming everything seen
+        for s in self.seen.into_iter().rev() {
+            folder = folder.consume(s);
```
cuviper, line 293: This can also use `consume_iter`.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,545 @@
...
+    if v.len() <= 1 {
+        None
+    } else {
+        let n = v.len() / 2;
+        Some(v.split_off(n))
+    }
+}
+
+/// Create a tree like postfix parallel iterator from an initial root node.
+/// Thre `children_of` function should take a node and iterate on all of its child nodes.
```
cuviper, line 362: Typo "Thre", and this too is now `breed`.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,545 @@
...
+
+/// ParallelIterator for arbitrary tree-shaped patterns.
+/// Returned by the [`walk_tree()`] function.
+///
+/// [`walk_tree()`]: fn.walk_tree_prefix.html
+#[derive(Debug)]
+pub struct WalkTree<S, B, I>(WalkTreePostfix<S, B, I>);
+
+/// Create a tree like parallel iterator from an initial root node.
+/// Thre `children_of` function should take a node and iterate on all of its child nodes.
```
cuviper, line 485: "Thre" and `breed`.

src/iter/walk_tree.rs
```diff
@@ -0,0 +1,545 @@
...
+///     12);
+/// ```
+///
+/// [`walk_tree_prefix()`]: fn.walk_tree_prefix.html
+/// [`walk_tree_postfix()`]: fn.walk_tree_postfix.html
+pub fn walk_tree<S, B, I>(root: S, breed: B) -> WalkTree<S, B, I>
+where
+    S: Send,
+    B: Fn(&S) -> I + Send + Sync,
+    I: IntoIterator<Item = S> + Send,
```
cuviper, line 522: Same question about maybe `I::IntoIter: Send`.

cuviper (commented):
There's some minor cleanup, but otherwise I'm satisfied.

Please also take the PR out of draft state if you think it's ready.

src/iter/walk_tree.rs
wagnerf42: i'm not sure i get it. what would send allow us to do ? IT would not be a parallel iterator so it would just mean using another thread. `I` is already Send.
so i'm not getting what this would potentially allow us to do and why `I` being Send is not enough

cuviper: `I: Send` doesn't actually give us much, only that after getting such a return value of `B`, we could hold and send that `I` value to other threads *before* calling `into_iter()`. Once that's called, we now have the `IntoIter` type, which we might want to lazily pull items out one at a time.

For example, suppose `to_explore` were instead a `Vec<Either<S, IntoIter>>`, storing either unvisited nodes or their unprocessed (or partially processed) brood. We would need `IntoIter: Send` to be able to split this between threads.

Maybe we don't need to do this, especially if we don't expect there to be very many children most of the time.

cuviper: Actually, `I: Send` is only necessary right now because you have that `PhantomData<I>`, but I don't think you actually need that. The structs themselves don't need the `I` parameter at all, only the `impl` constraints.

nikomatsakis:
@wagnerf42 should we take the PR out of draft state?

nikomatsakis:
you gotta click that "ready for review" button:

image

or we can...

cuviper:
Hi @wagnerf42! Sorry that it's been a while. I took the liberty of rebasing your PR and making a few of those tweaks I had suggested way back in my last review. Please take a look, and if you're happy with this then I think we should merge it!

wagnerf42:
hi, well it seems fine for me.

cuviper:
Thanks!

PR #837, opened 2021-03-26
impl IntoParallelIterator for arrays

cuviper:
This requires Rust 1.51, as tested in the build script.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,4 +3,7 @@ fn main() {
     if ac.probe_expression("(0..10).step_by(2).rev()") {
         autocfg::emit("step_by");
     }
+    if ac.probe_expression("{ fn foo<const N: usize>() {} }") {
+        autocfg::emit("min_const_generics");
+    }
 }
--- a/src/array.rs
+++ b/src/array.rs
@@ -1,0 +1,90 @@
+#![cfg(min_const_generics)]
+//! Parallel iterator types for [arrays] (`[T; N]`)
+//!
+//! You will rarely need to interact with this module directly unless you need
+//! to name one of the iterator types.
+//!
+//! Everything in this module requires const generics, stabilized in Rust 1.51.
+//!
+//! [arrays]: https://doc.rust-lang.org/std/primitive.array.html
+
+use crate::iter::plumbing::*;
+use crate::iter::*;
+use crate::slice::{Iter, IterMut};
+use crate::vec::DrainProducer;
+use std::mem::ManuallyDrop;
+
+/// This implementation requires const generics, stabilized in Rust 1.51.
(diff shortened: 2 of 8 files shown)


build.rs
```diff
@@ -3,4 +3,7 @@ fn main() {
     if ac.probe_expression("(0..10).step_by(2).rev()") {
         autocfg::emit("step_by");
     }
+    if ac.probe_expression("{ fn foo<const N: usize>() {} }") {
+        autocfg::emit("min_const_generics");
+    }
```
nikomatsakis, line 8: I...love? that we reinvented autoconf

src/array.rs
```diff
@@ -0,0 +1,90 @@
...
+use crate::vec::DrainProducer;
+use std::mem::ManuallyDrop;
+
+/// This implementation requires const generics, stabilized in Rust 1.51.
+impl<'data, T: Sync + 'data, const N: usize> IntoParallelIterator for &'data [T; N] {
+    type Item = &'data T;
+    type Iter = Iter<'data, T>;
+
+    fn into_par_iter(self) -> Self::Iter {
+        <&[T]>::into_par_iter(self)
```
nikomatsakis, line 23: Is there any reason we might want to do something more "hard-coded" the array length? I guess not, since we're going to be doing all the stealing and heuristics.

nikomatsakis (approved):
This looks great to me. r=me.

src/array.rs
```diff
@@ -0,0 +1,90 @@
...
+    type Item = &'data mut T;
+    type Iter = IterMut<'data, T>;
+
+    fn into_par_iter(self) -> Self::Iter {
+        <&mut [T]>::into_par_iter(self)
+    }
+}
+
+/// This implementation requires const generics, stabilized in Rust 1.51.
+impl<T: Send, const N: usize> IntoParallelIterator for [T; N] {
```
cuviper, line 38: This is possibly controversial, though technically a _minor_ breaking change. If someone write `array.into_par_iter()`, they would currently get a slice iterator, same as `array.par_iter()`. This PR will change that to a new iterator by value.

It's the same problem blocking the standard library from adding `IntoIterator` for arrays (rust-lang/rust#65819), but I hope the impact is more manageable in our case, at least less likely to come up.

nikomatsakis: Ah, I missed that. Annoying.

nikomatsakis: I agree rather less likely to come up.

src/array.rs
cuviper: Yeah, once you get into runtime-dynamic splitting, we can't really use the constant length anymore.

cuviper:
bors r=nikomatsakis

PR #841, opened 2021-03-29
Add a mean to get the current threadpool

lu-zero:
It is useful when you want to share the same threadpool across
threads outside the threadpool.


rayon-core/src/thread_pool/mod.rs
```diff
@@ -252,6 +252,12 @@ impl ThreadPool {
         // We assert that `self.registry` has not terminated.
         unsafe { spawn::spawn_fifo_in(op, &self.registry) }
     }
+
+    /// Returns the current threadpool.
+    pub fn current() -> ThreadPool {
+        let registry = crate::registry::Registry::current();
+        ThreadPool { registry }
```
cuviper, line 259: We need to at least call `registry.increment_terminate_count()` here, otherwise `Drop` will signal threads to exit.

cuviper (changes requested):
I hesitate to expose `ThreadPool` like this because it leaves us with no clear owner of the pool. The `Registry` is shared because each thread in that pool also holds a handle, but it would be nice if `ThreadPool` remains a bit stronger, as the `Drop` semantics imply by calling `registry.terminate()`. There have also been requests to support a synchronized shutdown, which should naturally consume `self` by value, but that semantic doesn't work as well if `ThreadPool` is a shared handle.

Can you say more about what you want to do with this? Most of the free functions already work naturally with the current pool, whether that's global or something custom, so I wonder why you actually need the `ThreadPool` itself.

rayon-core/src/thread_pool/mod.rs
lu-zero: Sure, thank you for spotting it!

lu-zero:
So far I used it to `println!()` the threadpool to make sure which is which.

nikomatsakis:
Maybe we can return some kind of opaque id

cuviper:
What would you be able to do with an opaque id?

It seems to me what's wanted is just to run things on the pool, not share ownership, so maybe we need some kind of `ThreadPoolRef` mirroring `ThreadPool`'s `&self` methods? But then there are still ownership questions -- does this act like `Weak` and fail/panic if the real `ThreadPool` owner has dropped? (since the threads will then be exiting...)

nikomatsakis:
@cuviper print it out so you can make sure which threadpool is which

lu-zero:
It is also nice so I do not have the annoying pattern of

```
match maybe_threadpool {
  Some(pool) => pool.spawn(run),
  None => rayon::spawn(run),
}
```

I currently have in `rav1e`.

nikomatsakis:
Doesn't `rayon::spawn` use the current threadpool, if any? I forget. Maybe we can make a function for *that*?

lu-zero:
> Doesn't `rayon::spawn` use the current threadpool, if any? I forget. Maybe we can make a function for _that_?

My usecase is to not have to special case between a custom pool and the global pool. Some users would enjoy setting the threadpool according to certain constraint, others would just use the global threadpool they already set for their needs or share a threadpool otherwise created.

lu-zero:
Shall we try to make progress on this?

PR #849, opened 2021-04-21
Make sure `ParBridge` threads consume all items after `done`

### LINKED ISSUE #848: ParBridge using too few threads concurrently
I have a long running single-threaded computation, of which I'd like to run many concurrent instances:

```rust
    let n = 320;
    let iter = repeat(()).take(n).par_bridge();

    let r: u32 = iter.map(|_| {
        // Some long running computation
        sleep(Duration::from_secs(10));
        1
    }).sum();
```

When I run this on many cores (96) often only a random number of cores are used to perform the computation, and the rest is idle.
The number of active cores does not change, over time.
However, it doesn't appear to be a deadlock: rayon works through the iterator in chunks of some random size,
and it terminates correctly.

I have written an [MWE here](https://github.com/ambiso/rayon_bug_mwe/blob/main/src/main.rs).

The MWE is best run with `RAYON_NUM_THREADS=96 cargo run --release`, which reproduces about 75% of the time on my 8 logical core system.

When I increase the item count `n` in the iterator to `threads * threads * 2` the bug does not occur anymore.
However it does still occur with `n = threads * threads * 2 - 1`.

Therefore, I think some logic related to [this line](https://github.com/rayon-rs/rayon/blob/master/src/iter/par_bridge.rs#L169) may be responsible.

Is this expected behavior? Is it a bug?

I will look into it further, I just wanted to get a report out before I forget about it.

Kind regards,
ambiso

cuviper:
We have the `done` flag to indicate when the bridged iterator is out of
items, but there may still be items left in the parallel deque as well!
Threads should check both, in that order, otherwise they may drop out of
the bridge party while there's still work to do.

Fixes #848.

### DIFF
--- a/src/iter/par_bridge.rs
+++ b/src/iter/par_bridge.rs
@@ -125,7 +125,9 @@ where
         let mut count = self.split_count.load(Ordering::SeqCst);
 
         loop {
-            let done = self.done.load(Ordering::SeqCst);
+            // Check if the iterator is exhausted *and* we've consumed every item from it.
+            let done = self.done.load(Ordering::SeqCst) && self.items.is_empty();
+
             match count.checked_sub(1) {
                 Some(new_count) if !done => {
                     match self.split_count.compare_exchange_weak(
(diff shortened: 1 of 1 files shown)


cuviper:
cc @ambiso -- this corrects your [MWE](https://github.com/ambiso/rayon_bug_mwe/) from #848 on my system, but I'd appreciate if you could test it yourself, and also with your original code.

ambiso:
I've tested the MWE and cannot reproduce the bug anymore! :tada: 
My original code is structurally identical, so it shouldn't make a difference.

nikomatsakis (approved):

PR #852, opened 2021-04-28
Use `&mut [MaybeUninit<T>]` in `iter::collect`

### LINKED ISSUE #851: UB in Collect:: reserve_get_tail_slice
It appears that [Collect::reserve_get_tail_slice](https://github.com/rayon-rs/rayon/blob/master/src/iter/collect/mod.rs#L139) might have undefined behavior. The code is reserving uninitialized space, then creating a mutable reference to it. This us UB according to this [documentation](https://doc.rust-lang.org/stable/reference/behavior-considered-undefined.html).

cuviper:
We create a mutable slice to the uninitialized tail when collecting into
a `Vec<T>`, but the Rust Reference says it's UB to produce "a reference
or `Box<T>` that is dangling, unaligned, or points to an invalid value."
[UCG#77] hasn't resolved the validity of references to uninitialized
data yet, but we can avoid it by using `MaybeUninit<T>`.

[UCG#77]: https://github.com/rust-lang/unsafe-code-guidelines/issues/77

Fixes #851.

### DIFF
--- a/src/iter/collect/consumer.rs
+++ b/src/iter/collect/consumer.rs
@@ -27,6 +27,8 @@ pub(super) struct CollectResult<'c, T> {
     target: &'c mut [MaybeUninit<T>],
     /// The current initialized length in `target`
     len: usize,
+    /// Lifetime invariance guarantees that the data flows from consumer to result,
+    /// especially for the `scope_fn` callback in `Collect::with_consumer`.
     invariant_lifetime: PhantomData<&'c mut &'c mut [T]>,
 }
 
--- a/src/iter/collect/consumer.rs
+++ b/src/iter/collect/consumer.rs
@@ -1,11 +1,12 @@
 use super::super::plumbing::*;
 use std::marker::PhantomData;
+use std::mem::MaybeUninit;
 use std::ptr;
 use std::slice;
 
 pub(super) struct CollectConsumer<'c, T: Send> {
     /// A slice covering the target memory, not yet initialized!
-    target: &'c mut [T],
+    target: &'c mut [MaybeUninit<T>],
 }
 
 pub(super) struct CollectFolder<'c, T: Send> {
@@ -20,7 +21,7 @@ pub(super) struct CollectFolder<'c, T: Send> {
 impl<'c, T: Send + 'c> CollectConsumer<'c, T> {
     /// The target memory is considered uninitialized, and will be
     /// overwritten without reading or dropping existing values.
-    pub(super) fn new(target: &'c mut [T]) -> Self {
+    pub(super) fn new(target: &'c mut [MaybeUninit<T>]) -> Self {
         CollectConsumer { target }
     }
 }
@@ -86,7 +87,8 @@ impl<'c, T: Send + 'c> Consumer<T> for CollectConsumer<'c, T> {
         CollectFolder {
             final_len: self.target.len(),
             result: CollectResult {
-                start: self.target.as_mut_ptr(),
+                // TODO: use `MaybeUninit::slice_as_mut_ptr`
+                start: self.target.as_mut_ptr() as *mut T,
                 len: 0,
                 invariant_lifetime: PhantomData,
             },
--- a/src/iter/collect/consumer.rs
+++ b/src/iter/collect/consumer.rs
@@ -9,15 +9,6 @@ pub(super) struct CollectConsumer<'c, T: Send> {
     target: &'c mut [MaybeUninit<T>],
 }
 
-pub(super) struct CollectFolder<'c, T: Send> {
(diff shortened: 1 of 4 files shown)


nikomatsakis (approved):
I like this. The code is clearer now, I think.

src/iter/collect/consumer.rs
```diff
@@ -86,7 +87,8 @@ impl<'c, T: Send + 'c> Consumer<T> for CollectConsumer<'c, T> {
         CollectFolder {
             final_len: self.target.len(),
             result: CollectResult {
-                start: self.target.as_mut_ptr(),
+                // TODO: use `MaybeUninit::slice_as_mut_ptr`
+                start: self.target.as_mut_ptr() as *mut T,
```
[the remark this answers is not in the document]
cuviper: I think there's not much value to that when it's a raw pointer, and we would also have to cast it when `CollectResult::drop` wants to drop a partial slice in place.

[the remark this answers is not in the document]
nikomatsakis: I do see some value, in that it clearly signals that the data may not be initialized, but I agree it's not needed for UB purposes. I think I'd rather see the case to `T` at the point where we are asserting that the data in that range *has been* initialized.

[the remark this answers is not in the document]
cuviper: Well, I'm trying it now. Writing the value is also annoying, as we either have to case the pointer to `*mut T` there, or deref it just to call `MaybeUninit::as_mut_ptr`.

Another idea is to just propagate the real slice here, then I think the only weirdness will be in the merging step.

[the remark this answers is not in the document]
nikomatsakis: this is ok too, maybe just add a comment on the field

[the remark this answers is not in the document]
cuviper: I pushed the slice idea -- let me know what you think.

src/iter/collect/consumer.rs
```diff
@@ -31,7 +23,9 @@ impl<'c, T: Send + 'c> CollectConsumer<'c, T> {
 /// the elements will be dropped, unless its ownership is released before then.
 #[must_use]
 pub(super) struct CollectResult<'c, T> {
-    start: *mut T,
+    /// A slice covering the target memory, initialized up to our separate `len`.
+    target: &'c mut [MaybeUninit<T>],
+    /// The current initialized length in `target`
     len: usize,
     invariant_lifetime: PhantomData<&'c mut &'c mut [T]>,
```
erickt, line 30: you should be able to get rid of this.

src/iter/collect/mod.rs
```diff
@@ -129,14 +130,16 @@ impl<'c, T: Send + 'c> Collect<'c, T> {
     /// and return a slice to the uninitialized tail of the vector
     ///
     /// Safety: The tail slice is uninitialized
-    unsafe fn reserve_get_tail_slice(vec: &mut Vec<T>, len: usize) -> &mut [T] {
+    unsafe fn reserve_get_tail_slice(vec: &mut Vec<T>, len: usize) -> &mut [MaybeUninit<T>] {
```
erickt, line 133: I don't think we need to mark this function as `unsafe` anymore. This has the added benefit of shrinking the `unsafe` block in `Collect::with_consumer` down to setting the final vector length when the collect completes.

src/iter/collect/consumer.rs
cuviper: I *think* you're right, but I'm not certain why it wanted to be invariant in `'c` in the first place. Simple `&'c mut [T]` would have already been invariant in `T`. @bluss, you wrote that in #759, do you recall why?

src/iter/collect/mod.rs
cuviper: Makes sense, done.

src/iter/collect/mod.rs
```diff
@@ -88,55 +89,56 @@ impl<'c, T: Send + 'c> Collect<'c, T> {
     where
         F: FnOnce(CollectConsumer<'_, T>) -> CollectResult<'_, T>,
```
bluss, line 90: I think the closure signature here is why the lifetime parameter wants to be invariant. Since it's internal apis, it's our choice if we want to keep that detail. The invariance just means that you can't play tricks and substitute some in other return value.

cuviper: Thanks, that is a nice guarantee, and zero-cost since it's a function of borrowck. I've added a comment so we can remember this reason later. :slightly_smiling_face:

nikomatsakis (approved):
I like it!

PR #855, opened 2021-05-04
Add `in_place_scope_fifo`

cuviper:
This matches the recent `in_place_scope`, but using `ScopeFifo`. The
scope constructors have also been refactored to avoid duplication.

### DIFF
--- a/rayon-core/src/scope/mod.rs
+++ b/rayon-core/src/scope/mod.rs
@@ -306,10 +306,8 @@ where
     R: Send,
 {
     in_worker(|owner_thread, _| {
-        let scope = Scope {
-            base: ScopeBase::new(owner_thread, owner_thread.registry().clone()),
-        };
-        unsafe { scope.base.complete(Some(owner_thread), || op(&scope)) }
+        let scope = Scope::<'scope>::new(Some(owner_thread), None);
+        scope.base.complete(Some(owner_thread), || op(&scope))
     })
 }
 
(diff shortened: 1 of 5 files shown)


cuviper:
cc @rocallahan

rayon-core/src/scope/mod.rs
```diff
@@ -608,9 +621,7 @@ impl<'scope> ScopeBase<'scope> {
 
     /// Executes `func` as a job, either aborting or executing as
     /// appropriate.
-    ///
-    /// Unsafe because it must be executed on a worker thread.
-    unsafe fn complete<FUNC, R>(&self, owner: Option<&WorkerThread>, func: FUNC) -> R
+    fn complete<FUNC, R>(&self, owner: Option<&WorkerThread>, func: FUNC) -> R
```
nikomatsakis, line 624: Why is this not unsafe? I guess it never had to be...?

rayon-core/src/scope/mod.rs
```diff
@@ -666,30 +673,29 @@ impl<'scope> ScopeBase<'scope> {
...
+    fn new(owner: Option<&WorkerThread>) -> Self {
+        match owner {
+            Some(owner) => ScopeLatch::Stealing {
+                latch: CountLatch::new(),
+                registry: Arc::clone(owner.registry()),
+                worker_index: owner.index(),
+            },
+            None => ScopeLatch::Blocking {
+                latch: CountLockLatch::new(),
+            },
```
nikomatsakis, line 698: +1, nice

rayon-core/src/scope/mod.rs
cuviper: Well, at least the comment "because it must be executed on a worker thread" is optional after #844. I didn't really change that functionality in this PR, just cleaning up.

I suppose there's a correctness issue if `Some(owner)` doesn't match the latch, but we do have `debug_assert_eq` for that, and it's not really a *safety* concern. Worst case is that we can't wake up the right thread. We control all the callers in this same module though, and they do trivially match from `new` to `complete`.

(We can't put the `&WorkerThread` directly in the scope or latch because it's not `Sync`.)

nikomatsakis:
r+ from my side :) @cuviper were you ready to merge?

cuviper:
Sure!

bors r=nikomatsakis

PR #860, opened 2021-05-17
Release rayon 1.5.1 / rayon-core 1.9.1

cuviper:


RELEASES.md
```diff
@@ -1,3 +1,30 @@
+# Release rayon 1.5.1 / rayon-core 1.9.1 (2021-05-21)
```
cuviper, line 1: Note the timebox, aiming for this Friday. We don't need to rush other pending PRs though, as there's no reason we should wait months again for the next release.

nikomatsakis (approved):

PR #862, opened 2021-05-19
Use in_place_scope to spawn skipped items

cuviper:
`Skip` still executes the skipped part of the iterator with a no-op
consumer in case there are side effects, as a point of compatibility
with sequential `Skip`. However, the `Producer` is not necessarily
`'static`, and `ProducerCallback::Output` doesn't require `Send`, so we
didn't have a way to run the no-op in parallel with the normal callback.

Now with `in_place_scope`, we can `spawn` the no-op consumer while we
run the callback directly. The scope will still wait for that no-op to
complete before returning, propagating panics if necessary.

### DIFF
--- a/src/iter/skip.rs
+++ b/src/iter/skip.rs
@@ -79,9 +79,16 @@ where
             where
                 P: Producer<Item = T>,
             {
-                let (before_skip, after_skip) = base.split_at(self.n);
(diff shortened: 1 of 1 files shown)


nikomatsakis (approved):

nikomatsakis:
That's actually really cool =)

PR #869, opened 2021-06-16
impl Pattern for &[char]

cuviper:
The standard library has had this since 1.0, so I don't know why we didn't add it here.


cuviper:
In contrast, I do know why we didn't add string patterns -- because that may depend on order if there's any repetition in the string. It's the same reason that `rsplit("...")` is not the same as `split("...").rev()`, but the difference would be non-deterministic in parallel.

PR #876, opened 2021-08-05
Update dependencies for cargo audit

nyanzebra:
Updates crossbeam and glium for cargo audit purposes. crossbeam had security vulnerability in 0.8.0 so am giving some range availability in which version to pick, ideally this makes updates like this pr needed less frequently

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ exclude = ["ci"]
 
 [dependencies]
 rayon-core = { version = "1.9.1", path = "rayon-core" }
-crossbeam-deque = "0.8.0"
+crossbeam-deque = "0.8"
 
 # This is a public dependency!
 [dependencies.either]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ exclude = ["ci"]
 
 [dependencies]
 rayon-core = { version = "1.9.1", path = "rayon-core" }
-crossbeam-deque = "0.8"
+crossbeam-deque = "0.8.1"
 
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -20,7 +20,7 @@ exclude = ["ci"]
 
 [dependencies]
 rayon-core = { version = "1.9.1", path = "rayon-core" }
-crossbeam-deque = "0.8.0"
+crossbeam-deque = "0.8"
```
cuviper, line 23: In Cargo dependencies, these mean the exact same thing.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html
> The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

> Caret requirements allow SemVer compatible updates to a specified version. An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping.

So the result is that the [CVE](https://github.com/crossbeam-rs/crossbeam/security/advisories/GHSA-pqqp-xmhj-wgcw) fixed version "0.8.1" is perfectly allowed either way. If we want to **require** that fix, we need to say "0.8.1" in our dependency (which will also allow later 0.8.x versions).

nikomatsakis (approved):

nikomatsakis:
This looks good to me, but I'll wait on @cuviper

cuviper:
I just updated `ci/compat-Cargo.lock`, which is our own little oddity, but otherwise it's good, thanks!

bors r+

PR #877, opened 2021-08-08
Update crossbeam-deque

bugadani:
0.8.0 has been yanked


cuviper:
You can update this in your own `Cargo.lock` with semver compatibility, but anyway #876 also updated the explicit dependency. Thanks anyway!

PR #880, opened 2021-08-24
Format better and remove unnecessary borrows

benarmstead:
Title self explanatory

### DIFF
--- a/tests/chars.rs
+++ b/tests/chars.rs
@@ -8,7 +8,7 @@ fn half_open_correctness() {
 
     let range = low..high;
     let mut chars: Vec<char> = range.into_par_iter().collect();
-    chars.sort();
+    chars.sort_unstable();
 
     assert_eq!(
         chars,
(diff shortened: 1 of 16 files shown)


nikomatsakis (approved):
These changes look pretty good to me.

benarmstead:
The CI seems to have brought up one or two errors, I will try and deal with them in the morning.

tests/chars.rs
```diff
@@ -8,7 +8,7 @@ fn half_open_correctness() {
 
     let range = low..high;
     let mut chars: Vec<char> = range.into_par_iter().collect();
-    chars.sort();
+    chars.sort_unstable();
```
cuviper, line 11: Weird, this shouldn't need sorting at all...

tests/chars.rs
```diff
@@ -26,7 +26,7 @@ fn closed_correctness() {
 
     let range = low..=high;
     let mut chars: Vec<char> = range.into_par_iter().collect();
-    chars.sort();
+    chars.sort_unstable();
```
cuviper, line 29: Same here, it should already be sorted.

PR #746, opened 2020-04-08
new scheduler from RFC 5

nikomatsakis:
Implementation of the scheduler described in https://github.com/rayon-rs/rfcs/pull/5 -- modulo the fact that the RFC is mildly out of date. There is a [walkthrough video available](https://youtu.be/HvmQsE5M4cY).

To Do List:

* [x] Fix the cargo lock
* [x] Address use of `AtomicU64`
* [x] Document the handling of rollover and wakeups and convince ourselves it's sound
* [ ] Adopt and document the [proposed scheme for the job event counter](https://github.com/rayon-rs/rayon/pull/746#issuecomment-612029239)
* [ ] Review RFC and list out the places where it differs from the branch

PR #882, opened 2021-09-01
Remove unused `lazy_static` from rayon-core

paolobarbolini:


cuviper:
Thanks! It looks like the last use was removed in commit bb599702cbfdbd03a29962fc1f5cf0d3f3cec42a of #746. We'll still have an indirect dependency through crossbeam, but it's good to clean up our own part.

PR #886, opened 2021-09-16
Add mini-section on Wasm usage

RReverser:
I wasn't sure at first if you'd be interested in linking to the adapter from Rayon docs, but seeing more and more projects using this integration already, figured it might be useful.

Feel free to close if you think otherwise.


nikomatsakis (approved):

nikomatsakis:
bors r+

Seems like a good idea to me

Issue #513, opened 2018-01-24
add `split` operation on indexed parallel iterators


nikomatsakis:
Currently given an `IndexedParallelIterator` there is no way to divide it up other than skip/take which have side-effects. We were discussing the possibility of a `split` that takes an indexed iterator and yields back two of them around a particular index.

cuviper:
Hmm, I started implementing this, but it's awkward for iterators that **own** their data, like `rayon::vec::IntoIter`.  (Whereas its `VecProducer` only holds `&mut [T]`, which can split.)

nikomatsakis:
@cuviper ah, interesting -- very good point =)

The fundamental thing that makes it awkward is that the two returned iterators need to be able to be dropped independently (per this design).

You really would prefer to be iterating over this stuff in place though. Hmm.

cuviper:
`Once` is also awkward to return `(Self, Self)`, unless we allow one of those to not actually iterate once.

`Enumerate` will have to track its start index, which it doesn't now until it's a `Producer`.  And similar to `Once`, would it be weird to have an instance of `Enumerate` floating around that doesn't start at 0?

Or an `Interleave` that starts with an item from the *second* iterator?  Or an `Intersperse` that starts with the interspersed item?

nikomatsakis:
It seems to me that there are enough design questions that it is worth taking this off the 1.0 milestone and iterating separately -- we can always add a `SplittableIndexedIterator` or something.

In any case, I think we are going to want this to be a callback style, like producer, so that the types can vary and not be limited by associated types.

cuviper:
Ah, add a separate trait because we're talking about a required method -- right.

I agree on delaying it.  When we get an RFC process going, this will be a good one to chew on.  I'm also playing with a simpler `trim` like I mentioned in #352, or maybe called `slice` or `range`...

PR #888, opened 2021-10-08
Question: Concatenate list of indexed parallel iterators

adamreichold:
This is admittedly more of a question than a pull request but I thought that it might be easiest to answer this with the code at hand. I have [a situation](https://github.com/adamreichold/rs-ecs/pull/17) in which I basically want to concatenate a
```rust
Vec<I> where I: IndexedParallelIterator
```
into a single indexed parallel iterator with the same semantics as if I would `.into_iter().flatten()` the vector.

I tried to implement this by looking at the existing `Chain` adapter which however handles a fixed number of inner iterators with possibly heterogeneous types instead of a unknown number of iterators with homogeneous types.

I am currently stuck at trying to collect the producers created by the inner iterators into a vector. This does not seem to work as it seems possible that each invocation of
```
<I as IndexParallelIterator>::with_producer
```
will call the given callback with a different producer type.

Leaving aside the overhead, boxing does not seem possible due to `Producer` not being object safe. I did consider defining a simpler `DynProducer` trait that would be, but I think this would end with
```rust
trait IntoIter<T>: Iterator<Item = T> + DoubleEndedIterator + ExactSizeIterator;
type IntoIter = Box<dyn IntoIter<Self::Item>>;
```
which feels prohibitively expensive.

I also considered building a producer out of a list of iterators each recording a list of split positions and only turn this into producers and split those when e.g. `Producer::into_iter` is actually called. But does seem to imply at least sharing (and hence synchronizing) the original list between tasks if a split happens to saddle the boundary between two of original iterators. Also, I am not sure if this would avoid the issue of different producer types eventually.

Does anybody know whether this is possible at all and if so how?

### DIFF
--- a/src/iter/concat.rs
+++ b/src/iter/concat.rs
@@ -1,0 +1,178 @@
+use super::{
+    plumbing::{bridge, Consumer, Producer, ProducerCallback, UnindexedConsumer},
+    IndexedParallelIterator, ParallelIterator,
+};
+
+/// TODO
+pub fn concat<I>(iters: Vec<I>) -> Concat<I>
+where
+    I: IndexedParallelIterator,
+{
+    Concat(iters)
+}
+
+/// TODO
+#[derive(Debug)]
+pub struct Concat<I>(Vec<I>);
+
+impl<I> ParallelIterator for Concat<I>
+where
+    I: IndexedParallelIterator,
+{
+    type Item = I::Item;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+    where
+        C: UnindexedConsumer<Self::Item>,
+    {
+        bridge(self, consumer)
+    }
+
+    fn opt_len(&self) -> Option<usize> {
+        Some(self.len())
+    }
+}
+
+impl<I> IndexedParallelIterator for Concat<I>
+where
+    I: IndexedParallelIterator,
+{
+    fn drive<C>(self, consumer: C) -> C::Result
+    where
+        C: Consumer<Self::Item>,
+    {
+        bridge(self, consumer)
+    }
+
+    fn len(&self) -> usize {
+        self.0.iter().map(IndexedParallelIterator::len).sum()
+    }
+
+    fn with_producer<CB>(self, callback: CB) -> CB::Output
+    where
+        CB: ProducerCallback<Self::Item>,
+    {
+        struct Collector<P1> {
+            len: usize,
+            prods: Vec<(usize, P1)>,
+        }
+
+        impl<'a, T, P1> ProducerCallback<T> for &'a mut Collector<P1>
+        where
+            P1: Producer<Item = T>,
+        {
+            type Output = ();
+
+            fn callback<P2>(self, producer: P2) -> Self::Output
+            where
+                P2: Producer<Item = T>,
+            {
+                // TODO: unify P1 and P2
+                self.prods.push((self.len, producer));
+            }
+        }
+
+        let mut collector = Collector {
+            len: 0,
+            prods: Vec::new(),
+        };
+
+        for iter in self.0 {
+            collector.len += iter.len();
+
+            iter.with_producer(&mut collector);
+        }
+
+        callback.callback(Concat(collector.prods))
+    }
+}
+
+impl<P> Producer for Concat<(usize, P)>
+where
+    P: Producer,
+{
+    type Item = P::Item;
+    type IntoIter = Concat<P::IntoIter>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        Concat(
+            self.0
+                .into_iter()
+                .map(|(_len, prod)| prod.into_iter())
+                .collect(),
+        )
+    }
+
+    fn split_at(self, index: usize) -> (Self, Self) {
+        // TODO: binary search
+        if let Some(pos) = self.0.iter().position(|(len, _prod)| *len > index) {
+            let mut lhs = self.0;
+            let mut rhs = lhs.split_off(pos);
+
+            let lhs_len = lhs.last().map_or(0, |(len, _prod)| *len);
+
+            let (rhs_len, rhs_prod) = rhs.remove(0); // TODO: deque
+
+            let (lhs_prod, rhs_prod) = rhs_prod.split_at(index - lhs_len);
+
+            lhs.push((index, lhs_prod));
+            rhs.insert(0, (rhs_len - index, rhs_prod)); // TODO: deque
+
+            (Concat(lhs), Concat(rhs))
+        } else {
+            (self, Concat(Vec::new()))
+        }
+    }
+}
+
+impl<I> Iterator for Concat<I>
+where
+    I: Iterator + DoubleEndedIterator + ExactSizeIterator,
+{
+    type Item = I::Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            let first = self.0.first_mut()?;
+
+            match first.next() {
+                Some(item) => return Some(item),
+                None => {
+                    self.0.remove(0); // TODO: deque
+                }
+            }
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let len = self.len();
+        (len, Some(len))
+    }
+}
+
+impl<I> DoubleEndedIterator for Concat<I>
+where
+    I: Iterator + DoubleEndedIterator + ExactSizeIterator,
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        loop {
+            let last = self.0.last_mut()?;
+
+            match last.next_back() {
+                Some(item) => return Some(item),
+                None => {
+                    self.0.pop();
+                }
+            }
+        }
+    }
+}
+
+impl<I> ExactSizeIterator for Concat<I>
+where
+    I: Iterator + DoubleEndedIterator + ExactSizeIterator,
+{
+    fn len(&self) -> usize {
+        self.0.iter().map(ExactSizeIterator::len).sum()
+    }
+}
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -106,6 +106,7 @@ mod chain;
 mod chunks;
 mod cloned;
 mod collect;
+mod concat;
 mod copied;
 mod empty;
 mod enumerate;
@@ -155,6 +156,7 @@ pub use self::{
     chain::Chain,
     chunks::Chunks,
     cloned::Cloned,
+    concat::{concat, Concat},
     copied::Copied,
     empty::{empty, Empty},
     enumerate::Enumerate,


adamreichold:
> because any callback style with Vec<I> would still have len() recursion levels,

As presented here, I think there should be no recursion. I am willing to pay collecting into producers into newly allocated vectors though. However, I do not think GAT alone would help with this approach as the lifetime associated with the producer is currently generative, i.e. it is created inside the closure and cannot be named outside to e.g. store this outside of the scope created by the callback as I am trying to do here.

I think this would be possible if the signature were something like
```rust
fn with_producer<'iter, CB, R>(&'iter mut self, callback: CB) -> R where CB: FnOnce(Self::Producer<'iter>) -> R;
```
to connect the producer lifetime with the lifetime of the reference to inner iterator. Making it a mutable reference would allow moving things out of the iterator at the price of having to perform the option dance or resort to `MaybeUninit`.

cuviper:
Well I'm hoping for a normal closure instead, but yes I suppose you could push to a captured `&mut Vec`.

I'm going to experiment with this -- first trying to make it a GAT closure, then perhaps `&mut self`. This is definitely a breaking change in any case, so we're talking about a possible Rayon 2.0 here, but it might be nice!

adamreichold:
I do wonder whether there are exist "insidious" implementations as considered above today which do pass completely unrelated producer types as e.g. optimizations. :thinking:

adamreichold:
> It's also weird that a &mut self method implies it can be called multiple times.

I think this is the same issue as for `Drop`, i.e. https://github.com/rust-lang/rfcs/issues/998

adamreichold:
> which also constrains the lifetime of that state, e.g. nothing self-referential.

It would not be possible to move the producer around otherwise, so that appears consistent to me.

cuviper:
> > It's also weird that a &mut self method implies it can be called multiple times.
> 
> I think this is the same issue as for `Drop`, i.e. [rust-lang/rfcs#998](https://github.com/rust-lang/rfcs/issues/998)

Yeah, sort of, but the compiler forbids explicit calls to `Drop::drop` altogether, so multiple calls are out of the question.

> > which also constrains the lifetime of that state, e.g. nothing self-referential.
> 
> It would not be possible to move the producer around otherwise, so that appears consistent to me.

I meant with that state in the iterator type, not the producer, but still. A concrete example is:
https://github.com/rayon-rs/rayon/blob/2991c044832f4667ec1489804b626a8084700a70/src/collections/binary_heap.rs#L103-L110

> Is the `DoubleEndedIterator` requirement on `Producer::IntoIter` essential?

That's for `rev()` -- see #150 and #224. I think the indirect nature of producers would make conditional constraints fall afoul of rust-lang/rust#20671, even if they weren't used opaquely.

schell:
I am also trying to solve this problem. Conceptually it seems like it _should_ be possible to construct an `IndexedParallelIterator` for a vector of `IndexedParallelIterator`s. Has anyone managed it in another crate? Or is there another way around the problem?

In my specific use case I am trying to zip the elements of range maps (where keys are `usize` and values are stored in contiguous ranges), while providing `None` for the elements that are outside a range. I can construct a `ParallelIterator` using `Flatten`, but then I can't `MultiZip` all of those up.

schell:
I have quite a naive (to me) implementation that uses `Vec::pop` and `rayon::iter::chain` that compiles: 

```
pub struct Concat<I>(Vec<I>);

impl<I: IndexedParallelIterator> ParallelIterator for Concat<I> {
    type Item = I::Item;

    where
        C: rayon::iter::plumbing::UnindexedConsumer<Self::Item> {
        rayon::iter::plumbing::bridge(self, consumer)
    }
}

impl<I: IndexedParallelIterator> IndexedParallelIterator for Concat<I> {
    fn len(&self) -> usize {
        self.0.iter().map(IndexedParallelIterator::len).sum()
    }

    fn drive<C: rayon::iter::plumbing::Consumer<Self::Item>>(mut self, consumer: C) -> C::Result {
        use rayon::iter::plumbing::Reducer;

        let len = self.len();
        if let Some(tail) = self.0.pop() {
            let (left, right, reducer) = consumer.split_at(len - tail.len());
            let (a, b) = rayon::join(|| self.drive(left), || tail.drive(right));
            reducer.reduce(a, b)
        } else {
            rayon::iter::empty().drive(consumer)
        }
    }

    fn with_producer<CB: ProducerCallback<Self::Item>>(mut self, callback: CB) -> CB::Output {
        if let Some(tail) = self.0.pop() {
            self.chain(tail).with_producer(callback)
        } else {
            rayon::iter::empty().with_producer(callback)
        }
    }
}
```

It compiles, but you can't use it as it hits the recursion limit while instantiating the big nest of `Chain`s. Any thoughts?

schell:
Here's another implementation which compiles and runs, but my tests show that it's dropping some elements (summing a 1000x1000 vector of 1's comes out to 736605 instead of 1_000_000) - so I think my splitting is off by one in at least one place - but maybe this is a workable solution? 

```
pub struct Concat<I>(Vec<I>);

impl<I: IndexedParallelIterator> ParallelIterator for Concat<I> {
    type Item = I::Item;

    where
        C: rayon::iter::plumbing::UnindexedConsumer<Self::Item>,
    {
        rayon::iter::plumbing::bridge(self, consumer)
    }
}

impl<I: IndexedParallelIterator> IndexedParallelIterator for Concat<I> {
    fn len(&self) -> usize {
        self.0.iter().map(IndexedParallelIterator::len).sum()
    }

    fn drive<C: rayon::iter::plumbing::Consumer<Self::Item>>(mut self, consumer: C) -> C::Result {
        use rayon::iter::plumbing::Reducer;

        let len = self.len();
        if let Some(tail) = self.0.pop() {
            let (left, right, reducer) = consumer.split_at(len - tail.len());
            let (a, b) = rayon::join(|| self.drive(left), || tail.drive(right));
            reducer.reduce(a, b)
        } else {
            rayon::iter::empty().drive(consumer)
        }
    }

    fn with_producer<CB: ProducerCallback<Self::Item>>(mut self, callback: CB) -> CB::Output {
        return callback.callback(ConcatProducer(vec![], self.0, vec![]));

        struct ConcatProducer<T: IndexedParallelIterator>(Vec<T::Item>, Vec<T>, Vec<T::Item>);

        impl<T: IndexedParallelIterator> Default for ConcatProducer<T> {
            fn default() -> Self {
                Self(vec![], vec![], vec![])
            }
        }

        impl<T: IndexedParallelIterator> rayon::iter::plumbing::Producer for ConcatProducer<T> {
            type Item = T::Item;

            type IntoIter = std::vec::IntoIter<Self::Item>;

            fn into_iter(self) -> Self::IntoIter {
                let mut vs: Vec<Self::Item> = self.0;
                vs.extend(self.1.into_par_iter().flatten().collect::<Vec<_>>());
                vs.extend(self.2);
                vs.into_iter()
            }

            fn split_at(self, index: usize) -> (Self, Self) {
                let (_, left, right) = self.1.into_iter().fold(
                    (0, ConcatProducer::default(), ConcatProducer::default()),
                    |(mut n, mut left, mut right), vs| {
                        let iter_len = vs.len();
                        if n + iter_len <= index {
                            // we have not yet accumulated enough for the split,
                            // everything goes in left
                            left.1.push(vs);
                        } else if n >= index {
                            // we already meet our split quota, everything goes in right
                            right.1.push(vs);
                        } else {
                            // we have to split this vec
                            let mut vs = vs.into_par_iter().collect::<Vec<_>>();
                            let vs_left = vs.splice(0..=(index - n), std::iter::empty());
                            left.2.extend(vs_left);
                            right.0.extend(vs);
                        }
                        n += iter_len;
                        (n, left, right)
                    },
                );
                (left, right)
            }
        }
    }
}
```

schell:
I realize now that impl is pretty broken, I didn't take into account the lefts and rights. I have another impl which passes tests, but I'm fairly certain it's pretty whack. I'm struggling a bit to understand the relationship between `with_producer` and `drive`: 

```
impl<I: IndexedParallelIterator> IndexedParallelIterator for Concat<I> {
    fn len(&self) -> usize {
        self.0.iter().map(IndexedParallelIterator::len).sum()
    }

    fn drive<C: rayon::iter::plumbing::Consumer<Self::Item>>(mut self, consumer: C) -> C::Result {
        use rayon::iter::plumbing::Reducer;
        println!("drive self.len(): {}", self.len());
        let len = self.len();
        if let Some(tail) = self.0.pop() {
            println!("  tail.len: {}", tail.len());
            println!("  len - tail.len: {}", len - tail.len());
            let (left, right, reducer) = consumer.split_at(len - tail.len());
            let (a, b) = rayon::join(|| self.drive(left), || tail.drive(right));
            reducer.reduce(a, b)
        } else {
            rayon::iter::empty().drive(consumer)
        }
    }

    fn with_producer<CB: ProducerCallback<Self::Item>>(mut self, callback: CB) -> CB::Output {
        return callback.callback(ConcatProducer(vec![], self.0, vec![]));

        struct ConcatProducer<T: IndexedParallelIterator>(Vec<T::Item>, Vec<T>, Vec<T::Item>);

        impl<T: IndexedParallelIterator> Default for ConcatProducer<T> {
            fn default() -> Self {
                Self(vec![], vec![], vec![])
            }
        }

        impl<T: IndexedParallelIterator> ConcatProducer<T> {
            fn total_len(&self) -> usize {
                self.0.len()
                    + self
                        .1
                        .iter()
                        .map(IndexedParallelIterator::len)
                        .sum::<usize>()
                    + self.2.len()
            }
        }

        impl<T: IndexedParallelIterator> rayon::iter::plumbing::Producer for ConcatProducer<T> {
            type Item = T::Item;

            type IntoIter = std::vec::IntoIter<Self::Item>;

            fn into_iter(self) -> Self::IntoIter {
                let mut vs: Vec<Self::Item> = self.0;
                vs.extend(self.1.into_par_iter().flatten().collect::<Vec<_>>());
                vs.extend(self.2);
                vs.into_iter()
            }

            fn split_at(mut self, index: usize) -> (Self, Self) {
                let mut lines = vec![];
                lines.push(format!(
                    "split_at: index:{} left:{}  center:{},{}  right:{}",
                    index,
                    self.0.len(),
                    self.1.len(),
                    self.1
                        .iter()
                        .map(IndexedParallelIterator::len)
                        .sum::<usize>(),
                    self.2.len()
                ));
                let total_len = self.total_len();
                let mut left = ConcatProducer::<T>::default();
                let taken_from_left = self
                    .0
                    .splice(0..index.min(self.0.len()), std::iter::empty())
                    .collect::<Vec<_>>();
                left.0 = taken_from_left;
                if !left.0.is_empty() {
                    lines.push(format!("  took {} from left", left.0.len()));
                }

                let (mut left, mut right) = std::mem::take(&mut self.1).into_iter().fold(
                    (left, self),
                    |(mut left, mut right), vs| {
                        let left_len = left.total_len();
                        let iter_len = vs.len();
                        if left_len + iter_len < index {
                            lines.push(format!("    {} + {} < {} - going left", left_len, iter_len, index));
                            // we have not yet accumulated enough for the split,
                            // everything goes in left
                            left.1.push(vs);
                        } else if left_len >= index {
                            lines.push(format!("    {} >= {} - going right", left_len, index));
                            // we already meet our split quota, everything goes in right
                            right.1.push(vs);
                        } else {
                            lines.push(format!(
                                "    index:{} splitting with splice 0..{}",
                                index,
                                index - left_len
                            ));
                            // we have to split this vec
                            let mut vs = vs.into_par_iter().collect::<Vec<_>>();
                            let vs_left = vs.splice(0..(index - left_len), std::iter::empty());
                            left.2.extend(vs_left);
                            right.0.extend(vs);
                        }
                        (left, right)
                    },
                );

                let left_len = left.total_len();
                if left_len < index {
                    left.2.extend(right.2.splice(0..(index - left_len), std::iter::empty()));
                }

                std::thread::sleep_ms(10);
                println!("{}", lines.join("\n"));
                assert_eq!(left.total_len(), index, "{} /= {}", left.total_len(), index);
                assert_eq!(
                    left.total_len() + right.total_len(),
                    total_len,
                    "{} + {} /= {}",
                    left.total_len(),
                    right.total_len(),
                    total_len
                );
                (left, right)
            }
        }
    }
}
```

schell:
Doing this would be a lot easier with #513, because we could easily split the iterator that's on the index.

PR #897, opened 2021-10-27
Use the wording 'rayon job' for with_min/max_len

bluss:
Because it depends on scheduling and job size how many jobs are
processed per thread, it seems more consistent to say rayon job here.


cuviper:
Makes sense, thanks!

bors r+

PR #906, opened 2021-12-31
Make dump fn non-generic

kornelski:
Otherwise two versions of the function are needlessly generated.

### DIFF
--- a/rayon-core/src/log.rs
+++ b/rayon-core/src/log.rs
@@ -377,7 +377,7 @@ impl SimulatorState {
         }
     }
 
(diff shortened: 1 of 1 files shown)


cuviper:
How much code size is actually saved here? Trimming two functions down to one seems like an easy win, but it also means we need a full vtable of `Write` functions for each caller type, no? Those may be individually small enough, but I want to check.

kornelski:
cargo bloat shows 2x 3.8KB for `dump`.

kornelski:
I see. I assumed that vtable for Write would be tiny. Thanks for checking this.

PR #907, opened 2022-01-04
Address SB/provenance issues found by cargo miri test

saethlin:
I've submitted a few PRs to other libraries to address errors from running `cargo miri test` on them. Not all of the code that Miri rejects as not conforming to Stacked Borrows is necessarily invalid in terms of current rustc+LLVM semantics, but some of it is a little dubious on its own, and also I think it would be really nice for core libraries to be clean against miri so that we can use miri to understand how well people follow the prototype rules without Miri being halted by non-conforming code in dependencies.

To that end, I'm here because I was running `cargo miri test --features=rayon` in the `hashbrown` crate, and I got an error that I could trace back to this crate. Then I started running `cargo miri test` with `MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-tag-raw-pointers"` on this crate's tests, and I found a few things. This PR fixes everything I found. My motivation is a little bit of column A (some of this code is like... maybe let's not do it that way, but also not definitely broken) and a little bit of column B (I got here via another crate's tests).

In two cases, this code was going through the `Deref` to a slice impl on `Vec` in order to get a pointer into the uninitialized region between len and capacity. As far as I understand, under the Stacked Borrows model, references narrow provenance, thus it is not valid to use such a pointer outside the initialized region.

The collecting code has a related problem, where it is trying to merge slices that were originally derived from the same array by rebuilding a larger slice by using a pointer based on one of the subslices. To construct a valid merged slice we need access to the original provenance of the whole array. It is possible to resolve this by keeping a `&mut[]` as well as a `*mut`, using the raw pointer to construct the merged slice because it has sufficent provenance. But I think that ends up juggling a few too many pieces of data, so I implemented a fix by replacing the slice with a pointer and length. I also tried to add a bit more description to some names to help out in reading the more arcane implementation.

### DIFF
--- a/src/iter/collect/consumer.rs
+++ b/src/iter/collect/consumer.rs
@@ -4,16 +4,34 @@ use std::mem::MaybeUninit;
 use std::ptr;
 use std::slice;
 
+/// We need to store raw pointers in a `Send` struct to remember
+/// provenance (see `CollectResult`).
+#[derive(Clone, Copy)]
+struct SendPtr<T>(*mut T);
+
+// SAFETY: !Send for raw pointers is not for safety, just as a lint
+unsafe impl<T> Send for SendPtr<T> {}
+
 pub(super) struct CollectConsumer<'c, T: Send> {
-    /// A slice covering the target memory, not yet initialized!
-    target: &'c mut [MaybeUninit<T>],
+    /// See `CollectConsumer` for explanation of why this is not a slice
+    start: SendPtr<MaybeUninit<T>>,
+    len: usize,
+    marker: PhantomData<&'c mut T>,
 }
 
 impl<'c, T: Send + 'c> CollectConsumer<'c, T> {
     /// The target memory is considered uninitialized, and will be
     /// overwritten without reading or dropping existing values.
-    pub(super) fn new(target: &'c mut [MaybeUninit<T>]) -> Self {
-        CollectConsumer { target }
+    pub(super) fn new(
+        start: *mut MaybeUninit<T>,
+        len: usize,
+        marker: PhantomData<&'c mut T>,
+    ) -> Self {
+        CollectConsumer {
+            start: SendPtr(start),
+            len,
+            marker,
+        }
     }
 }
 
@@ -23,10 +41,14 @@ impl<'c, T: Send + 'c> CollectConsumer<'c, T> {
 /// the elements will be dropped, unless its ownership is released before then.
 #[must_use]
 pub(super) struct CollectResult<'c, T> {
-    /// A slice covering the target memory, initialized up to our separate `len`.
-    target: &'c mut [MaybeUninit<T>],
-    /// The current initialized length in `target`
-    len: usize,
+    /// This pointer and length has the same representation as a slice,
+    /// but retains the provenance of the entire array so that we can merge
+    /// these regions together in `CollectReducer`.
+    /// Constructing a slice from this start + start_l
+    start: SendPtr<MaybeUninit<T>>,
+    total_len: usize,
+    /// The current initialized length after `start`
+    initialized_len: usize,
     /// Lifetime invariance guarantees that the data flows from consumer to result,
     /// especially for the `scope_fn` callback in `Collect::with_consumer`.
     invariant_lifetime: PhantomData<&'c mut &'c mut [T]>,
@@ -37,25 +59,26 @@ unsafe impl<'c, T> Send for CollectResult<'c, T> where T: Send {}
 impl<'c, T> CollectResult<'c, T> {
     /// The current length of the collect result
     pub(super) fn len(&self) -> usize {
-        self.len
+        self.initialized_len
     }
 
     /// Release ownership of the slice of elements, and return the length
     pub(super) fn release_ownership(mut self) -> usize {
-        let ret = self.len;
-        self.len = 0;
+        let ret = self.initialized_len;
+        self.initialized_len = 0;
         ret
     }
 }
 
 impl<'c, T> Drop for CollectResult<'c, T> {
     fn drop(&mut self) {
-        // Drop the first `self.len` elements, which have been recorded
+        // Drop the first `self.initialized_len` elements, which have been recorded
         // to be initialized by the folder.
         unsafe {
-            // TODO: use `MaybeUninit::slice_as_mut_ptr`
-            let start = self.target.as_mut_ptr() as *mut T;
-            ptr::drop_in_place(slice::from_raw_parts_mut(start, self.len));
+            ptr::drop_in_place(slice::from_raw_parts_mut(
+                self.start.0 as *mut T,
+                self.initialized_len,
+            ));
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/iter/collect/consumer.rs
```diff
@@ -4,16 +4,34 @@ use std::mem::MaybeUninit;
 use std::ptr;
 use std::slice;
 
+/// We need to store raw pointers in a `Send` struct to remember
+/// provenance (see `CollectResult`).
+#[derive(Clone, Copy)]
+struct SendPtr<T>(*mut T);
+
+// SAFETY: !Send for raw pointers is not for safety, just as a lint
```
adamreichold, line 12: Is this really correct? Raw pointers could point into thread-local data or short-lived stack data, couldn't they?

src/vec.rs
```diff
@@ -143,9 +143,8 @@ impl<'data, T: Send> IndexedParallelIterator for Drain<'data, T> {
 
             // Create the producer as the exclusive "owner" of the slice.
             let producer = {
-                // Get a correct borrow lifetime, then extend it to the original length.
```
adamreichold, line 146: This seems to have been written intentionally so that the lifetime of slice would be constrained by the function body whereas now `slice` could produce any lifetime required by `DrainProducer::new`.  Is it possible that inference produces an invalid lifetime here as the removed comment seemed to suggest?

src/iter/collect/consumer.rs
saethlin: They sure can. But you can make any raw pointer Send by wrapping it in AtomicPtr using 100%-safe code. The safety comment I wrote is based on: https://doc.rust-lang.org/nomicon/send-and-sync.html?highlight=Lint#send-and-sync

src/vec.rs
saethlin: It is _more likely_ that the new code produce a wrong lifetime when subjected to future changes, so I've pinned down the lifetime to be sure.

src/vec.rs
```diff
@@ -143,9 +143,8 @@ impl<'data, T: Send> IndexedParallelIterator for Drain<'data, T> {
...
-                let mut slice = &mut self.vec[start..];
-                slice = slice::from_raw_parts_mut(slice.as_mut_ptr(), self.range.len());
+                let ptr = self.vec.as_mut_ptr().add(start);
+                let slice: &'data mut [T] = slice::from_raw_parts_mut(ptr, self.range.len());
```
cuviper, line 147: Doesn't this slice alias `Drain`'s `&'data mut Vec<T>` that will be used in `Drop`? The idea before was to have a properly re-borrowed (stacked?) lifetime just for the more local scope. I'm not sure whether pointers represent that nesting when you create a new slice.

Another way to do this is with a function that ties the input and output lifetimes together. (elided or explicit)

saethlin:
By the way, related to @adamreichold's comment above, this code already exists in `rayon` to evade the `!Send` impl on raw pointers:

https://github.com/rayon-rs/rayon/blob/f45eee8fa49c1646a00f084ca78d362f381f1b65/src/slice/mergesort.rs#L500-L502
https://github.com/rayon-rs/rayon/blob/f45eee8fa49c1646a00f084ca78d362f381f1b65/src/slice/mergesort.rs#L601-L603

Is this the preferred mechanism? I think it would be best to be consistent across the codebase, and if possible I would selfishly prefer to use a `SendPtr` type as above, since at the moment miri just doesn't know what to do with the pointers that you get back out of an integer. Miri just throws up its hands and tosses an error (below). One could also argue that a `SendPtr` type communicates intent better than passing a `usize` that is secretly a pointer.
```
error: Undefined Behavior: no item granting write access to tag <untagged> at alloc5517403 found in borrow stack.
    --> /home/ben/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/intrinsics.rs:2083:14
     |
2083 |     unsafe { copy_nonoverlapping(src, dst, count) }
     |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no item granting write access to tag <untagged> at alloc5517403 found in borrow stack.
     |
     = help: this indicates a potential bug in the program: it performed an invalid operation, but the rules it violated are still experimental
     = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
             
     = note: inside `std::intrinsics::copy_nonoverlapping::<i32>` at /home/ben/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/intrinsics.rs:2083:14
     = note: inside `rayon::slice::mergesort::merge::<i32, &for<'r, 's> fn(&'r i32, &'s i32) -> bool {<i32 as std::cmp::PartialOrd>::lt}>` at /home/ben/rayon/src/slice/mergesort.rs:160:9
     = note: inside `rayon::slice::mergesort::mergesort::<i32, for<'r, 's> fn(&'r i32, &'s i32) -> bool {<i32 as std::cmp::PartialOrd>::lt}>` at /home/ben/rayon/src/slice/mergesort.rs:366:13
```

src/iter/collect/consumer.rs
cuviper: I think that's ok, but we probably want `T: Send` here for it to be safe to dereference on another thread. The rest of the code already has that, but I think it would be cleaner to have it here too.

It could also use `NonNull` to get the niche, but we don't want covariance, so it would take another `PhantomData` to counteract that.

src/iter/collect/consumer.rs
```diff
@@ -4,16 +4,34 @@ use std::mem::MaybeUninit;
...
+unsafe impl<T> Send for SendPtr<T> {}
+
 pub(super) struct CollectConsumer<'c, T: Send> {
-    /// A slice covering the target memory, not yet initialized!
-    target: &'c mut [MaybeUninit<T>],
+    /// See `CollectConsumer` for explanation of why this is not a slice
+    start: SendPtr<MaybeUninit<T>>,
```
cuviper, line 17: With raw pointers, I don't think we need `MaybeUninit` anymore, do we?

cuviper:
> I think it would be best to be consistent across the codebase, and if possible I would selfishly prefer to use a `SendPtr` type as above, since at the moment miri just doesn't know what to do with the pointers that you get back out of an integer. Miri just throws up its hands and tosses an error (below). One could also argue that a `SendPtr` type communicates intent better than passing a `usize` that is secretly a pointer.

I agree, that sounds like a good idea.

src/iter/collect/mod.rs
```diff
@@ -90,7 +90,11 @@ impl<'c, T: Send + 'c> Collect<'c, T> {
         F: FnOnce(CollectConsumer<'_, T>) -> CollectResult<'_, T>,
     {
         let slice = Self::reserve_get_tail_slice(&mut self.vec, self.len);
-        let result = scope_fn(CollectConsumer::new(slice));
+        let result = scope_fn(CollectConsumer::new(
+            slice.as_mut_ptr().cast(),
+            slice.len(),
+            std::marker::PhantomData,
```
cuviper, line 96: This marker is meant to tie the lifetime, but here we're producing it out of thin air, so it could be anything.

src/vec.rs
cuviper: You marked this comment about `'data` aliasing as resolved, but I don't see how it was addressed in any way...

cuviper (commented):
I'm still not comfortable with some of the lifetime hygiene, but I went ahead and made further changes myself. I'll push that in a moment, and if you're able to return a review on those, I would appreciate it.

src/vec.rs
saethlin: Same. It's been a while, I don't know why I did that.

I'm not entirely sure what you mean by `'data` aliasing. I agree that this code could become dangerous if the lifetime is wrong. But there's no aliasing issue in terms of Stacked Borrows if you get lifetimes wrong. But in any case, I think that a helper function is a good idea here, especially because I'm struggling to come up with a good name for it.

cuviper: In my commit, I moved this to a `DrainProducer::from_vec` to borrow the lifetime from `&mut Vec`.

saethlin: Oh jeez I see this slipped out of my unreads again without me responding. In any case, I like what you've done here and thanks for plowing ahead.

cuviper: No worries, thanks for checking back!

PR #925, opened 2022-04-02
Added A `par_chunk_by` method

jakeKonrad:
Hello, I found this method useful in my code and thought it could be used here. However it relies on #![feature(slice_group_by)] unstable feature of the rust std library and you would probably want to wait for that method to be stabilized anyways so as to stay consistent with whatever naming they choose. This pull request is here for when that happens. Also if any one has any comments on the code I'd really appreciate hearing them.

### DIFF
--- a/src/slice/group_by.rs
+++ b/src/slice/group_by.rs
@@ -104,7 +104,7 @@ where
     }
 }
 
-// Mutable 
+// Mutable
 
 struct GroupByMutProducer<'data, 'p, T, P> {
     pred: &'p P,
--- a/src/slice/group_by.rs
+++ b/src/slice/group_by.rs
@@ -7,37 +7,13 @@ where
 {
     let n = xs.len() / 2;
 
-    for (start, end) in (0..).scan(true, |cont, i| {
-        if *cont {
-            let offset = 2 * i;
-            let start = n.saturating_sub(offset);
-            let end = n + offset;
-            Some(
-                if !(1..xs.len()).contains(&start) || !(2..xs.len()).contains(&end) {
-                    *cont = false;
-                    (0, xs.len())
-                } else {
-                    (start, end)
-                },
-            )
-        } else {
-            None
-        }
-    }) {
-        match xs[start..end]
-            .windows(2)
-            .enumerate()
-            .find_map(|(i, win)| {
-                if pred(&win[0], &win[1]) {
-                    None
-                } else {
-                    Some(i)
-                }
-            })
-            .map(|i| start + i)
-        {
-            Some(i) => return Some(i),
-            None => {}
+    for m in (1..((n / 2) + 1)).map(|x| 2 * x) {
+        let start = n.saturating_sub(m);
+        let end = std::cmp::min(n + m, xs.len());
+        for i in start..(end - 1) {
+            if !pred(&xs[i], &xs[i + 1]) {
+                return Some(start + i);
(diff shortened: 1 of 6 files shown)


cuviper:
Thanks, it's an interesting feature! Do you think you'll try a mutable version as well?

And yes, we'll want to wait for the API to stabilize in the standard library so we can match it.

jakeKonrad:
That should be the mutable version.

cuviper:
Just for reference, that `slice_group_by` feature is tracked in rust-lang/rust#80552.

stevenliebregt:
I'm interested in this, but I found a bug in it, the following call:

```rust
let data = vec![1, 1, 2, 2, 2, 2, 2, 3, 3, 3, 4, 4];
let groups: Vec<_> = data.par_group_by(|&a, &b| a == b).collect::<Vec<_>>();

println!("{:?}", groups);
```

will result in 5 groups instead of the expected 4, these groups are:

```
[[1, 1], [2, 2, 2, 2, 2], [3], [3, 3], [4, 4]]
```

jakeKonrad:
I think that fixed it

stevenliebregt:
> I think that fixed it

It did, nice job!

cuviper:
FYI, that stabilization just landed with a rename to `chunk_by` / `chunk_by_mut`:
https://github.com/rust-lang/rust/pull/117678

src/slice/group_by.rs
```diff
@@ -0,0 +1,194 @@
...
+    F: Fn(&T, &T) -> bool,
+{
+    let n = (xs.len() / 2).saturating_sub(1);
+
+    for m in (1..((n / 2) + 1)).map(|x| 2 * x) {
+        let start = n.saturating_sub(m);
+        let end = std::cmp::min(n + m, xs.len());
+        let fsts = &xs[start..end];
+        let (_, snds) = fsts.split_first()?;
+        match fsts.iter().zip(snds).position(|(x , y)| !pred(x, y)) {
```
cuviper, line 15: If I understand correctly, these nested loops are performing a widening search, right? With `n` as the midpoint, it's basically scanning `mid-2..mid+2`, `mid-4..mid+4`, `mid-6..mid+6`, etc.

If so, that's O(n^2) as it will keep re-testing the middle portion that the previous outer loop already tried.

cuviper (commented):
I wonder if this can be refactored to use the internal `SplitProducer` infrastructure, because that's basically what this iterator is doing overall, splitting on a pair-predicate. That API won't work directly as-is, but I've also been working on modifications to add `split_inclusive`, which might be instructive on how to extend it.

src/slice/group_by.rs
jakeKonrad: Yes I believe you're correct. Or at least that's what I was trying to accomplish. It was the most obviously correct way for me. I had an optimisation planned where it doesn't rescan the center but haven't yet worked it out in a way that preserves correctness.

cuviper:
> I wonder if this can be refactored to use the internal `SplitProducer` infrastructure,

I didn't find a good way to shoehorn this into `SplitProducer`, per se, but I've now made the implementation quite similar to that, and a test asserts that we get exactly `len() - 1` comparisons.

cuviper:
Thanks @jakeKonrad! Hope you don't mind that I wrapped this up myself.

PR #933, opened 2022-05-07
Rename autocfg probe method to _foo to avoid dead code warning

danielframpton:
If rayon is built in a configuration with warnings as error (e.g., `RUSTFLAGS=-Dwarnings`), then the probe will fail even though the compiler does have the necessary const generic support.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,7 +3,7 @@ fn main() {
     if ac.probe_expression("(0..10).step_by(2).rev()") {
         autocfg::emit("has_step_by_rev");
     }
-    if ac.probe_expression("{ fn foo<const N: usize>() {} }") {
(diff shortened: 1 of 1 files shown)


cuviper:
This is fine, although I must say I'm generally not a fan of `RUSTFLAGS=-Dwarnings`.
(We'll probably start [applying `--cap-lints`](https://github.com/cuviper/autocfg/issues/41) to avoid such problems...)

bors r+

cuviper:
The failure is not your fault -- I'm looking into what changed on nightly...

cuviper:
https://github.com/rust-lang/rust/issues/88038#issuecomment-1125494273

cuviper:
That should be fixed by #935.

bors retry

PR #937, opened 2022-05-16
Use std:: thread:: available_parallelism() instead of num_cpus dependency

andrewdavidmackenzie:
From rust 1.59 we have a call for available_parallelism() in the std library and we can remove the dependency on num_cpu.

Open questions for maintainers:
- what is the minimum version of rust that you want to support
- what is the best default value when the call to get parallelism fails? I chose "1".

### DIFF
--- a/ci/compat-Cargo.lock
+++ b/ci/compat-Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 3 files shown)


cuviper:
> * what is the minimum version of rust that you want to support

We have an [RFC](https://github.com/rayon-rs/rfcs/blob/master/accepted/rfc0003-minimum-rustc.md) for that, setting policy for at least one year. Right now that would be 1.52, while we're currently still sitting on the last MSRV bump to 1.36.

We could `autocfg` this change in the build script, but we can't make the `num_cpus` dependency conditional that way.

> * what is the best default value when the call to get parallelism fails? I chose "1".

Yeah, that's what `num_cpus` does if it fails internally.

andrewdavidmackenzie:
OK, quite a while to wait then.
Feel free to close the PR, or not, as you see fit.
I had removed num_cpu dependency from my project and then saw it was still there via rayon, so thought I'd take a shot at getting it removed totally.

adamreichold:
I just wanted to add that AFAIK, `std`'s implementation does not consider cgroups (as of Rust 1.59) in contrast to `num_cpus`. So there still is a functional difference between the two options.

This seems especially relevant as Rayon-based programs will often aim to saturate the detected number of hardware threads so I this sounds like it could lead to significant oversubscription if cgroups are involved.

cuviper:
That's true -- cgroup support is coming in 1.61, per rust-lang/rust#92697.

lukehsiao:
Just updating here that it has landed now that 1.61 is released :).

grego:
What about making the `num_cpus` dependency optional?

cuviper:
That cgroup support was only for cgroupv2, and this caused a regression for cgroupv1 users when Cargo switched from `num_cpus`, which they're now reverting. See rust-lang/rust#97549.

> What about making the `num_cpus` dependency optional?

Then we would have one thread if the feature is not enabled? That's not great...
(I mind that less for errors since that's an unlikely runtime condition.)

grego:
I meant the `available_parallelism` feature would be used with `num_cpus` disabled.

cuviper:
Just to note, Rust 1.64.0 added cgroupv1 support in rust-lang/rust#97925.

cuviper:
Going by our policy, 1.64.0 will be a year old tomorrow, so we can merge this as long as I don't publish too soon. :)

However, I think I'll leave the strict MSRV at 1.63 (from #948) anyway, because that keeps us compatible with the system rustc on Debian 12 (bookworm), where AFAICS they use cgroups-v2 by default anyway.

bors r+

cuviper:
(fixed a typo in docs)

bors r+

PR #946, opened 2022-06-15
Replace `docopt` with `clap`

palango:
This PR replaces `docopt` with `clap`.

Something similar was worked on in #737, but now `clap` includes the the derive macros that `structopt` had.

This also makes it possible to remove the `serde` dependency.

### DIFF
--- a/rayon-demo/Cargo.toml
+++ b/rayon-demo/Cargo.toml
@@ -8,17 +8,13 @@ publish = false
 [dependencies]
 rayon = { path = "../" }
 cgmath = "0.18"
-docopt = "1"
+clap = { version = "3.2.3", features = ["derive"] }
 fixedbitset = "0.4"
 glium = "0.31"
 lazy_static = "1"
 rand = "0.8"
 rand_xorshift = "0.3"
 regex = "1"
-
-[dependencies.serde]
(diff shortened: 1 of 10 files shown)


cuviper:
I notice that the options are positioned less flexibly now -- is there a way to relax that?

e.g. `life play --fps 144` fails, while `life --fps 144 play` works, but both worked before.

LoganDark:
IMHO, `clap` may be overkill here. You could probably use something like [`gumdrop`](https://docs.rs/gumdrop) instead, which outperforms every other crate in the ecosystem except for [`getargs`](https://github.com/j-tai/getargs) (which does not have derive).

palango:
> I notice that the options are positioned less flexibly now -- is there a way to relax that?

Not that I'm aware of.


Do you think that this is worth changing, or should I just close the PR?

cuviper:
Like I said in #737, I don't really have an opinion on which crate we use -- especially now that I've moved that example program under `rayon-demo`, so this choice doesn't affect anything we actually publish.

The code does look cleaner here, so that's nice, but the option order inflexibility is a little annoying. Maybe those can be simple positional arguments instead of sub-commands? But I don't know if that will show up well in `--help`.

PR #964, opened 2022-07-18
Upgrade to 2021 edition, MSRV 1.56.0

cuviper:


cuviper:
Note that our MSRV policy is for a year of compatibility -- Rust 1.56.0 came out on October 21, 2021, so we have a few months before this should land.

Issue #773, opened 2020-06-20
Feature request: values.par_iter().fold_groups(group_size: usize, |left: T, right: T| -> T)


bergkvist:
## Problem:
- I have around 7000 input files (xml), and 25 output files (csv). 
- Every input file might write to all 25 output files.
- I don't have enough memory to fit all 7000 files in memory at the same time.
 
## Potential solution:
Performing 7000 * 25 IO append operations becomes my bottleneck. What if some of these writes could be grouped together in memory first?

Basically, given a group size, an associative fold operation, we could combine individual par_iter-items into groups. Assuming a group size of 25, the number ouf IO write operations would be reduced from 7000*25 to 7000.

```
Illustration with a group size of 3. Notice that the last group only combines 2 inputs.

t0---\
t1---->---g0---->
t2---/
t3---\
t4---->---g1---->
t5---/
t6---\
t7---->---g2---->
```

cuviper:
The existing [`chunks`](https://docs.rs/rayon/1.3.1/rayon/iter/trait.IndexedParallelIterator.html#method.chunks) might work for you, but maybe you don't want the `Vec` allocation for each chunk of items.

I guess I could see this as a new method on `IndexedParallelIterator`, folding in a custom way rather than just into a `Vec`. I would give that a signature more like our `fold`:

```rust
fn fold_chunks<T, ID, F>(self, chunk_size: usize, identity: ID, fold_op: F) -> Fold<Self, ID, F>
where
    F: Fn(T, Self::Item) -> T + Sync + Send,
    ID: Fn() -> T + Sync + Send,
    T: Send,
```

A corresponding `fold_with` -> `fold_chunks_with` would make sense too.

The signature you gave, with `|left: T, right: T| -> T`, looks more like what we would call a `reduce`. Maybe we could also support a chunk/group version of `reduce`, producing another parallel iterator instead of the final item.

willcrozi:
I stumbled across this FR after implementing something similar to this a while back, I've brought it up to date here: https://github.com/willcrozi/rayon/tree/fold_chunks

My use case was grouping strings for output, avoiding the extra allocations associated with `chunks`.

I think it might be slightly different to the above: my `fold_chunks` is a parallel iterator that yields the sequential fold of each chunk  and it's implemented it as a `Producer` (I used `chunks` for reference). I guess a chained call to `fold` would make it equivalent?

[...]

PR #970, opened 2022-08-19
add fold_chunks and fold_chunks_with parallel iterators

willcrozi:
This adds two iterators to allow working with fixed-size chunks in a similar way to `chunks()` but without incurring the overhead of allocating a `Vec<T>` for each chunk.

Something similar was discussed in #773.

A call such as:
```
par_iter.fold_chunks(chunk_len, || init(), |acc, x| foo(acc, x))
```
would be functionally equivalent to:
```
par_iter.chunks(chunk_len)
    .map(|chunk|
        chunk.into_iter()
            .fold(init(), |acc, x| foo(acc, x|))
```
A final `fold()` (or `for_each()`, `sum()`, etc) across all mapped chunks can be chained by user code if needed, like this I think it's more flexible.

### DIFF
--- a/src/iter/fold_chunks_with.rs
+++ b/src/iter/fold_chunks_with.rs
@@ -30,8 +30,13 @@ where
     F: Fn(U, I::Item) -> U + Send + Sync,
 {
     /// Creates a new `FoldChunksWith` iterator
-    pub(super) fn new(base: I, chunk_size: usize, item: U, fold_op: F, ) -> Self {
-        FoldChunksWith { base, chunk_size, item, fold_op }
+    pub(super) fn new(base: I, chunk_size: usize, item: U, fold_op: F) -> Self {
+        FoldChunksWith {
+            base,
+            chunk_size,
+            item,
+            fold_op,
+        }
     }
 }
 
@@ -256,9 +261,14 @@ where
 
 #[test]
 fn check_fold_chunks_with() {
-    let words = "bishbashbosh!".chars().collect::<Vec<_>>()
+    let words = "bishbashbosh!"
+        .chars()
+        .collect::<Vec<_>>()
         .into_par_iter()
-        .fold_chunks_with(4, String::new(), |mut s, c| { s.push(c); s } )
+        .fold_chunks_with(4, String::new(), |mut s, c| {
+            s.push(c);
+            s
+        })
         .collect::<Vec<_>>();
 
     assert_eq!(words, vec!["bish", "bash", "bosh", "!"]);
--- a/src/iter/fold_chunks_with.rs
+++ b/src/iter/fold_chunks_with.rs
@@ -1,0 +1,265 @@
+use std::cmp::min;
+
+use super::plumbing::*;
+use super::*;
+use crate::math::div_round_up;
+
+/// `FoldChunksWith` is an iterator that groups elements of an underlying iterator and applies a
+/// function over them, producing a single value for each group.
+///
+/// This struct is created by the [`fold_chunks_with()`] method on [`IndexedParallelIterator`]
+///
+/// [`fold_chunks_with()`]: trait.IndexedParallelIterator.html#method.fold_chunks
+/// [`IndexedParallelIterator`]: trait.IndexedParallelIterator.html
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+#[derive(Debug, Clone)]
+pub struct FoldChunksWith<I, U, F>
+where
+    I: IndexedParallelIterator,
+{
+    base: I,
+    chunk_size: usize,
+    item: U,
+    fold_op: F,
+}
+
+impl<I, U, F> FoldChunksWith<I, U, F>
+where
+    I: IndexedParallelIterator,
+    U: Send + Clone,
+    F: Fn(U, I::Item) -> U + Send + Sync,
+{
+    /// Creates a new `FoldChunksWith` iterator
+    pub(super) fn new(base: I, chunk_size: usize, item: U, fold_op: F, ) -> Self {
+        FoldChunksWith { base, chunk_size, item, fold_op }
+    }
+}
+
+impl<I, U, F> ParallelIterator for FoldChunksWith<I, U, F>
+where
+    I: IndexedParallelIterator,
+    U: Send + Clone,
+    F: Fn(U, I::Item) -> U + Send + Sync,
+{
+    type Item = U;
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+    where
+        C: Consumer<U>,
+    {
+        bridge(self, consumer)
+    }
+
+    fn opt_len(&self) -> Option<usize> {
+        Some(self.len())
+    }
+}
+
+impl<I, U, F> IndexedParallelIterator for FoldChunksWith<I, U, F>
+where
+    I: IndexedParallelIterator,
+    U: Send + Clone,
+    F: Fn(U, I::Item) -> U + Send + Sync,
+{
+    fn len(&self) -> usize {
+        div_round_up(self.base.len(), self.chunk_size)
+    }
+
+    fn drive<C>(self, consumer: C) -> C::Result
+    where
+        C: Consumer<Self::Item>,
+    {
+        bridge(self, consumer)
+    }
+
+    fn with_producer<CB>(self, callback: CB) -> CB::Output
+    where
+        CB: ProducerCallback<Self::Item>,
+    {
+        let len = self.base.len();
+        return self.base.with_producer(Callback {
+            chunk_size: self.chunk_size,
+            len,
+            item: self.item,
+            fold_op: self.fold_op,
+            callback,
+        });
+
+        struct Callback<CB, T, F> {
+            chunk_size: usize,
+            len: usize,
+            item: T,
+            fold_op: F,
+            callback: CB,
+        }
+
+        impl<T, U, F, CB> ProducerCallback<T> for Callback<CB, U, F>
+        where
+            CB: ProducerCallback<U>,
+            U: Send + Clone,
+            F: Fn(U, T) -> U + Send + Sync,
+        {
+            type Output = CB::Output;
+
+            fn callback<P>(self, base: P) -> CB::Output
+            where
+                P: Producer<Item = T>,
+            {
+                self.callback.callback(FoldChunksWithProducer {
+                    chunk_size: self.chunk_size,
+                    len: self.len,
+                    item: self.item,
+                    fold_op: &self.fold_op,
+                    base,
+                })
+            }
+        }
+    }
+}
+
+struct FoldChunksWithProducer<'f, P, U, F>
+where
+    P: Producer,
+{
+    chunk_size: usize,
+    len: usize,
+    item: U,
+    fold_op: &'f F,
+    base: P,
+}
+
+impl<'f, P, U, F> Producer for FoldChunksWithProducer<'f, P, U, F>
+where
+    P: Producer,
+    U: Send + Clone,
+    F: Fn(U, P::Item) -> U + Send + Sync,
+{
+    type Item = F::Output;
+    type IntoIter = FoldChunksWithSeq<'f, P, U, F>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        FoldChunksWithSeq {
+            chunk_size: self.chunk_size,
+            len: self.len,
+            item: self.item,
+            fold_op: self.fold_op,
+            inner: if self.len > 0 { Some(self.base) } else { None },
+        }
+    }
+
+    fn min_len(&self) -> usize {
+        div_round_up(self.base.min_len(), self.chunk_size)
+    }
+
+    fn max_len(&self) -> usize {
+        self.base.max_len() / self.chunk_size
+    }
+
+    fn split_at(self, index: usize) -> (Self, Self) {
+        let elem_index = min(index * self.chunk_size, self.len);
+        let (left, right) = self.base.split_at(elem_index);
+        (
+            FoldChunksWithProducer {
+                chunk_size: self.chunk_size,
+                len: elem_index,
+                item: self.item.clone(),
+                fold_op: self.fold_op,
+                base: left,
+            },
+            FoldChunksWithProducer {
+                chunk_size: self.chunk_size,
+                len: self.len - elem_index,
+                item: self.item,
+                fold_op: self.fold_op,
+                base: right,
+            },
+        )
+    }
+}
+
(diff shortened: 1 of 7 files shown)


willcrozi:
I've created a small `criterion` benchmark suite to test for various chunk lengths against various equivalent approaches for both 'noop' and 'sum' fold operations: https://github.com/willcrozi/rayon/tree/fold-chunks-bench

Summary graphs here:

noop_bench
sum_bench

Benchmark runs using the new iterators in this PR have a prefix of `par_foldchunks_...` and `par_foldchunkswith_...` and are coloured yellow and blue respectively.

willcrozi:
Hi there! Is there interest in merging something like this? I can see there's some other proposals related to chunking/grouping perhaps making this PR redundant?.

In case there is interest, I feel like there should be a benchmark added in `rayon-demo` since this work is performance related but I don't see an obvious fit in the existing benchmarks. Maybe a new benchmark for some workload where chunking/grouping operations are inherent?

I tried adding a version of the factorial benchmark using `fold_chunks_with` just to see how performance held up. Chunk size was tweaked manually to the best performing power-of-two.
```rust
#[bench]
/// Compute the Factorial using rayon::fold_chunks_with.
fn factorial_fold_chunks_with(b: &mut test::Bencher) {
    fn fact(n: u32) -> BigUint {
        (1..n + 1)
            .into_par_iter()
            .fold_chunks_with(64, BigUint::from(1_u32), |acc, x| acc.mul(x))
            .reduce_with(Mul::mul)
            .unwrap()
    }

    let f = factorial(N);
    b.iter(|| assert_eq!(fact(test::black_box(N)), f));
}
```

Unexpectedly, it's consistently the fastest (ran on Ubuntu 20.04 x86_64 with quad core Sandy Bridge i7-2760QM).
```
running 5 tests
test factorial::factorial_fold_chunks                                   ... bench:   1,396,631 ns/iter (+/- 238,207)
test factorial::factorial_iterator                                      ... bench:  12,819,803 ns/iter (+/- 2,561,952)
test factorial::factorial_join                                          ... bench:   1,590,948 ns/iter (+/- 402,093)
test factorial::factorial_par_iter                                      ... bench:   1,559,018 ns/iter (+/- 314,690)
test factorial::factorial_recursion                                     ... bench:   1,672,888 ns/iter (+/- 399,144)
```

cuviper:
Hi! And sorry for not replying sooner. I do think this is compelling, even if we do explore other grouping/chunking ideas as well. This one seems straightforward enough that it can be a relatively easy option to recommend that users try.

Benchmarking is hard, and I'm generally not satisfied with what we have, but your factorial benchmark looks like a good addition, especially since it's winning. :smile: Note that the `join` and `par_iter` versions are converting every number to `BigUint`, causing extra memory allocation overhead, while your fold is multiplying `BigUint * u32`. Here's a closer comparison:

```rust
#[bench]
/// Compute the Factorial using rayon::fold_chunks_with.
fn factorial_fold_with(b: &mut test::Bencher) {
    fn fact(n: u32) -> BigUint {
        (1..n + 1)
            .into_par_iter()
            .fold_with(BigUint::from(1_u32), |acc, x| acc.mul(x))
            .reduce_with(Mul::mul)
            .unwrap()
    }

    let f = factorial(N);
    b.iter(|| assert_eq!(fact(test::black_box(N)), f));
}
```

I get these results on a AMD Ryzen 7 5800X running Fedora 37:

```
running 6 tests
test factorial::factorial_fold_chunks_with                              ... bench:     360,718 ns/iter (+/- 44,535)
test factorial::factorial_fold_with                                     ... bench:     394,420 ns/iter (+/- 25,759)
test factorial::factorial_iterator                                      ... bench:   5,798,113 ns/iter (+/- 30,231)
test factorial::factorial_join                                          ... bench:     414,736 ns/iter (+/- 23,264)
test factorial::factorial_par_iter                                      ... bench:     413,722 ns/iter (+/- 30,028)
test factorial::factorial_recursion                                     ... bench:     697,092 ns/iter (+/- 9,624)
```

However, adding `.with_min_len(64)` to mine does get its result consistently within noise of yours.

Still, that's not to say that we shouldn't add this! I suspect this kind of chunking is probably easier to understand than the loose `with_min_len`/`with_max_len` controls, even though those are more flexible.

src/iter/mod.rs
```diff
@@ -2417,6 +2421,82 @@ pub trait IndexedParallelIterator: ParallelIterator {
...
+        self,
+        chunk_size: usize,
+        identity: ID,
+        fold_op: F,
+    ) -> FoldChunks<Self, ID, F>
+    where
+        ID: Fn() -> U + Send + Sync,
+        F: Fn(U, Self::Item) -> U + Send + Sync,
+    {
+        assert!(chunk_size != 0, "chunk_size must not be zero");
```
adamreichold, line 2462: I think it would be good to mention the panic here and below in the doc comment.

cuviper: And maybe `#[track_caller]` to follow #977.

willcrozi: Thanks for taking a look!

> Note that the `join` and `par_iter` versions are converting every number to `BigUint`, causing extra memory allocation overhead, while your fold is multiplying `BigUint * u32`....
> ...
> However, adding `.with_min_len(64)` to mine does get its result consistently within noise of yours.

Explains my surprise (should have noticed that!)

I'll push my factorial benchmark here shortly. To address the 'like-for-like' issue should it match the `join` and `par_iter` version by mapping everything to `BigUint` first? E.g.
```rust
            .into_par_iter()
            .map(BigUint::from)
            .fold_chunks_with(64, BigUint::from(1_u32), |acc, x| acc.mul(x))
            ...
```
... or maybe supplement with the fastest version your `factorial_fold_with` instead?

willcrozi: > I suspect this kind of chunking is probably easier to understand than the loose `with_min_len`/`with_max_len` controls, even though those are more flexible.

Yes, and also with `fold_chunks` or `chunks` you can guarantee something like: 'all chunks have size _x_ except for the last chunk which may be smaller', a common requirement for protocols/standards. Whereas, if I understand correctly, using `with_min_len`/`with_max_len` controls would not provide such a guarantee(?).

willcrozi: Also, it would be nice to compare against a version using `par_iter().chunks()...` since this was the original motivation for `fold_chunks`: e.g.
```rust
#[bench]
/// Compute the Factorial using rayon::chunks/map and a sequential fold.
fn factorial_chunks_fold(b: &mut test::Bencher) {
    println!("factorial_fold_chunks, chunk size: {}", N);
    fn fact(n: u32) -> BigUint {
        (1..n + 1)
            .into_par_iter()
            .chunks(64)
            .map(|chunk| chunk.into_iter().fold(BigUint::from(1_u32), |acc, x| acc.mul(x)))
            .reduce_with(Mul::mul)
            .unwrap()
    }

    let f = factorial(N);
    b.iter(|| assert_eq!(fact(test::black_box(N)), f));
}
```
However the performance is within noise of `fold_chunks` and the difference seen in the criterion benchmark results above is not seen (use of `BigUint` masks it?). I think it would need to be a whole new benchmark module which didn't require allocation in its 'core' operation, something like summing an iterator of `u32`s into a `u64` perhaps.

willcrozi: Agreed, however there's quite a few places that do the same without mentioning in the doc comment (`IndexedParallelIterator::chunks` and various methods on `ParallelSlice`). Perhaps it would be a cleaner git history to address all of these together in their own PR?

willcrozi:
Thanks for taking a look!

> Note that the `join` and `par_iter` versions are converting every number to `BigUint`, causing extra memory allocation overhead, while your fold is multiplying `BigUint * u32`....
> ...
> However, adding `.with_min_len(64)` to mine does get its result consistently within noise of yours.

Explains my surprise (should have noticed that!)

I'll push my factorial benchmark here shortly. To address the 'like-for-like' issue should it match the `join` and `par_iter` version by mapping everything to `BigUint` first? E.g.
```rust
            .into_par_iter()
            .map(BigUint::from)
            .fold_chunks_with(64, BigUint::from(1_u32), |acc, x| acc.mul(x))
            ...
```
... or maybe supplement with the fastest version your `factorial_fold_with` instead?

willcrozi:
> I suspect this kind of chunking is probably easier to understand than the loose `with_min_len`/`with_max_len` controls, even though those are more flexible.

Yes, and also with `fold_chunks` or `chunks` you can guarantee something like: 'all chunks have size _x_ except for the last chunk which may be smaller', a common requirement for protocols/standards. Whereas, if I understand correctly, using `with_min_len`/`with_max_len` controls would not provide such a guarantee(?).

willcrozi:
Also, it would be nice to compare against a version using `par_iter().chunks()...` since this was the original motivation for `fold_chunks`: e.g.
```rust
#[bench]
/// Compute the Factorial using rayon::chunks/map and a sequential fold.
fn factorial_chunks_fold(b: &mut test::Bencher) {
    println!("factorial_fold_chunks, chunk size: {}", N);
    fn fact(n: u32) -> BigUint {
        (1..n + 1)
            .into_par_iter()
            .chunks(64)
            .map(|chunk| chunk.into_iter().fold(BigUint::from(1_u32), |acc, x| acc.mul(x)))
            .reduce_with(Mul::mul)
            .unwrap()
    }

    let f = factorial(N);
    b.iter(|| assert_eq!(fact(test::black_box(N)), f));
}
```
However the performance is within noise of `fold_chunks` and the difference seen in the criterion benchmark results above is not seen (use of `BigUint` masks it?). I think it would need to be a whole new benchmark module which didn't require allocation in its 'core' operation, something like summing an iterator of `u32`s into a `u64` perhaps.

willcrozi:
So I've added my best guess as to what's needed (too many questions probably! :upside_down_face:). I've kept the benchmark performing `BigUint * u32` but also added your version for a fairer comparison.

Let me know if you'd like the commits squashed/rebased onto main etc.

src/iter/fold_chunks_with.rs
```diff
@@ -0,0 +1,286 @@
...
+        }
+    }
+
+    fn min_len(&self) -> usize {
+        div_round_up(self.base.min_len(), self.chunk_size)
+    }
+
+    fn max_len(&self) -> usize {
+        self.base.max_len() / self.chunk_size
+    }
```
hhirtz, line 172: Hi, sorry if it's not an actual issue but if I understand correctly `FoldChunks{,With}Producer::max_len` must also use `div_round_up`, otherwise `min_len` may end up larger than `max_len`. For example with the following configuration:

- `self.base.max_len() == 3`
- `self.base.min_len() == 3`
- `self.chunk_size == 2`

then `min_len == 2` and `max_len == 1`.

willcrozi: I think you're correct, good catch! I took this from the `chunks::ChunkProducer` impl without properly checking :flushed:

@cuviper: shouldn't the impl in `chunks::ChunkProducer` be doing the same thing?

cuviper: Well, you can already set inconsistent values using `with_min_len` and `with_max_len`, and the way it's currently [documented](https://docs.rs/rayon/latest/rayon/iter/trait.IndexedParallelIterator.html#method.with_max_len) and implemented is that the minimum "wins". The maximum is kind of a soft limit, and we might not be able to satisfy both even when the maximum _is_ more than the minimum.

With chunks, either here or the `chunks` adaptor, it seems messier -- but I think it's correct the way it is. We want to find `n` such that `chunk_size * n` respects the inner limit. So for the minimum, we round `n` up to make sure `chunk_size * n` is never less than the inner minimum, and likewise the maximum should round down so `chunk_size * n` is never more than the inner maximum.

With the example of inner `min == max == 3`, and `chunk_size == 2`, there's no possible way to satisfy this entirely. If we pick `n == 1`, then `n * chunk_size == 2` violates the inner minimum, but with `n == 2` then `n * chunk_size == 4` violates the inner maximum. So, I think it's simplest if `min_len` and `max_len` go ahead and compute their ideal limits independently, and then the rest of the code that *uses* these limits will favor the minimum, as I mentioned before.

src/iter/mod.rs
```diff
@@ -2417,6 +2421,88 @@ pub trait IndexedParallelIterator: ParallelIterator {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    /// let nums = vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+    /// let chunk_sums = nums.into_par_iter().fold_chunks(2, || 0, |a, n| a + n).collect::<Vec<_>>();
+    /// assert_eq!(chunk_sums, vec![3, 7, 11, 15, 19]);
+    /// ```
+    #[track_caller]
+    fn fold_chunks<ID, F, U>(
```
cuviper, line 2455: This is a tiny thing, but maybe this should be `fold_chunks<T, ID, F>` to be consistent with `fold`?

src/iter/mod.rs
```diff
@@ -2417,6 +2421,88 @@ pub trait IndexedParallelIterator: ParallelIterator {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use rayon::prelude::*;
+    /// let nums = vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+    /// let chunk_sums = nums.into_par_iter().fold_chunks_with(2, 0, |a, n| a + n).collect::<Vec<_>>();
+    /// assert_eq!(chunk_sums, vec![3, 7, 11, 15, 19]);
+    /// ```
+    #[track_caller]
+    fn fold_chunks_with<T, F>(
```
cuviper, line 2492: But then we have the inconsistency of `fold_with<F, T>`, i.e. `T` second, which I never noticed before...

cuviper (approved):
Thanks, I think this is good! Just one last consideration on generic params.

I also have some ideas of how to refactor common code between the various "chunks" APIs, but I'll try to follow up separately. That won't affect any public API, at least.

src/iter/mod.rs
willcrozi: I've gone with consistency with `fold`, which is consistent with 'std::iter::Iterator::fold` (closure argument goes last).

willcrozi:
I've noticed there's quite a few tests for `chunks` in `iter/test` and so have added the equivalent to `fold_chunks` and `fold_chunks_with`, I've kept them in their respective modules for now (can move over to `test` module if desired).

There's quite a lot of duplication though so it's understandable if you'd rather not include these.

cuviper:
OK, I added another `T: Send` for consistency, and applied test formatting, but otherwise I think I've dithered long enough. :)

bors r+

Issue #971, opened 2022-08-20
Feature request: `.tuples()`


WhyNotHugo:
I see `tuples()` listed in [Probably applicable but did not yet open issues](https://github.com/rayon-rs/rayon/wiki/Itertools-Parity#probably-applicable-but-did-not-yet-open-issues)

I'd find this very useful. I'm re-encoding images from LE-XRBG to BE-RGB:

```rust
    let mmap: memmap2::Mmap = ...
    let mut image_data = vec![0u8; mmap.len() * 3 / 4];
    mmap.iter()
        .tuples()
        .zip(image_data.iter_mut().tuples())
        .for_each(|((b, g, r, _), (r2, g2, b2))| (*r2, *g2, *b2) = (*r, *g, *b));
```

This can clearly be parallelised quite a bit, but I don't see any obvious way of writing it without `tuples()`.

cuviper:
The ugly part of `tuples()` is trying to be variadic in length, which doesn't have any language support. We deal with this a little in the `IntoParallelIterator` for tuples into a `MultiZip`, which we manually (macro) implement up to length 12. `Itertools` depends on a similar manual implementation of its own `HomogeneousTuple` trait.

For your particular case with slice-able data, you could also use `par_chunks_exact(4)` and `par_chunks_exact_mut(3)`. You can still pattern-match that if you treat it as a refutable pattern within your `for_each`, like `if let ([b, g, r, _], [r2, g2, b2]) {...}`, even though that `if let` will never fail in practice. Eventually it can be a truly irrefutable argument pattern with arrays, #789, but I've been hoping that the standard library would stabilize the regular `Iterator` API for that first.

cuviper:
> I'm re-encoding images from LE-XRBG to BE-RGB:

> This can clearly be parallelised quite a bit,

FWIW, it's not necessarily true that this task will benefit from pixel-level parallelization, especially since there's little real CPU work, just data movement. If you have many images to convert, it may work better to parallelize that coarsely with each image working serially, but ultimately you're going to be limited by I/O and memory bandwidth.

WhyNotHugo:
I tend to agree with that last statement; it seems to be mostly memory IO limited (rather than CPU), so it might be a pointless optimisation. FWIW, the original code can also be re-written as:

```rust
    let mut image_data = vec![0u8; mmap.len() * 3 / 4];
    mmap.iter()
        .tuples()
        .zip(image_data.iter_mut().tuples())
[...]

PR #974, opened 2022-08-31
Add const-generic `IndexedParallelIterator::arrays`

cuviper:
`arrays<N>()` works like `chunks(N)` without remainder, and yielding `[Item; N]` rather than allocated `Vec<Item>`.

There's no sequential precedent for this yet, but it's proposed in [itertools#549](https://github.com/rust-itertools/itertools/pull/549). It may also be useful in places that would want `Itertools::tuples` (#971), because arrays are a first-class alternative to homogeneous tuples.


src/iter/mod.rs
```diff
@@ -2417,6 +2421,36 @@ pub trait IndexedParallelIterator: ParallelIterator {
...
     }
 
+    /// Splits an iterator up into fixed-size arrays.
+    ///
+    /// Returns an iterator that returns arrays with the given number of elements.
+    /// If the number of elements in the iterator is not divisible by `chunk_size`,
+    /// the remaining items are ignored.
+    ///
+    /// See also [`par_array_chunks()`] and [`par_array_chunks_mut()`] for similar
+    /// behavior on slices, although they yield array references instead.
```
cuviper, line 2431: This reference depends on #789.

cuviper:
I missed that the standard library now has an unstable `Iterator::array_chunks` -- rust-lang/rust#100450.

PR #977, opened 2022-09-18
Attribute invalid input assertions to callers

kornelski:
`#[track_caller]` makes assert location point to the caller of the function. This helps users find which of their calls provided the invalid inputs.

### DIFF
--- a/rayon-core/README.md
+++ b/rayon-core/README.md
@@ -8,4 +8,4 @@ Please see [Rayon Docs] for details about using Rayon.
 
 [Rayon Docs]: https://docs.rs/rayon/
 
-Rayon-core currently requires `rustc 1.36.0` or greater.
+Rayon-core currently requires `rustc 1.46.0` or greater.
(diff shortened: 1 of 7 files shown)


kornelski:
It requires Rust 1.46

src/iter/mod.rs
```diff
@@ -2341,6 +2341,7 @@ pub trait IndexedParallelIterator: ParallelIterator {
     /// // we should never get here
     /// assert_eq!(1, zipped.len());
     /// ```
+    #[track_caller]
```
cuviper, line 2344: Since we're changing the panic location, the assertion below needs textual context, something like "iterators must have the same length".

rayon-core/README.md
```diff
@@ -8,4 +8,4 @@ Please see [Rayon Docs] for details about using Rayon.
 
 [Rayon Docs]: https://docs.rs/rayon/
 
-Rayon-core currently requires `rustc 1.36.0` or greater.
+Rayon-core currently requires `rustc 1.46.0` or greater.
```
cuviper, line 11: Technically, `rayon-core` is unaffected here, though it will be untested unless we add another CI job.

But I intend to bump to 2021 in about a month (#964), so I probably won't publish until then.

PR #978, opened 2022-09-21
Fix typos

kianmeng:
Found via `codespell -L crate,fromt,deques`

### DIFF
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -560,7 +560,7 @@ impl<S> ThreadPoolBuilder<S> {
     /// to true, however, workers will prefer to execute in a
     /// *breadth-first* fashion -- that is, they will search for jobs at
     /// the *bottom* of their local deque. (At present, workers *always*
-    /// steal from the bottom of other worker's deques, regardless of
+    /// steal from the bottom of other workers' deques, regardless of
     /// the setting of this flag.)
     ///
     /// If you think of the tasks as a tree, where a parent task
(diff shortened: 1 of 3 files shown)


rayon-core/src/lib.rs
```diff
@@ -560,7 +560,7 @@ impl<S> ThreadPoolBuilder<S> {
     /// to true, however, workers will prefer to execute in a
     /// *breadth-first* fashion -- that is, they will search for jobs at
     /// the *bottom* of their local deque. (At present, workers *always*
-    /// steal from the bottom of other worker's deques, regardless of
+    /// steal from the bottom of other worker's dequeues, regardless of
```
cuviper, line 563: This is not a typo -- just the plural of "deque" that you left on the line above.

rayon-core/src/registry.rs
```diff
@@ -719,7 +719,7 @@ impl WorkerThread {
         while !latch.probe() {
             // Try to find some work to do. We give preference first
             // to things in our local deque, then in other workers
-            // deques, and finally to injected jobs from the
+            // dequeues, and finally to injected jobs from the
```
cuviper, line 722: Ditto, not a typo.

src/slice/mergesort.rs
```diff
@@ -570,7 +570,7 @@ unsafe fn recurse<T, F>(
 
     // After recursive calls finish we'll have to merge chunks `(start, mid)` and `(mid, end)` from
     // `src` into `dest`. If the current invocation has to store the result into `buf`, we'll
-    // merge chunks from `v` into `buf`, and viceversa.
+    // merge chunks from `v` into `buf`, and vice-versa.
```
cuviper, line 573: This should just be two words.

```suggestion
    // merge chunks from `v` into `buf`, and vice versa.
```

cuviper (changes requested):

rayon-core/src/lib.rs
cuviper: The plural possessive should be updated though...
```suggestion
    /// steal from the bottom of other workers' deques, regardless of
```

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #979, opened 2022-09-23
Fix tiny typo in licensing section of readme

idanmuze:
Right now it says: 
_"Opening a pull requests is assumed to signal agreement with these licensing terms."_ 

It should probably instead say one of the two below:

- _"Opening a pull REQUEST is assumed to signal agreement with these licensing terms."_
- _"The opening of pull requests is assumed to signal agreement with these licensing terms."_


cuviper:
You're right, thanks!

bors r+

idanmuze:
Keep up the good work!

Issue #713, opened 2019-12-18
Feature request: try_collect_into_vec


jplatte:
Using the `Result` impl for `FromParallelIterator`, one can already collect the result of many fallible operations into a `Result<Vec<_>, _>`. However, AFAICT, it is currently impossible to do this multiple times while reusing the same results buffer, which is possible for infallible operations using `collect_into_vec`. It would be nice if this was supported. I'd expect the semantics around errors to be the same as those for `try_for_each`, with the backing `Vec` additionally being cleared if an `Err` is encountered.

cuviper:
My initial thought was that we could add `ParallelExtend for Result`, but this wouldn't let you keep your backing storage in case of an `Err`.

> with the backing `Vec` additionally being cleared if an `Err` is encountered.

It would be easy to `set_len(0)` on `Err` -- or really with the way `src/iter/collect/` works now we would just leave it at 0 to begin with, never setting the final length. But it's harder to properly drop the partial writes because they're not sequential, and at least right now we don't keep any map of what was already written. We kind of _have_ to keep it clear because of that, and those partial items will just be leaked. The immediate `Vec` memory will be reclaimed, but I mean we won't run `Drop` for those items, so they will leak any indirect memory, file handles, etc.

We already leak like this if you panic during a `collect_into_vec` -- see the [everyone poops](https://cglab.ca/~abeinges/blah/everyone-poops/) post for some history on that kind of design. But panics are meant to be infrequent, exceptional, whereas leaking for normal `Result` error handling seems like more of a problem.

> Is this just something that hasn't been implemented yet, or have I missed something that makes this impossible to implement? Would a PR for this be accepted if it can be done? (I haven't previously contributed to rayon and don't know whether I'd be able to do it, but I might try at least)

I wouldn't say it's impossible, but I think it's non-trivial given my concerns above. I'm happy to try to mentor you on this though, talking through possible designs, and then implementation if we figure out something satisfactory.

jplatte:
> My initial thought was that we could add `ParallelExtend for Result`, but this wouldn't let you keep your backing storage in case of an `Err`.

For my current use case, I'm only interested in keeping the backing storage as long as the repeated calls all return `Ok(vec![...])`. However, `std::iter::Extend` isn't implemented for `Result`, so this would be a deviation from `std`. I'm also having trouble wrapping my head around what such an impl would do (though I haven't tried for very long yet).
 
[...]

PR #987, opened 2022-11-04
Add `IndexedParallelIterator::[try_]fill_slice`

SabrinaJewson:
Motivation: #713 is difficult to implement without external tracking of which items have been written in, so instead we offer `try_fill_slice` which requires the user to provide dummy values that can always be dropped (which isn’t usually too difficult). The implementation uses entirely safe code.

Public API:
```rs
trait IndexedParallelIterator {
    fn fill_slice(self, target: &mut [Self::Item]);

    fn try_fill_slice<T>(
        self,
        target: &mut [T],
    ) -> <<Self::Item as Try>::Residual as Residual<()>>::TryType
    where
        Self::Item: Try<Output = T>,
        T: Send,
        <Self::Item as Try>::Residual: Residual<()> + Send;
}
```

The reason for adding `Residual` is to enable the generic function of `Option<T>` → `Option<()>`, `Result<T, E>` → `Result<(), E>` etc which wouldn’t be possible otherwise.

### DIFF
--- a/src/iter/fill_slice.rs
+++ b/src/iter/fill_slice.rs
@@ -1,0 +1,80 @@
+use super::{noop::NoopReducer, plumbing, IndexedParallelIterator};
+
+/// Fills the slice with the contents of the parallel iterator.
+///
+/// This is called by `IndexedParallelIterator::fill_slice`.
+pub(super) fn fill_slice<I, T>(pi: I, target: &mut [T])
+where
+    I: IndexedParallelIterator<Item = T>,
+    T: Send,
+{
+    assert_eq!(
+        pi.len(),
+        target.len(),
+        "slice length not equal to iterator length"
+    );
+
+    pi.drive(Consumer { target })
+}
+
+struct Consumer<'target, T> {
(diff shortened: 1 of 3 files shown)


cuviper:
I'm wary of replicating the `TryType` type constructor, because eventually we'd like to transition to the official `Try`, but nothing in `std` has stabilized this `TryType` mechanism yet. I wouldn't want to add something in our API that makes us incompatible with whatever `std` lands on. At least the `Try` trait is constrained by the stable `?` operator and `try_fold`, etc.

But if you have a ready fill buffer, can't you (as a user) already do this with `for_each` and `try_for_each`?

```rust
my_par_iter.zip_eq(&mut target).try_for_each(|(src, dest)| {
    *dest = src?;
    Ok(()) // or whatever your Try type is...
)
```

(i.e. we don't need generic `TryType` if you're writing that for your specific situation.)

SabrinaJewson:
Ah, I didn’t realize that I could do that 😅. Thanks a lot, I’ll probably close this then

PR #989, opened 2022-11-16
Unify `chunks`, `fold_chunks`, and `fold_chunks_with`

cuviper:


cuviper:
We only have `factorial_fold_chunks_with` as a benchmark, but its result is unaffected by this change, so I think it's safe from that perspective. Otherwise there's no public API change here, only code reduction.

bors r+

Issue #795, opened 2020-09-02
Extra CPU usage when not enough work in par_bridge


hniksic:
On a particular workload I noticed that Rayon uses all available CPU, but doesn't do its best, i.e.  does better when the number of threads is reduced. The workload involves work that is easy to parallelize (essentially order-insensitive stream processing using `par_bridge()` with almost no mutable shared state) running on a 64-CPU server. On my laptop the it nicely parallelizes to 4 cores, running approximately 4x faster than sequentially. But on the 64-CPU server the work definitely doesn't proceed with speed anywhere near 64x of the sequential version on the same machine, while all the CPUs are churning at 100% the entire time.

Experimenting with `RAYON_NUM_THREADS` shows that performance peaks around 8-16 CPUs and gets worse with 32 and 64:

* 1 thread: 6.5s / million records
* 2 threads: 3.5s
* 4 threads: 1.9s
* 8 threads: 1s
* 16 threads: 0.9s
* 32 threads: 1.7s
* 64 threads (default): 1.7s

Measurements were done with release build and Rayon 1.4.0. The interesting thing is that in all tests all cores were at 100%, as measured by `htop`.

I was unable to reproduce the issue with simpler minimal examples, where Rayon seemed to distribute the load nicely. So I rewrote the code to use a `crossbeam_channel` to distribute the work to 64 manually created threads instead. Using that, it took 3.8s to process a million records, but all cores were only at 15-20% of CPU usage. I then added some batching (a thousand records in a batch) and sent the batches over the channel to reduce the overhead. With this change I got the speed of the channel version to 0.65s/million records, with CPUs averaging about 25% usage.

Of course, now I had to implement similar batching in Rayon, which I did, bringing the numbers to:

* 1 thread: 6.4s/million records
* 2 threads: 3.4s
* 4 threads: 1.76s
* 8 threads: 0.91s
* 16 threads: 0.77s
* 32 threads: 0.77s
* 64 threads (default): 0.85s

With batching Rayon is no longer able to outperform channel-based dispatch, although it comes close at its best numbers. Note that Rayon is always using 100% of the cores it runs its threads on, and in case of 64 threads it uses 100% of all cores, despite being slower than channel variant, which only uses about 25% of the cores.

[...]

PR #996, opened 2022-12-05
par_bridge: use naive locking of the Iterator

njaard:
Fix issue #795

instead of using crossbeam_deque, just lock the Iterator and get the next item, naively. The original code would spin until more data became available, which would cause par_bridge to not be useful in cases where the source Iterator did slow IO or was computationally costly in itself.

This causes there to be virtually no change in runtime in many cases, but in cases where there are a lot CPUs and even slightly slower IO, significantly less runtime.

In pretty much every case, total CPU time was much lower.

I ran the below test program on a 1.5GiB file from a fast SSD and on a machine with 64 cores and mean runtime went from 44s to 30s. In fact, according to hyperfine, average runtime with this patch was lower than _minimum_ runtime without (25s and 41s, respectively). But more importantly, total CPU time was almost 10x less with this patch.

I'm not sure what realistic usage patterns could be worse with this patch and therefor I'd love to hear other opinions.

I wrote a basic test program:

```
use std::io::{BufReader,BufRead};
use rayon::prelude::*;
use std::sync::atomic::{Ordering,AtomicUsize};
use std::os::unix::io::FromRawFd;

fn main() {
  let counter = AtomicUsize::new(0);
  let stdin = BufReader::new(unsafe { std::fs::File::from_raw_fd(0) });

  stdin.lines().par_bridge()
    .for_each( |row| {
      let row = row.unwrap();
      for _ in 0 .. 1000
      {
        let c = row.len();
        counter.fetch_add(c, Ordering::Relaxed);
      }
    }
  );

  println!("{}", counter.load(Ordering::Relaxed));
}
```

Which directly wastes some CPU time (`for _ in 0 .. 1000`). The unsafe is necessary to make the input stream Send. I feed this program a multigigabyte file as part of a test. Once over a relatively slow network and once off an SSD.

### DIFF
--- a/src/iter/par_bridge.rs
+++ b/src/iter/par_bridge.rs
@@ -1,4 +1,4 @@
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Mutex;
 
 use crate::current_num_threads;
@@ -77,13 +77,11 @@ where
     {
         let split_count = AtomicUsize::new(current_num_threads());
 
-        let done = AtomicBool::new(false);
         let iter = Mutex::new(self.iter.fuse());
 
         bridge_unindexed(
             IterParallelProducer {
                 split_count: &split_count,
-                done: &done,
                 iter: &iter,
             },
             consumer,
@@ -93,7 +91,6 @@ where
 
 struct IterParallelProducer<'a, Iter: Iterator> {
     split_count: &'a AtomicUsize,
-    done: &'a AtomicBool,
     iter: &'a Mutex<std::iter::Fuse<Iter>>,
 }
 
@@ -102,7 +99,6 @@ impl<'a, Iter: Iterator + 'a> Clone for IterParallelProducer<'a, Iter> {
     fn clone(&self) -> Self {
         IterParallelProducer {
             split_count: self.split_count,
-            done: self.done,
             iter: self.iter,
         }
     }
@@ -119,22 +115,18 @@ where
 
         loop {
             // Check if the iterator is exhausted
-            let done = self.done.load(Ordering::SeqCst);
-            match count.checked_sub(1) {
-                Some(new_count) if !done => {
-                    match self.split_count.compare_exchange_weak(
-                        count,
-                        new_count,
-                        Ordering::SeqCst,
-                        Ordering::SeqCst,
-                    ) {
-                        Ok(_) => return (self.clone(), Some(self)),
-                        Err(last_count) => count = last_count,
-                    }
+            if let Some(new_count) = count.checked_sub(1) {
+                match self.split_count.compare_exchange_weak(
+                    count,
+                    new_count,
+                    Ordering::SeqCst,
+                    Ordering::SeqCst,
+                ) {
+                    Ok(_) => return (self.clone(), Some(self)),
+                    Err(last_count) => count = last_count,
                 }
-                _ => {
-                    return (self, None);
-                }
+            } else {
+                return (self, None);
             }
         }
     }
@@ -152,13 +144,11 @@ where
                         return folder;
                     }
                 } else {
-                    self.done.store(true, Ordering::SeqCst);
                     return folder;
                 }
             } else {
                 // any panics from other threads will have been caught by the pool,
                 // and will be re-thrown when joined - just exit
-                self.done.store(true, Ordering::SeqCst);
                 return folder;
             }
         }
--- a/src/iter/par_bridge.rs
+++ b/src/iter/par_bridge.rs
@@ -1,9 +1,9 @@
 use std::sync::Mutex;
 
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use crate::current_num_threads;
 use crate::iter::plumbing::{bridge_unindexed, Folder, UnindexedConsumer, UnindexedProducer};
 use crate::iter::ParallelIterator;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 
 /// Conversion trait to convert an `Iterator` to a `ParallelIterator`.
 ///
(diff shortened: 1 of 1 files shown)


njaard:
@cupiver good idea

I ran the two demos that that appear to reference par_bridge, `nbody` and `life`.

# Without my patch
```
$ hyperfine "./target/release/rayon-demo  life  bench"
Benchmark 1: ./target/release/rayon-demo  life  bench
  Time (mean ± σ):     811.4 ms ±  26.4 ms    [User: 2784.5 ms, System: 81.1 ms]
  Range (min … max):   751.2 ms … 849.7 ms    10 runs

$ hyperfine "./target/release/rayon-demo  nbody bench"
Benchmark 1: ./target/release/rayon-demo  nbody bench
  Time (mean ± σ):      7.660 s ±  0.212 s    [User: 17.835 s, System: 0.044 s]
  Range (min … max):    7.416 s …  8.070 s    10 runs
```

# With my patch
```
$ hyperfine "./target/release/rayon-demo  life  bench"
Benchmark 1: ./target/release/rayon-demo  life  bench
  Time (mean ± σ):     503.0 ms ±  21.3 ms    [User: 1073.8 ms, System: 472.3 ms]
  Range (min … max):   487.0 ms … 548.7 ms    10 runs

$ hyperfine "./target/release/rayon-demo  nbody bench"
Benchmark 1: ./target/release/rayon-demo  nbody bench
  Time (mean ± σ):      7.699 s ±  0.144 s    [User: 18.002 s, System: 0.051 s]
  Range (min … max):    7.507 s …  7.922 s    10 runs
  
```

# Summary
The "nbody" task is slightly worse with my patch, but not statistically significantly. The "life" task is considerably faster with my patch.

# Machine
Performed on a 4-core CPU

njaard:
# With my patch
```
test nbody::bench::nbody_par_bridge                                     ... bench:  10,193,883 ns/iter (+/- 14,547,250)
test nbody::bench::nbody_par_iter                                       ... bench:  10,313,040 ns/iter (+/- 9,173,607)
test nbody::bench::nbody_parreduce                                      ... bench:  21,930,914 ns/iter (+/- 10,491,332)
test nbody::bench::nbody_seq                                            ... bench:  37,272,289 ns/iter (+/- 34,929,418)

test life::bench::generations                                           ... bench:  73,372,701 ns/iter (+/- 6,076,736)
test life::bench::par_bridge_generations                                ... bench: 384,203,006 ns/iter (+/- 29,417,758)
test life::bench::par_iter_generations                                  ... bench:  23,879,561 ns/iter (+/- 1,021,068)
```

# Without my patch
```
test nbody::bench::nbody_par_bridge                                     ... bench:  10,263,138 ns/iter (+/- 9,227,833)
test nbody::bench::nbody_par_iter                                       ... bench:  11,594,927 ns/iter (+/- 12,503,276)
test nbody::bench::nbody_parreduce                                      ... bench:  22,593,455 ns/iter (+/- 10,537,023)
test nbody::bench::nbody_seq                                            ... bench:  36,330,253 ns/iter (+/- 35,849,257)

test life::bench::generations                                           ... bench:  71,825,838 ns/iter (+/- 3,260,082)
test life::bench::par_bridge_generations                                ... bench: 738,822,305 ns/iter (+/- 135,174,311)
test life::bench::par_iter_generations                                  ... bench:  24,250,743 ns/iter (+/- 1,219,741)
```

Looks like my patch provides a huge benefit.

src/iter/par_bridge.rs
```diff
@@ -1,12 +1,9 @@
...
-
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::sync::{Mutex, TryLockError};
-use std::thread::yield_now;
+use std::sync::Mutex;
 
 use crate::current_num_threads;
 use crate::iter::plumbing::{bridge_unindexed, Folder, UnindexedConsumer, UnindexedProducer};
 use crate::iter::ParallelIterator;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
```
cuviper, line 6: Can you move this up with the other `use std::` statement? Or move the other one down, either way is fine as long as it's consistent. (I think `rustfmt` doesn't touch the order of "groups" separated by empty lines.)

src/iter/par_bridge.rs
```diff
@@ -126,8 +119,7 @@ where
 
         loop {
             // Check if the iterator is exhausted *and* we've consumed every item from it.
```
cuviper, line 121: The `&&` part is gone now that each split is pulling one item at a time for their own consumption.

```suggestion
            // Check if the iterator is exhausted.
```

src/iter/par_bridge.rs
```diff
@@ -152,65 +144,22 @@ where
         F: Folder<Self::Item>,
     {
         loop {
-            match self.items.steal() {
-                Steal::Success(it) => {
+            if let Ok(mut iter) = self.iter.lock() {
+                if let Some(it) = iter.next() {
```
cuviper, line 148: We either need to check `done` before calling this, or else `.fuse()` the iterator, so we never call the underlying iterator again after `None`.

cuviper: I wonder if we should just get rid of `done` entirely in favor of just fusing the iterator, since we're now expecting each consumer to lock for every item already. It would be simpler not to micro-optimize that end state.

njaard: @cuviper Getting rid of `done` would require the Iterator to also be `Peekable`, so that we know when to stop `split`ting. I think that's going a little bit out of the realm of "micro-optimization"

cuviper: The splits are still limited by `split_count`, which starts as the number of threads in the pool. I think that's probably not enough to worry about optimizing away, but it's not a big deal either way.

njaard: I agree, implemented now without `done`

adamreichold:
Does the removal of `done` have any measurable effect on the benchmarks?

cuviper:
Looks good, thanks! I'm going to followup with an attempt at preventing mutex recursion (#690), then we can publish.

bors r+

PR #997, opened 2022-12-08
Guard `par_bridge` against work-stealing recursion

cuviper:
It doesn't make sense for the `par_bridge` producer to run nested on one
thread, since each split is capable of running the entire iterator to
completion. However, this can still happen if the serial iterator or the
consumer side make any rayon calls that would block, where it may go
into work-stealing and start another split of the `par_bridge`. With the
iterator in particular, this is a problem because we'll already be
holding the mutex, so trying to lock again will deadlock or panic.

We now set a flag in each thread when they start working on the bridge,
and bail out if we re-enter that bridge again on the same thread. The
new `par_bridge_recursion` test would previously hang almost every time
for me, but runs reliably with this new check in place.

### DIFF
--- a/src/iter/par_bridge.rs
+++ b/src/iter/par_bridge.rs
@@ -1,9 +1,9 @@
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::Mutex;
 
-use crate::current_num_threads;
 use crate::iter::plumbing::{bridge_unindexed, Folder, UnindexedConsumer, UnindexedProducer};
 use crate::iter::ParallelIterator;
+use crate::{current_num_threads, current_thread_index};
 
 /// Conversion trait to convert an `Iterator` to a `ParallelIterator`.
 ///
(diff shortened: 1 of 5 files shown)


src/iter/par_bridge.rs
```diff
@@ -135,6 +123,25 @@ where
...
+        // (If this isn't a rayon thread, then there's no work-stealing anyway...)
+        if let Some(i) = current_thread_index() {
+            // Note: If the number of threads in the pool ever grows dynamically, then
+            // we'll end up sharing flags and may falsely detect recursion -- that's
+            // still fine for overall correctness, just not optimal for parallelism.
+            let thread_started = &self.threads_started[i % self.threads_started.len()];
+            if thread_started.swap(true, Ordering::Relaxed) {
+                // We can't make progress with a nested mutex, so just return and let
+                // the outermost loop continue with the rest of the iterator items.
+                // eprintln!("par_bridge recursion detected!");
```
adamreichold, line 140: Looks like left-over debug printing?

src/iter/par_bridge.rs
```diff
@@ -135,6 +123,25 @@ where
...
         F: Folder<Self::Item>,
     {
+        // Guard against work-stealing-induced recursion, in case `Iter::next()`
+        // calls rayon internally, so we don't deadlock our mutex. We might also
+        // be recursing via `folder` methods, which doesn't present a mutex hazard,
+        // but it's lower overhead for us to just check this once, rather than
+        // updating additional shared state on every mutex lock/unlock.
+        // (If this isn't a rayon thread, then there's no work-stealing anyway...)
+        if let Some(i) = current_thread_index() {
+            // Note: If the number of threads in the pool ever grows dynamically, then
```
adamreichold, line 133: Two ideas:

* If this potentially approximate in any case, couldn't we use e.g. an `AtomicU64` as a bitmask indicating which of the 64 groups of threads has started work on the bridge and avoid the `Vec`?
* Since we deadlock when the current thread already owns the mutex, wouldn't it suffice to store the current owner's thread ID instead of separate flags and then check if the current thread already holds the mutex? (I am admittedly unsure how to properly synchronize holding the mutex and- updating/checking the current-owner-thread-ID.)

adamreichold: > (I am admittedly unsure how to properly synchronize holding the mutex and updating/checking the current-owner-thread-ID.)

But that might not be a problem since we only really care about the case of synchronizing with ourselves, right? So updating the owner ID from "none" to "me" after locking and before unlocking should suffice so that the current thread can load it and check "it is none or not me" before trying to lock, shouldn't it? (Because either we are racing with someone else and don't care or we are going to deadlock and then it is guaranteed that the atomic was updated before we got here for the second locking attempt.)

src/iter/par_bridge.rs
cuviper: Ah, yeah, I was confirming that my testcase actually got into the bad situation. :)

PR #1011, opened 2023-01-19
Use pointers instead of `&self` in `Latch::set`

cuviper:
`Latch::set` can invalidate its own `&self`, because it releases the
owning thread to continue execution, which may then invalidate the latch
by deallocation, reuse, etc. We've known about this problem when it
comes to accessing latch fields too late, but the possibly dangling
reference was still a problem, like rust-lang/rust#55005.

The result of that was rust-lang/rust#98017, omitting the LLVM attribute
`dereferenceable` on references to `!Freeze` types -- those containing
`UnsafeCell`. However, miri's Stacked Borrows implementation is finer-
grained than that, only relaxing for the cell itself in the `!Freeze`
type. For rayon, that solves the dangling reference in atomic calls, but
remains a problem for other fields of a `Latch`.

This easiest fix for rayon is to use a raw pointer instead of `&self`.
We still end up with some temporary references for stuff like atomics,
but those should be fine with the rules above.

### DIFF
--- a/rayon-core/src/latch.rs
+++ b/rayon-core/src/latch.rs
@@ -1,3 +1,5 @@
+use std::marker::PhantomData;
+use std::ops::Deref;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Condvar, Mutex};
 use std::usize;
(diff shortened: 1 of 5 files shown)


rayon-core/src/latch.rs
```diff
@@ -374,7 +383,7 @@ where
     L: Latch,
 {
     #[inline]
-    fn set(&self) {
-        L::set(self);
+    unsafe fn set(this: *const Self) {
+        L::set(&**this);
```
saethlin, line 387:
```suggestion
        L::set(*this);
```

rayon-core/src/latch.rs
```diff
@@ -374,7 +383,7 @@ where
...
     }
```
saethlin, line 388: This impl makes me uncomfortable. I previously tried to implement this PR and I think I came up with almost exactly the same patch as you have here, but I wasn't sure what to do with this. Of course I couldn't come up with a case where this actually causes an issue, but implementing this for `&Latch` seems like an invitation to create the references we're trying to avoid.

cuviper: I added a new `LatchRef` to replace this -- let me know what you think!

saethlin: That is very cool. I like it.

PR #1012, opened 2023-01-19
Add TakeAny and SkipAny

yu-re-ka:
Currently the `.take()` function is only available on IndexedParallellIterator, e.g. does not work after a `.filter()`. As shown in the example, this new `.take_unindexed()` function works together with `.filter()`.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -141,6 +141,7 @@ mod reduce;
 mod repeat;
 mod rev;
 mod skip;
+mod skip_any;
 mod splitter;
 mod step_by;
 mod sum;
@@ -186,6 +187,7 @@ pub use self::{
     repeat::{repeat, repeatn, Repeat, RepeatN},
     rev::Rev,
     skip::Skip,
+    skip_any::SkipAny,
     splitter::{split, Split},
     step_by::StepBy,
     take::Take,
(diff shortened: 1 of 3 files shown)


wagnerf42:
hi, does your doctest pass ? because you filter even elements and assert on even and odds.

also, the doctest kind of seems to expect things to be the 5 first but I think your code produces non deterministic results.

yu-re-ka:
Fixed doctest

cuviper:
I think `take_any` would be more consistent with some other adaptors with a similar trade-off, like `find_any`.

Extending the idea, should we also have `skip_any`? And how about some variant of `take_while`/`skip_while`?
(But those ideas don't need to be added to this PR.)

yu-re-ka:
Can you clarify: Should I change the naming to `take_any` and/or add a `skip_any`? I don't really care about either of those, but I would care about getting this functionality into rayon.

cuviper:
Please do rename to `take_any` and `TakeAny`.

If you're not interested in `skip_any`, that's fine, we can do it separately (if at all).

cuviper (commented):
Thanks! The code looks good, but the docs need to explain what these do, how they're different than `take` and `skip`.

cuviper:
I wrote some docs explaining the differences, but your branch is not open to maintainer pushes.
https://github.com/wobcom/rayon/compare/take-unindexed...cuviper:rayon:take-unindexed

It would be great if you could pull that in, or if I don't hear from you I may open a separate PR with everything.

PR #1019, opened 2023-02-15
Add a fallback when threading is unsupported

### LINKED ISSUE #861: A shim mode for targets without threading support
Maybe I haven't looked through issues closely enough, so apologies if this was brought up before.

One thing that comes up often when using Rayon is making sure that code works correctly on targets without threading support (e.g. wasm32-wasi or regular wasm32-unknown-unknown without threads support in general).

Probably the most common way to do this in a way that abstracts away platform differences is to build custom shim for the used `rayon` APIs that uses regular `Iterator` instead and expose it as `mod rayon` that will be included instead of `rayon` crate when corresponding feature is disabled (e.g. [1](https://github.com/shssoichiro/oxipng/blob/master/src/rayon.rs), [2](https://github.com/xiph/rav1e/blob/0f057ed00f7923e637c83eeb76c71627623afa23/src/lib.rs#L89), [3](https://github.com/Renmusxd/RustQIP/blob/dec77d3f82bc2ebaeef2a01fb01150e498f8d8be/src/rayon_helper/mod.rs#L22-L38), [4](https://github.com/ImageOptim/libimagequant/blob/main/src/rayoff.rs)).

However, Rayon itself feels like a lot more appropriate place for such shimming, since it has more control over APIs and would be easier to maintain than custom shims in each separate app that wants to build Rayon-dependant code in a cross-platform manner.

So, the question is: how realistic would it be to make Rayon use regular Iterators when a corresponding Cargo feature is enabled? Given that Rayon can already handle `RAYON_NUM_THREADS=1` case, and the behaviour should be pretty similar, it seems doable but I'm not familiar with internals enough to estimate amount of work this would take.

cuviper:
Fixes #861.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,16 +65,24 @@ jobs:
       - run: cargo test --verbose --package rayon
       - run: cargo test --verbose --package rayon-core
 
-  # wasm won't actually work without threading, but it builds
+  # wasm32-unknown-unknown builds, and even has the runtime fallback for
+  # unsupported threading, but we don't have an environment to execute in.
+  # wasm32-wasi can test the fallback by running in wasmtime.
   wasm:
     name: WebAssembly
     runs-on: ubuntu-latest
+    env:
+      CARGO_TARGET_WASM32_WASI_RUNNER: /home/runner/.wasmtime/bin/wasmtime
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
         with:
-          target: wasm32-unknown-unknown
+          targets: wasm32-unknown-unknown,wasm32-wasi
       - run: cargo check --verbose --target wasm32-unknown-unknown
+      - run: cargo check --verbose --target wasm32-wasi
+      - run: curl https://wasmtime.dev/install.sh -sSf | bash
+      - run: cargo test --verbose --target wasm32-wasi --package rayon
+      - run: cargo test --verbose --target wasm32-wasi --package rayon-core
 
   fmt:
     name: Format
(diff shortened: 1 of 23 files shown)


kornelski:
Could you remove the `compile_error!` on  wasm32-unknown-unknown? 

Fallback to simple single-threaded execution without JS shims or experimental browser features is my primary use-case for the shim.

> error: the wasm32-unknown-unknown target is not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
>    --> ~/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.3/src/lib.rs:219:9
>     |
> 219 | /         compile_error!("the wasm32-unknown-unknown target is not supported by \
> 220 | |                         default, you may need to enable the \"js\" feature. \
> 221 | |                         For more information see: \
> 222 | |                         https://docs.rs/getrandom/#webassembly-support");
>     | |________________________________________________________________________^

cuviper:
@kornelski we only have `getrandom` as a dev-dependency via `rand`, but that shouldn't affect rayon users. We even have a CI target to check `wasm32-unknown-unknown` that's been there for a few years!

cuviper:
bors try

emilk:
Is there an ETA for when we will have a new `rayon` release with this feature? 🙏

cuviper:
I've queued up #1031 for tomorrow, if everything looks good.

cuviper:
In case anyone else runs across this -- the `rayon-wasm` and `rayon-core-wasm` crates are not official. They appear to be snapshots from this repo, meant to include this PR for wasm use. I did confirm that there were no changes in that code besides the crate metadata for new names and versions.

Officially, we released this in `rayon 1.7.0` and `rayon-core 1.11.0`, so there's no need for such a snapshot. I did try to reach out to @aaronyee-eth to see if they would consider yanking theirs to avoid confusion, but I haven't heard back.

PR #1033, opened 2023-03-09
Use cargo's sparse protocol

cuviper:
Cargo 1.68.0 added a "sparse" protocol, which we can opt into with an environment value.

I also added explicit "git" protocol setting to our MSRV testing -- 1.59 doesn't appear to read it anyway, but it may become relevant once the MSRV is bumped higher.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,11 +5,16 @@ on:
       - staging
       - trying
 
+env:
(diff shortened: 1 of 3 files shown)


cuviper:
I suppose by the time we bump MSRV again, this will be the default on stable anyway, and we can just revert the explicit choice. But for now, this did shave a whole minute off the CI time, nice!

bors r+

PR #1034, opened 2023-03-10
TSP Readme; add not for dj10 source

Horki:
Nothing special; just added a single note into tsp README


cuviper:
Sure, thanks!

bors r+

Issue #393, opened 2017-07-02
Prefix scans


jaupe:
I have a use-case to do prefix scans to gather cumulative totals/products of various quantitative measures. This is how I envisage how to do prefix scans from an API perspective:

'''
// prefix sum
let seq = vec![1i64, 2, 3, 4, 5];
let sums = seq.par_iter().sums(); 
assert_eq!(sums vec![1, 3, 6, 10, 15]);

// prefix product
let seq = vec![1i64, 2, 3, 4, 5];
let prods= seq.par_iter().products(); 
assert_eq!(prods, vec![1, 2, 6, 24, 120]);
'''

Would you like a PR for this or should I implement it as a separate crate?

cuviper:
Is that API on any `ParallelIterator`, or only `IndexedParallelIterator`? Does it always return `Vec`, or is it more general like `collect()`?

I'd also like to hear how you would parallelize this. It seems inherently sequential, although we had some ideas for `scan()` (#329), which is a more general case.

jaupe:
Hey @cuviper,

Looking at the Rayon API, I think it would good to make it as generic as possible; so ideally all iterators and using `collect()`.

The algorithm in a pseudo-vector-algebra notation:

1. Break the data into sub-slices and each sub-slice is processed in parallel.
```
[1,2,3,4,5,6,7,8,9,10]= [1,2,3],[4,5,6],[7,8,9],[10]
```
2. A prefix scan is processed on each sub-slice and waits for all prefix scans to be completed.
```
[1,2,3],[4,5,6],[7,8,9],[10]=> [1,3,6],[4,9,15],[7,15,24],[10]
```
3. Once all scans have been processed, sequentially process each sub-slice by adding a offset to each slice. Adding the offset can be done in parallel.
```
[1,3,6],[4,9,15],[7,15,24],[10]  
=> [1,3,6],6+[4,9,15],21+[7,15,24],45+[10] 
=> [1,3,6],[10,15,21],[28 36 45],[55] 
```

cbarrick:
> Adding the offset can be done in parallel

Is this true? It seems to me that you have to compute the final result for one group before you know what offset to add to the next. If so, you're back to the sequential algorithm.

For example, in this step:

    [1,3,6],[4,9,15],[7,15,24],[10]  
    => [1,3,6],6+[4,9,15],21+[7,15,24],45+[10] 
    => [1,3,6],[10,15,21],[28,36,45],[55] 

[...]

Issue #329, opened 2017-05-01
Add an Iterator::scan equivalent


cuviper:
```rust
fn scan<St, B, F>(self, initial_state: St, f: F) -> Scan<Self, St, F>
    where F: FnMut(&mut St, Self::Item) -> Option<B>
{ ... }
```

@nikomatsakis's notes from the [wiki](https://github.com/nikomatsakis/rayon/wiki/Iterator-Trait-Parity):

- this is generally possible (it will be a bit different; more like reduce), but it is non-trivial
- it requires >1 pass and hence produces an intermediate collection
- basically you would do a parallel fold operation, performing a sequential scan for each segment
  - the result of your scan is a VecTree<Vec<T>> of same total length as your input iterator
  - this gives you partial sums (to consider one concrete application) at each point
- you can then traverse the tree (sequentially) to compute, for each vector, the total sum of preceding vectors
- you can then re-traverse the tree (in parallel) and add this total sum to each element of the vector (in place)
- that last step can itself drive another consumer, so this can still be a parallel adapter
  - if we are clever, it can even be an exact one, though it will require a nice VecTree implementation

martinhath:
I'd like to try this out, but I'm having a hard time understanding the notes from the wiki.

> basically you would do a parallel fold operation, performing a sequential scan for each segment

Ok, so the iterator will be `split` up to a point, and then we `scan` each part individually (more or less by using `std::iter::iterator::scan`), but in parallel? If so, how do we handle the `state`, as it is `&mut`? Or do we assume the state is `Copy`, and copy it for each sequential scan such that all segments starts with a fresh state? The alternative (the one I see, anyways) is to just do everything sequentially, but what do we gain by that?

>    the result of your scan is a VecTree<Vec> of same total length as your input iterator

I'm not sure what this means. VecTree<Vec>? Is this just the "split tree", in which the nodes are the segments as `Vec`s?

The last three points are also unclear, but I don't want to flat out take a guess at what is meant, as it might be clear if I can understand the above 😄

cuviper:
I also asked @nikomatsakis to clarify, because I don't really get it either. 😄 

I can answer about `VecTree`: that's a hypothetical tree-of-vectors that we had thought about for general purpose `collect` -- basically folding into a vector locally, then reducing those into a tree that maintains structure.  But we realized that the structure doesn't actually matter, only the order, so today the `collect` implementations that need it just use `LinkedList<Vec<T>>`.

[...]

PR #1036, opened 2023-03-16
[WIP] Create Scan parallel iterator

arieluy:
Hi, here's a draft of a parallel scan function based on discussion in #329 and #393.

### General Approach

1. Use a consumer to split the iterator into pieces, and perform sequential scan on each, accumulating into a `Vec<T>`. The reduce step combines them into a `LinkedList<Vec<T>>`.
2. Sequentially, compute the "offset" we need to add to each section, by looking at the last element of each Vec (and doing another scan). Also, collect the linked list into a `Vec<Vec<T>>` so we can operate on it in parallel again. 
3. Create an unindexed producer which adds the offsets to each Vec in parallel. 

In order to maximize the parallelism, we don't want to split too much, since that creates more sequential work. I've been using `.with_min_len()` in order to prevent that.

Happy to take feedback on better ways to do any of these steps.

### Benchmarks

I also tried writing some benchmarks. I ran these on my M2 Mac with 12 threads.
- For a regular prefix sum or product on ints, the parallel overhead is too much to see any improvement with the parallel version. 
- Only sufficiently slow operations will benefit, but they can see a fairly large improvement. I tried testing this by writing an addition function that waits a certain amount of time before returning. With a 100000 size input:
    - With a delay time of 100ns, the parallel speedup is 2.99
    - With a delay time of 10000ns, the parallel speedup is 5.27
- A slightly more realistic example is matrix multiplication. Here's the execution time on various numbers of threads, compared to the sequential baseline.
<p align="center">matrix scan</p>

### Potential Improvements
- Write a specialized version for prefix sums/products
- Output an indexed producer instead of unindexed. This is definitely harder, but I have a few ideas for how it might be doable. Is this important?
- Testing on higher core machines
- General cleanup + docs

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -81,6 +81,7 @@
 
 use self::plumbing::*;
 use self::private::Try;
+use self::scan::Scan;
 pub use either::Either;
 use std::cmp::{self, Ordering};
 use std::iter::{Product, Sum};
@@ -157,6 +158,8 @@ mod update;
 mod while_some;
 mod zip;
 mod zip_eq;
+
+mod scan;
 
 pub use self::{
     chain::Chain,
@@ -1382,6 +1385,14 @@ pub trait ParallelIterator: Sized + Send {
         S: Send + Sum<Self::Item> + Sum<S>,
     {
         sum::sum(self)
+    }
+
+    fn scan<F>(self, scan_op: F, id: Self::Item) -> Scan<Self::Item, F>
+    where
+        F: Fn(&Self::Item, &Self::Item) -> Self::Item + Sync + Send,
+        <Self as ParallelIterator>::Item: Send + Sync,
+    {
+        scan::scan(self, scan_op, id)
     }
 
     /// Multiplies all the items in the iterator.
--- a/rayon-demo/Cargo.toml
+++ b/rayon-demo/Cargo.toml
@@ -16,6 +16,7 @@ once_cell = "1.17.1"
 rand = "0.8"
 rand_xorshift = "0.3"
 regex = "1"
+ndarray = "0.15.6"
 
 [dependencies.serde]
 version = "1.0.85"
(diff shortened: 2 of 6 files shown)


src/iter/mod.rs
```diff
@@ -1384,6 +1387,14 @@ pub trait ParallelIterator: Sized + Send {
         sum::sum(self)
     }
 
+    fn scan<F>(self, scan_op: F, id: Self::Item) -> Scan<Self::Item, F>
```
cuviper, line 1390: Let's start here, documenting at the API level, especially for a user who is looking for an equivalent to `Iterator::scan`. How would you describe what this does, and importantly how is it different than the sequential version? The signature of `F` is quite different, looking more like some flavor of `reduce`.

arieluy: Good question, I wasn't exactly sure what the signature should be, and it could change. But it's similar to the way `reduce` compares to `fold`. This should have an identical result to sequential scan when the operation doesn't have internal state, and is associative. If not, it won't make sense to run it in parallel.

arieluy: Hi @cuviper, could you take another look at this?

cuviper: I don't feel you addressed my first comment? The signature is part of it, but please make an attempt at adding documentation describing what this does. And frankly, I don't think many people even use `Iterator::scan`, so it's not sufficient to just reference that. Suppose this existed independently - describe what it does, and what points are important for the user to think about.

arieluy: Sorry, I misunderstood you last time. I've added documentation.

I was hoping to get some feedback on the signature and general approach. These are the differences from the other ParallelIterator functions, and my rationale for them:
1. The scan operation currently has type `Fn(&Item, &Item) -> Item`, since when we call it iteratively, we need to keep each intermediate result and not consume it. 
2.  Since `scan_op` takes in references, we don't need multiple copies of `identity`. So, I have `identity` as type `Item`, rather than `Fn() -> Item`, since it's simpler. The other functions use `Fn() -> Item`, though, and it would be easy to change. 

Do you have any preferences on those?

arieluy:
I've rebased and made a few minor updates that had made it into my crate [rayon-scan](https://crates.io/crates/rayon-scan).

PR #1042, opened 2023-04-22
mark spawn function unsafe

kitcatier:
https://github.com/rayon-rs/rayon/blob/3ffed8e7e2d57ed9b69bcb3b50e32fdc2265bf9b/rayon-core/src/spawn/mod.rs#L60-L66
https://github.com/rayon-rs/rayon/blob/3ffed8e7e2d57ed9b69bcb3b50e32fdc2265bf9b/rayon-core/src/spawn/mod.rs#L130-L136
hello, if a function's entire body is unsafe, the function is itself unsafe and should be marked appropriately. Marking them unsafe also means that callers must make sure they know what they're doing.

### DIFF
--- a/rayon-core/src/broadcast/test.rs
+++ b/rayon-core/src/broadcast/test.rs
@@ -251,7 +251,7 @@ fn broadcast_after_spawn() {
 
     // Queue a regular spawn on a thread-local deque.
     crate::registry::in_worker(move |_, _| {
-        crate::spawn(move || tx.send(22).unwrap());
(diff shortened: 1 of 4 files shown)


nikomatsakis:
> if a function's entire body is unsafe, the function is itself unsafe and should be marked appropriately. 

This isn't correct. The `unsafe` in the body indicates that the code asserts that the call is safe, and the comment explains why. The function only needs to be marked unsafe if there is something extra that the caller needs to do to assert safety. In this case, there isn't.

PR #1043, opened 2023-05-04
clarify collect/unzip_into_vec

TheIronBorn:
"cleared" is more precise than "truncated" as Rust's truncate may be partial


cuviper:
Sounds good, thanks!

bors r+

Issue #688, opened 2019-09-03
Consider adding an easier way of synchronously shutting down a thread-pool.


emilio:
AFAICT `ThreadPool::drop` is async, so there's no easy way to ensure that all threads have shot down.

Ideally, I'd like the guarantee of all threads having shot down, including running TLS destructors, by the time `ThreadPool::drop` (or whatever API this would be) returns. This is so I can do deterministic leak-checking.

I think you can do something using `exit_handler` plus `std::sync::Barrier` or something like that, but without the TLS-destructors-ran guarantee. That'd only guarantee that all pre-existing jobs have finished.

cc @cuviper

cuviper:
An existing option that should even solve your TLS is [`build_scoped`](https://docs.rs/rayon/1.2.0/rayon/struct.ThreadPoolBuilder.html#method.build_scoped), like:

```rust
ThreadPoolBuilder::new().build_scoped(
    |thread| thread.run(), // do other setup before running if you like
    |pool| pool.install(work),
)?;
// threads will be totally gone here
```

I think there's also room for something like `ThreadPool::wait(self)`, but that may be complicated by the fact that we didn't require join handles from the custom `spawn_handler`. :/

emilio:
Right, in this case I want a global thread pool that on shutdown I can wait on and drop, though, so I think build_scoped doesn't work.

cuviper:
> in this case I want a global thread pool that on shutdown I can wait on and drop

Ah, that's a harder one, and now I remember where I was playing with this before -- #483.

gsquire:
I was working through a similar issue of needing to await all the threads in a `ThreadPool` recently and Google pointed me to my own [comment](https://github.com/rayon-rs/rayon/issues/650#issuecomment-479260717). 😄

Would using a `WaitGroup` help with your needs as well? I added a call to `wait` inside of my type's `Drop` implementation to await all tasks that were ran via `spawn`. I can submit a patch if this is something that would be beneficial for all users.

MarijnS95:
Is this something that is still looked into?

My use-case is a `rayon::ThreadPool` to  with a bunch of asynchronous `loop {}` workers spawned via `spawn_broadcast()` processing work received from an `mpsc`/`crossbeam` `channel`. Their output is a file on disk, so there's no synchronization back to the caller (i.e. no `broadcast()` which would block and yield the results from every thread).

[...]

PR #1063, opened 2023-06-25
core: Introduce ThreadPoolBuilder::use_current_thread.

emilio:
This generalizes the approach used by targets that don't support threading like
wasm, allowing the builder thread to be part of a new thread-pool.

This PR:

 * Builds on top of the PoC implementation from that issue.
 * Renames the API as per the comments there.
 * Adds a way to clean up the WorkerThread storage once the pool is dropped.
 * Documents and tests the APIs.

Feedback welcome. `clean_up_use_current_thread` is not a great name, but I
think it's descriptive, and maybe good enough given it's a rather niche API for
non-global pools?

### DIFF
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -147,6 +147,7 @@ pub struct ThreadPoolBuildError {
 #[derive(Debug)]
 enum ErrorKind {
     GlobalPoolAlreadyInitialized,
+    CurrentThreadAlreadyInPool,
     IOError(io::Error),
 }
 
@@ -543,11 +544,6 @@ impl<S> ThreadPoolBuilder<S> {
     /// the thread-pool will generally not be picked up automatically by this thread unless you
     /// yield to rayon in some way, like via [`yield_now()`], [`yield_local()`], or [`scope()`].
     ///
-    /// # Panics
-    ///
-    /// This function won't panic itself, but [`ThreadPoolBuilder::build()`] will panic if you've
-    /// called this function and the current thread is already part of another [`ThreadPool`].
-    ///
     /// # Local thread-pools
     ///
     /// Using this in a local thread-pool means the registry will be leaked. In future versions
@@ -759,18 +755,22 @@ impl ThreadPoolBuildError {
 const GLOBAL_POOL_ALREADY_INITIALIZED: &str =
     "The global thread pool has already been initialized.";
 
+const CURRENT_THREAD_ALREADY_IN_POOL: &str =
+    "The current thread is already part of another thread pool.";
+
 impl Error for ThreadPoolBuildError {
     #[allow(deprecated)]
     fn description(&self) -> &str {
         match self.kind {
             ErrorKind::GlobalPoolAlreadyInitialized => GLOBAL_POOL_ALREADY_INITIALIZED,
+            ErrorKind::CurrentThreadAlreadyInPool => CURRENT_THREAD_ALREADY_IN_POOL,
             ErrorKind::IOError(ref e) => e.description(),
         }
     }
 
     fn source(&self) -> Option<&(dyn Error + 'static)> {
         match &self.kind {
-            ErrorKind::GlobalPoolAlreadyInitialized => None,
+            ErrorKind::GlobalPoolAlreadyInitialized | ErrorKind::CurrentThreadAlreadyInPool => None,
             ErrorKind::IOError(e) => Some(e),
         }
     }
@@ -779,6 +779,7 @@ impl Error for ThreadPoolBuildError {
 impl fmt::Display for ThreadPoolBuildError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match &self.kind {
+            ErrorKind::CurrentThreadAlreadyInPool => CURRENT_THREAD_ALREADY_IN_POOL.fmt(f),
             ErrorKind::GlobalPoolAlreadyInitialized => GLOBAL_POOL_ALREADY_INITIALIZED.fmt(f),
             ErrorKind::IOError(e) => e.fmt(f),
         }
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -535,6 +535,23 @@ impl<S> ThreadPoolBuilder<S> {
     }
 
     /// Use the current thread as one of the threads in the pool.
+    ///
+    /// The current thread is guaranteed to be at index 0, and since the thread is not managed by
+    /// rayon, the spawn and exit handlers do not run for that thread.
+    ///
+    /// Note that the current thread won't run the main work-stealing loop, so jobs spawned into
+    /// the thread-pool will generally not be picked up automatically by this thread unless you
+    /// yield to rayon in some way, like via [`yield_now()`], [`yield_local()`], or [`scope()`].
+    ///
+    /// # Panics
+    ///
+    /// This function won't panic itself, but [`ThreadPoolBuilder::build()`] will panic if you've
+    /// called this function and the current thread is already part of another [`ThreadPool`].
+    ///
+    /// # Local thread-pools
+    ///
+    /// Using this in a local thread-pool means the registry will be leaked. In future versions
+    /// there might be a way of cleaning up the current-thread state.
     pub fn use_current_thread(mut self) -> Self {
         self.use_current_thread = true;
         self
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -175,7 +175,7 @@ pub struct ThreadPoolBuilder<S = DefaultSpawn> {
     num_threads: usize,
 
     /// The thread we're building *from* will also be part of the pool.
-    use_current: bool,
+    use_current_thread: bool,
 
     /// Custom closure, if any, to handle a panic that we cannot propagate
     /// anywhere else.
@@ -230,7 +230,7 @@ impl Default for ThreadPoolBuilder {
     fn default() -> Self {
         ThreadPoolBuilder {
             num_threads: 0,
-            use_current: false,
+            use_current_thread: false,
             panic_handler: None,
             get_thread_name: None,
             stack_size: None,
@@ -441,7 +441,7 @@ impl<S> ThreadPoolBuilder<S> {
             spawn_handler: CustomSpawn::new(spawn),
             // ..self
             num_threads: self.num_threads,
-            use_current: self.use_current,
+            use_current_thread: self.use_current_thread,
             panic_handler: self.panic_handler,
             get_thread_name: self.get_thread_name,
             stack_size: self.stack_size,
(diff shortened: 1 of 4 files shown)


rayon-core/src/thread_pool/mod.rs
```diff
@@ -461,6 +461,39 @@ pub fn yield_local() -> Option<Yield> {
...
+/// Calling this function from a thread pool job will block indefinitely.
+///
+/// Calling this function before before the thread-pool has been dropped will cause the thread to
+/// not return control flow to the caller until that happens (stealing work as necessary).
+///
+/// # Panics
+///
+/// If the calling thread is no the creator thread of a thread-pool, or not part of that
+/// thread-pool, via [`ThreadPoolBuilder::use_current_thread()`].
+pub fn clean_up_use_current_thread() {
```
adamreichold, line 477: Concerning the name: How about changing the verb to avoid doubling it, e.g. `release_current_thread`, `drop_current_thread`?

More generally, could this become part of `impl Drop for ThreadPool` (automatically detecting if the `use_current_thread` was used) as the thread pool created by `build_global` would never be dropped anyway?

emilio: > Concerning the name: How about changing the verb to avoid doubling it, e.g. `release_current_thread`, `drop_current_thread`?

I'm fine with either. I kinda prefer `use_current_thread` or similar being in the name, because this is specific to that API. Maybe `stop_using_current_thread`?

> More generally, could this become part of `impl Drop for ThreadPool` (automatically detecting if the `use_current_thread` was used) as the thread pool created by `build_global` would never be dropped anyway?

Not really, because:

 * `Drop for ThreadPool` has no guarantee of running on the same thread as `ThreadPoolBuilder::build`.
 * The external thread is not guaranteed to ever yield to Rayon's pool.
 * `ThreadPool`'s drop impl doesn't guarantee that all of the threads have exited anyways (because it might be running from a worker).

emilio:
Fixed the formatting of the tests, whoops.

emilio:
@cuviper review ping? Modulo the naming issue discussed above (happy to tweak as needed) I think this should be in decent shape?

emilio:
Also ideas to be able to test the error (deadlocking) case of calling `clean_up_use_current_thread` from the inner job would be great, but probably not a blocker.

cuviper:
Hi, sorry -- I'm going to be unavailable the rest of this week, but I'll try to get to this early next week.

emilio:
No worries, I was just making sure this didn't fall through the cracks, thanks! (And enjoy your time off if that's the reason for your availability :)).

emilio:
Gentle ping?

rayon-core/src/thread_pool/mod.rs
cuviper: Maybe it could be rolled into a `ThreadPool::exit(self)` method, somewhat like #688? So if the "used" thread is the one to call that, we can also handle the extra cleanup. If another thread takes the `ThreadPool` and calls it, it's still *safe* to leak that one.

(Though back then you wanted that for the global pool, and also thread destructors are still hard to promise.)

rayon-core/src/lib.rs
```diff
@@ -532,6 +537,34 @@ impl<S> ThreadPoolBuilder<S> {
         self
     }
 
+    /// Use the current thread as one of the threads in the pool.
+    ///
+    /// The current thread is guaranteed to be at index 0, and given the thread is not managed by
+    /// rayon, so spawn and exit handlers do not run for that thread.
```
cuviper, line 543:
```suggestion
    /// The current thread is guaranteed to be at index 0, and since the thread is not managed by
    /// rayon, the spawn and exit handlers do not run for that thread.
```

rayon-core/src/lib.rs
```diff
@@ -532,6 +537,34 @@ impl<S> ThreadPoolBuilder<S> {
...
+    ///
+    /// Note that the current thread won't run the main work-stealing loop, so jobs spawned into
+    /// the thread-pool will generally not be picked up automatically by this thread unless you
+    /// yield to rayon in some way, like via [`yield_now()`], [`yield_local()`], [`scope()`], or
+    /// [`clean_up_use_current_thread()`].
+    ///
+    /// # Panics
+    ///
+    /// This function won't panic itself, but [`ThreadPoolBuilder::build()`] will panic if you've
+    /// called this function and the current thread is already part of another [`ThreadPool`].
```
cuviper, line 553: Maybe we _should_ panic opportunistically? I guess you could call this and then never build, but what's the point of that...

An error might be appropriate too, but that doesn't fit the builder pattern well.

emilio: I'm not a fan of this, both because it complicates the code and because maybe you create a builder in one thread but call `build()` in another? Though I haven't checked if `ThreadPoolBuilder` is `Send`.

rayon-core/src/lib.rs
emilio: Done, thanks! Also rebased :)

rayon-core/src/lib.rs
cuviper: It's not `Send` because of unconstrained `dyn Fn` fields, like `get_thread_name`.

emilio: Alright, let me know if you want me to panic before-hand then (and add a static assert that ThreadPoolBuilder isn't Send if you want).

But given we also need the check in build(), I don't think it's worth it necessarily...

emilio: @cuviper ping about the above? Happy to make this change (or any other) if you want.

cuviper: Eh, this is ok. The cleanup strategy is the other lingering question.

rayon-core/src/thread_pool/mod.rs
emilio: I'm ok rolling `clean_up_use_current_thread` into a `ThreadPool::exit` or probably `impl Drop for ThreadPool`?

emilio: I sent a follow-up patch doing that [here](https://github.com/rayon-rs/rayon/pull/1063/commits/a6653df1f974eaa8b8a7d99e15763516ef58d59a), let me know if that looks good, it's definitely simpler.

If somebody needs to clean-up dropping the thread-pool in a different thread, sending a patch re-introducing the clean-up API seems fine.

emilio: (Now with CI passing)

rayon-core/src/thread_pool/mod.rs
```diff
@@ -368,6 +368,7 @@ impl ThreadPool {
 impl Drop for ThreadPool {
     fn drop(&mut self) {
         self.registry.terminate();
+        self.registry.clean_up_creator_thread_if_needed();
```
cuviper, line 371: This makes me nervous because it's very possible to smuggle the `ThreadPool` into a job running on the pool, especially since `use_current_pool()` puts us in that pool's context where we can refer to it implicitly. Then a drop can destroy the `WorkerThread` while there's still active references on the call stack above us.

```rust
fn main() {
    let pool = rayon::ThreadPoolBuilder::new()
        .use_current_thread()
        .build()
        .unwrap();
        
    rayon::scope(|_| drop(pool));
}
```

`scope` calls `registry::in_worker`, which calls back with a `&WorkerThread`, which will be used to wait for spawns and clean up after the closure returns -- after we've dropped it.

If we can detect this situation, then I guess we can just leak the "current" thread.

cuviper: (Miri and valgrind both take issue with that example.)

emilio: I pushed 8c840bd5113a2a475bc5a8aa40e07693f9da7e2d to deal with that, but I'd also be happy to just removing the clean-up for now (that is, always leak) and moving those bits to a separate PR if that makes it easier to look at in isolation?

I'm a bit on the fence on whether we should try to deal with the edge cases of broken clean-ups. Exposing `clean_up_use_current_thread` as an `unsafe` function, making it the caller's responsibility to clean up at the right time would be ok for me, I believe. Though I doubt `in_local_op` would be a performance issue in practice, it makes me a bit sad having to track that state just for this.

rayon-core/src/registry.rs
```diff
@@ -924,6 +950,21 @@ impl WorkerThread {
...
         }
     }
+
+    fn execute_local_op<OP, R>(&self, op: OP) -> R
+    where
+        OP: FnOnce(&WorkerThread, bool) -> R + Send,
+        R: Send,
+    {
+        let old_in_local_op = self.in_local_op.get();
+        self.in_local_op.set(true);
```
cuviper, line 960:
```suggestion
        let old_in_local_op = self.in_local_op.replace(true);
```

rayon-core/src/thread_pool/mod.rs
cuviper: Hmm, `in_worker` isn't sufficient, because `rayon::in_place_scope(|s| s.spawn(|_| drop(pool)));` also hits use-after-free. Or for a different flavor, `rayon::spawn(|| dbg!(drop(pool))); rayon::yield_now();` deadlocks waiting for the thread to terminate, because the `spawn` has to complete to deprecate its count.

So yeah, I'm not feeling confident about closing all the safe loopholes -- unsafe cleanup is probably the way to go, as distasteful as that seems. Or as you say, just leaking for now is ok, and still safe in Rust terms.

adamreichold: > So yeah, I'm not feeling confident about closing all the safe loopholes -- unsafe cleanup is probably the way to go, as distasteful as that seems. Or as you say, just leaking for now is ok, and still safe in Rust terms.

I think leaking would then be preferable as an intermediate step to avoid committing Rayon to an unsafe API which has to be supported until the next major version bump.

emilio: Alright. I updated the PR as per the discussion here, with the documentation that this leaks for local thread-pools.

I'll send a separate PR for the unsafe clean-up function but that doesn't need to block on this.

emilio:
The `PR / Format` failure doesn't seem related (temporary github actions hiccup?)

rayon-core/src/lib.rs
```diff
@@ -525,6 +530,29 @@ impl<S> ThreadPoolBuilder<S> {
...
+    /// rayon, the spawn and exit handlers do not run for that thread.
+    ///
+    /// Note that the current thread won't run the main work-stealing loop, so jobs spawned into
+    /// the thread-pool will generally not be picked up automatically by this thread unless you
+    /// yield to rayon in some way, like via [`yield_now()`], [`yield_local()`], or [`scope()`].
+    ///
+    /// # Panics
+    ///
+    /// This function won't panic itself, but [`ThreadPoolBuilder::build()`] will panic if you've
+    /// called this function and the current thread is already part of another [`ThreadPool`].
```
cuviper, line 545: Actually, the build methods return a `Result`, so why should this panic instead of just an `Err`?

emilio: Yeah, good call. Fixed and added a test.

cuviper:
OK, I think we're good to go -- thanks for your persistence! :)

bors r+

emilio:
> OK, I think we're good to go -- thanks for your persistence! :)

Thank you!

PR #1065, opened 2023-06-25
Fix #1064

### LINKED ISSUE #1064: yield_local can cause stack overflow
When a job calls yield_local, Rayon loads another job onto the stack on top of it. If lots of jobs are calling it, this can cause a stack overflow.

To fix this, Rayon should give each thread a flag that tracks whether yield_local is already on the stack. If it is and is called again, yield_local should return immediately. (There may need to be a third value in the Yield enum for this case.)

Example stack trace:
[stack_trace.txt](https://github.com/rayon-rs/rayon/files/11860812/stack_trace.txt)

Pr0methean:
Fixes #1064 by preventing two jobs on the same thread from yielding at once.

### DIFF
--- a/rayon-core/src/registry.rs
+++ b/rayon-core/src/registry.rs
@@ -16,10 +16,11 @@ use std::hash::Hasher;
 use std::io;
 use std::mem;
 use std::ptr;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, Once};
 use std::thread;
 use std::usize;
+use Ordering::Relaxed;
 
 /// Thread builder used for customization via
 /// [`ThreadPoolBuilder::spawn_handler`](struct.ThreadPoolBuilder.html#method.spawn_handler).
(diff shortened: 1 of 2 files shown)


rayon-core/src/registry.rs
```diff
@@ -857,19 +865,37 @@ impl WorkerThread {
     }
 
     pub(super) fn yield_now(&self) -> Yield {
+        let yielding = &self.registry.thread_infos[self.index].is_yielding;
```
emilio, line 868: Seems this could just be a `Cell<bool>` in `WorkerThread` (since that's where the thread-local stuff generally lives) instead of living on the registry?

rayon-core/src/thread_pool/mod.rs
```diff
@@ -468,4 +468,7 @@ pub enum Yield {
     Executed,
     /// No available work was found.
     Idle,
+    /// Another job that has yielded is already on this thread's stack, so to prevent a stack
+    /// overflow we will not start a third job.
+    Recursive,
```
cuviper, line 473: This is a breaking change, because we didn't make it `#[non_exhaustive]`.

cuviper (changes requested):
I don't agree that this is something we should block, but I'll write more in #1064.

PR #1084, opened 2023-09-01
`add` and `mul` now use a `[T, T]` array instead of chained iter::once

Lucretiel:
Minor update to `sum` and `product` parallel iterators, to have them use arrays instead of chains of iter::once.


cuviper:
Works for me, thanks!

bors r+

PR #1085, opened 2023-09-02
Use a custom Iterator implementation for `enumerate`

Lucretiel:
This PR replaces the current `Zip`-based implementation of `Enumerate` with a custom iterator. This allows us to specialize the common iterator methods (like `fold` and `count`), and to provide a simpler implementation of `DoubleEndedIterator`.

### DIFF
--- a/src/iter/enumerate.rs
+++ b/src/iter/enumerate.rs
@@ -1,8 +1,5 @@
 use super::plumbing::*;
 use super::*;
-use std::iter;
-use std::ops::Range;
-use std::usize;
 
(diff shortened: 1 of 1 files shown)


cuviper:
Do you have any workloads that demonstrate benefit from this? The only benchmark in rayon-demo that uses it is `sieve_parallel`, and I don't see any difference there.

cuviper:
Absent any evidence, I'm inclined to leave this alone. While stuff like `fold` could be worthwhile optimizations, we also lose the kinds of specialization that only the standard library can write, like `TrustedLen`.

PR #1087, opened 2023-09-04
core: registry: Factor out "wait till out of work" part of the main loop.

emilio:
This was originally done for #1063, in order to reuse this to allow cleaning up the TLS data allocated by use_current_thread.

We ended up not using that, but this refactoring seems useful on its own, perhaps.

### DIFF
--- a/rayon-core/src/registry.rs
+++ b/rayon-core/src/registry.rs
@@ -809,6 +809,20 @@ impl WorkerThread {
         mem::forget(abort_guard); // successful execution, do not abort
     }
 
+    unsafe fn wait_until_out_of_work(&self) {
+        debug_assert_eq!(self as *const _, WorkerThread::current());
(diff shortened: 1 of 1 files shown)


emilio:
The check failure looks spurious.

rayon-core/src/registry.rs
```diff
@@ -924,6 +917,20 @@ unsafe fn main_loop(thread: ThreadBuilder) {
     }
 }
 
+unsafe fn wait_until_out_of_work(worker_thread: &WorkerThread) {
```
cuviper, line 920: Could this be a `&self` method instead? (next to the existing `wait_until` methods)

emilio: Yeah, that works and is nicer / more in-line with the other methods indeed.

Issue #1105, opened 2023-11-20
cooperative yield in ThreadPool::install() causes unexpected behavior in nested pools


benkay86:
It appears calls to `ThreadPool::install(op)` will start running `op` inside the child thread pool and then cooperatively yield, _potentially switching to a different task_. This causes unexpected behavior with nested iterators using thread pools.  It would be nice if rayon didn't yield on `ThreadPool::install()`, or if the authors believe that it should yield, to place some kind of warning in the documentation for `ThreadPool::install()`.

Consider the following example.  You have an outer loop that performs some operation that requires a lot of memory.  You have an inner loop that uses a lot of cpu but is memory-efficient.  To avoid unbounded memory growth, you will run the outer and inner loops in separate thread pools where `ncpus_outer` is a small number and `ncpus_inner` is a big number.

```rust
let pool_outer = ThreadPoolBuilder::default().num_threads(ncpus_outer).build().unwrap();
let pool_inner = ThreadPoolBuilder::default().num_threads(ncpus_inner).build().unwrap();
pool_outer.install(|| {
    some_par_iter.for_each({ |foo|
        println!("Entered outer loop.");
        let bar = memory_intensive_op(foo);
        pool_inner.install(|| {
            println!("Entered inner pool.");
            thing.into_par_iter().for_each({ |baz|
                cpu_intensive_op(baz);
            });
        });
        println!("Left inner pool.");
    });
});
```

Suppose `ncpus_outer` is equal to one.  You might reasonably expect that only one `bar` will be allocated at any given time, giving you control over memory use.  However, the output of the program may actually look like:

```
Entered outer loop.
Entered inner pool.
Entered outer loop.
Entered inner pool.
Entered outer loop.
Entered outer loop.
Left inner pool.
...
```

This is possible because, rather than blocking until `pool_inner.install()` returns, the outer thread pool will yield and cooperatively schedule another iteration of the outer loop _on the same thread_. This gives you essentially zero control over how many instances of `bar` will be allocated at a given time.

This is an even bigger problem if you have to do some kind of synchronization (yes there are real uses cases for this):

```rust
let pool_outer = ThreadPoolBuilder::default().num_threads(ncpus_outer).build().unwrap();
let pool_inner = ThreadPoolBuilder::default().num_threads(ncpus_inner).build().unwrap();
let mutex = std::sync::Mutex::new(());
pool_outer.install(|| {
    some_par_iter.for_each({ |foo|
        println!("Entered outer loop.");
[...]

PR #1107, opened 2023-11-29
Document implicit yield in install() per #1105

benkay86:
Add documentation to `rayon::ThreadPool::install()` warning about implicit yield behavior, as discussed in #1105.

### DIFF
--- a/rayon-core/src/thread_pool/mod.rs
+++ b/rayon-core/src/thread_pool/mod.rs
@@ -80,6 +80,43 @@ impl ThreadPool {
     /// thread-local data from the current thread will not be
     /// accessible.
     ///
+    /// # Warning: execution order
+    ///
+    /// If the current thread is part of a different thread pool, it will try to
+    /// keep busy while the `op` completes in its target pool, similar to
+    /// calling [`ThreadPool::yield_now()`] in a loop. Therefore, it may
+    /// potentially schedule other tasks to run on the current thread in the
+    /// meantime. For example
+    ///
+    /// ```rust
+    /// # use rayon_core as rayon;
+    /// fn main() {
+    ///     rayon::ThreadPoolBuilder::new().num_threads(1).build_global().unwrap();
+    ///     let pool = rayon_core::ThreadPoolBuilder::default().build().unwrap();
+    ///     let do_it = || {
+    ///         print!("one ");
+    ///         pool.install(||{});
+    ///         print!("two ");
+    ///     }
+    ///     rayon::join(|| do_it(), || do_it());
+    /// }
+    /// ```
+    ///
+    /// Since we configured just one thread in the global pool, one might
+    /// expect `do_it()` to run sequentially, producing:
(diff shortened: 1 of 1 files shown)


rayon-core/src/thread_pool/mod.rs
```diff
@@ -80,6 +80,41 @@ impl ThreadPool {
     /// thread-local data from the current thread will not be
     /// accessible.
     ///
+    /// # Warning: execution order
+    ///
+    /// `install()` implicitly calls [`ThreadPool::yield_now()`],
+    /// potentially scheduling another task to run on the current thread. For
+    /// example:
```
cuviper, line 87: I would like this to clarify the context a bit more, something like:
```suggestion
    /// If the current thread is part of a different thread pool, it will try to
    /// keep busy while the `op` completes in its target pool, similar to
    /// calling [`ThreadPool::yield_now()`] in a loop. Therefore, it may
    /// potentially schedule other tasks to run on the current thread in the
    /// meantime. For example:
```

Does that make sense to you?

benkay86: Yes, changed to your phrasing via force push.

cuviper:
Thanks!

rayon-core/src/thread_pool/mod.rs
```diff
@@ -80,6 +80,43 @@ impl ThreadPool {
...
+    /// potentially schedule other tasks to run on the current thread in the
+    /// meantime. For example
+    ///
+    /// ```rust
+    /// # use rayon_core as rayon;
+    /// use rayon::iter::{IntoParallelIterator, ParallelIterator};
+    /// fn main() {
+    ///     rayon::ThreadPoolBuilder::new().num_threads(1).build_global().unwrap();
+    ///     let pool = rayon::ThreadPoolBuilder::default().build().unwrap();
+    ///     (0..3).into_par_iter().for_each(|_| {
```
cuviper, line 97: Oh, we can't use APIs from `rayon` proper while testing `rayon-core`. Maybe you can make the example with `join` instead?

benkay86: Foiled!  Rewritten to use `join()` as requested.

rayon-core/src/thread_pool/mod.rs
```diff
@@ -80,6 +80,43 @@ impl ThreadPool {
...
+    /// ```rust
+    /// # use rayon_core as rayon;
+    /// fn main() {
+    ///     rayon::ThreadPoolBuilder::new().num_threads(1).build_global().unwrap();
+    ///     let pool = rayon_core::ThreadPoolBuilder::default().build().unwrap();
+    ///     let do_it = || {
+    ///         print!("one ");
+    ///         pool.install(||{});
+    ///         print!("two ");
+    ///     }
```
cuviper, line 100:
```suggestion
    ///     };
```

PR #1110, opened 2023-12-25
Use wasm-sync to allow usage on the main browser thread

RReverser:
One of the most common complaints I've been receiving in [wasm-bindgen-rayon](https://github.com/RReverser/wasm-bindgen-rayon) that prevents people from using Rayon on the Web is the complexity of manually splitting up the code that uses Rayon into a Web Worker from code that drives the UI.

It requires custom message passing for proxying between two threads (Workers), which, admittedly, feels particularly silly when using a tool that is meant to simplify working with threads for you.

This all stems from a [Wasm limitation](https://github.com/WebAssembly/threads/issues/177) that disallows `atomic.wait` on the main browser thread. In theory, it's a reasonable limitation, since blocking main thread on the web is more problematic than on other platforms as it blocks web app's UI from being responsive altogether, and because there is no limit on how long atomic wait can block. In practice, however, it causes enough issues for users that various toolchains - even Emscripten - work around this issue by spin-locking when on the main thread.

Rust / wasm-bindgen decided not to adopt the same workaround, following the core Wasm decision, which is also a fair stance for general implementation of `Mutex` and other blocking primitives, but I believe Rayon usecase is sufficiently special. Code using parallel iterators is almost always guaranteed to run for less or, worst-case, ~same time as code using regular iterators, so it doesn't make sense to "punish" Rayon users and prevent them from being able to use parallel iterators on the main thread when it will lead to a _more_ responsive UI than using regular iterators.

This PR adds a `cfg`-conditional dependency on [wasm_sync](https://docs.rs/wasm_sync/latest/wasm_sync/) that automatically switches to the allowed spin-based `Mutex` and `Condvar` when it detects it's running on the main thread, and to regular `std::sync` based implementation otherwise, thus avoiding the `atomics.wait` error. This dependency will only be added when building for `wasm32-unknown-unknown` - that is, not affecting WASI and Emscripten users - and only when building with `-C target-feature=+atomics`, so not affecting users who rely on Rayon's single-threaded fallback mode either. I hope this kind of very limited override will be acceptable as it makes it much easier to use Rayon on the web.

When this is merged, I'll be able to leverage it in wasm-bindgen-rayon and [significantly simplify](https://github.com/RReverser/wasm-bindgen-rayon/compare/main...RReverser:wasm-bindgen-rayon:wasm-sync?expand=1) demos, tests and docs by avoiding that extra Worker machinery (e.g. see `demo/wasm-worker.js` and `demo/index.js` merged into single simple JS file in the linked diff).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,6 +71,7 @@ jobs:
         include:
           - toolchain: stable
           - toolchain: nightly
+            cargoflags: --features web_spin_lock
             rustflags: -C target-feature=+atomics,+bulk-memory,+mutable-globals
     env:
       RUSTFLAGS: ${{ matrix.rustflags }}
@@ -80,7 +81,7 @@ jobs:
         with:
           toolchain: ${{ matrix.toolchain }}
           targets: wasm32-unknown-unknown
-      - run: cargo build --verbose --target wasm32-unknown-unknown
+      - run: cargo build --verbose --target wasm32-unknown-unknown ${{ matrix.cargoflags }}
 
   # wasm32-wasi can test the fallback by running in wasmtime.
   wasi:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,7 +63,6 @@ jobs:
 
   # wasm32-unknown-unknown builds, and even has the runtime fallback for
   # unsupported threading, but we don't have an environment to execute in.
-  # wasm32-wasi can test the fallback by running in wasmtime.
   wasm:
     name: WebAssembly (standalone)
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -111,6 +111,6 @@ jobs:
   done:
     name: Complete
     runs-on: ubuntu-latest
-    needs: [check, test, demo, i686, wasm, fmt]
+    needs: [check, test, demo, i686, wasm, wasi, fmt]
     steps:
       - run: exit 0
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,9 +71,7 @@ jobs:
       matrix:
         include:
           - toolchain: stable
-            targets: wasm32-unknown-unknown
           - toolchain: nightly
-            targets: wasm32-unknown-unknown
             rustflags: -C target-feature=+atomics,+bulk-memory,+mutable-globals
     env:
       RUSTFLAGS: ${{ matrix.rustflags }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,7 +65,29 @@ jobs:
   # unsupported threading, but we don't have an environment to execute in.
   # wasm32-wasi can test the fallback by running in wasmtime.
   wasm:
-    name: WebAssembly
+    name: WebAssembly (standalone)
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - toolchain: stable
+            targets: wasm32-unknown-unknown
+          - toolchain: nightly
+            targets: wasm32-unknown-unknown
+            rustflags: -C target-feature=+atomics,+bulk-memory,+mutable-globals
+    env:
+      RUSTFLAGS: ${{ matrix.rustflags }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.toolchain }}
+          targets: wasm32-unknown-unknown
+      - run: cargo build --verbose --target wasm32-unknown-unknown
+
+  # wasm32-wasi can test the fallback by running in wasmtime.
+  wasi:
+    name: WebAssembly (WASI)
     runs-on: ubuntu-latest
     env:
       CARGO_TARGET_WASM32_WASI_RUNNER: /home/runner/.wasmtime/bin/wasmtime
@@ -73,10 +95,9 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
         with:
-          targets: wasm32-unknown-unknown,wasm32-wasi
-      - run: cargo check --verbose --target wasm32-unknown-unknown
-      - run: cargo check --verbose --target wasm32-wasi
+          targets: wasm32-wasi
       - run: curl https://wasmtime.dev/install.sh -sSf | bash
+      - run: cargo build --verbose --target wasm32-wasi
       - run: cargo test --verbose --target wasm32-wasi --package rayon
       - run: cargo test --verbose --target wasm32-wasi --package rayon-core
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,9 @@ exclude = ["ci"]
 [dependencies]
 rayon-core = { version = "1.12.0", path = "rayon-core" }
 
+[target.'cfg(all(target_arch = "wasm32", target_os = "unknown", target_feature = "atomics"))'.dependencies]
+wasm_sync = "0.1.0"
+
 # This is a public dependency!
 [dependencies.either]
 version = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,14 +20,19 @@ exclude = ["ci"]
 
 [dependencies]
 rayon-core = { version = "1.12.0", path = "rayon-core" }
-
-[target.'cfg(all(target_arch = "wasm32", target_os = "unknown", target_feature = "atomics"))'.dependencies]
-wasm_sync = "0.1.0"
+wasm_sync = { version = "0.1.0", optional = true }
 
 # This is a public dependency!
 [dependencies.either]
 version = "1.0"
 default-features = false
+
+[features]
+# This feature switches to a spin-lock implementation on the browser's
+# main thread to avoid the forbidden `atomics.wait`.
+#
+# Only useful on the `wasm32-unknown-unknown` target.
+web_spin_lock = ["wasm_sync", "rayon-core/web_spin_lock"]
 
 [dev-dependencies]
 rand = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ default-features = false
 # main thread to avoid the forbidden `atomics.wait`.
 #
 # Only useful on the `wasm32-unknown-unknown` target.
-web_spin_lock = ["wasm_sync", "rayon-core/web_spin_lock"]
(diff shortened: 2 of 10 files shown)


rayon-core/src/lib.rs
```diff
@@ -103,6 +103,20 @@ pub use self::thread_pool::current_thread_index;
...
+    target_feature = "atomics"
+)))]
+pub use std::sync;
+
+#[cfg(all(
+    target_arch = "wasm32",
+    target_os = "unknown",
+    target_feature = "atomics"
+))]
+pub use wasm_sync as sync;
```
cuviper, line 118: I definitely do not want to put these in the public API. If `rayon` needs to use it too, that should be a dependency on its own.

src/result.rs
```diff
@@ -7,7 +7,7 @@
 
 use crate::iter::plumbing::*;
 use crate::iter::*;
-use std::sync::Mutex;
+use rayon_core::sync::Mutex;
```
cuviper, line 10: This case only ever uses `try_lock` -- is that still a problem for the web main?

cuviper (commented):
What is the current failure mode when it tries to block on the main web thread? Some kind of panic/abort of the wasm module?

RReverser:
> Some kind of panic/abort of the wasm module?

More specifically, Wasm engine itself throws an error on the instruction, so it's not even something one can catch & handle from Wasm.

src/result.rs
RReverser: Ah, no, I think that should be fine.

rayon-core/src/lib.rs
RReverser: Okay. I figured sharing might be easier, but I don't have strong opinion. I'll change when I'm back at computer.

src/result.rs
RReverser: Do you think we could change `iter/par_bridge.rs` to use `try_lock` instead of `lock` as well? It already uses custom loop, so functionally it would be equivalent, but I'm not sure how common it is for `try_lock` to fail there and, thus, whether such spinning would cause performance regression on other platforms.

If we could, then we could avoid dependency on `wasm_sync` in `rayon` altogether and keep it only in `rayon-core`.

RReverser: @cuviper Ping - which route would you prefer - should I change it to `try_lock` or add `wasm_sync` dependency to `rayon` as well?

cuviper: Spinning on `try_lock` would regress #795, so let's not go there.

The `par_bridge` `lock` should normally only be called from within the pool, in which case `wasm_sync` will properly block instead of spinning, right? However, it's also possible to make the main thread part of the pool since #1063. So either we need to let `wasm_sync` spin there, or we need to document that wasm users shouldn't do that when they have real threading available.

Anyway, it's not a big deal to add the dependency to `rayon` when it's already forced by `rayon-core`.

RReverser: Fair enough. Based on this, sounds like adding `wasm_sync` to `rayon` is indeed a better option for cross-platform compat. I'll do that.

RReverser: Done.

cuviper:
Would it be possible to test this in CI?

RReverser:
> Would it be possible to test this in CI?

I am testing this on CI on the wasm-bindgen-rayon side, but I was also wondering if we should move / add CI covering usage with wasm-bindgen-rayon for rayon too, given that you have a more comprehensive test suite...

Running tests requires some plumbing - as in, regular `#[test]` doesn't "just work" with wasm-bindgen, and I wasn't sure if you'd be open to having to maintain that in the main rayon repo, but if it's something you're interested in, I can try and add such plumbing in a follow-up PR.

cuviper:
If runtime in CI is complicated, then maybe just a build test is enough. I suppose that needs `-Zbuild-std` for atomics?

RReverser:
> I suppose that needs `-Zbuild-std` for atomics?

Just for a build test - I think not necessarily, enabling target-feature for Rayon itself (via `-C target-features`) should be enough. Then the standard library won't work properly in multithreaded environment (e.g. things like allocator won't be thread-safe), but as long as you only build and don't try to run it, it should be enough.

RReverser:
I'll try to add such basic check.

RReverser:
Heh made a mistake of trying to add `wasm32-wasi-preview1-threads` to CI too while at it. Big mistake - apparently that target is not quite there yet, as failing with random failures like unaligned atomic access or accessing memory out of Wasm bounds.

Will stick to the original scope instead.

RReverser:
@cuviper Done. I split out WASI CI from Wasm CI since they do different things, and because I need `nightly` only for `wasm32-unknown-unknown` (otherwise `target-features` is ignored).

Btw, I'm curious - why are Wasm sections set to only run on `merge_group` workflow, but not on PR workflows? Seems like there's a risk that a PR could break them w/o noticing? For example, those workflows don't run on this PR - I had to create a separate branch to trigger CI on my own fork to verify that they actually work (https://github.com/RReverser/rayon/actions/runs/7504219725).

cuviper:
> Btw, I'm curious - why are Wasm sections set to only run on `merge_group` workflow, but not on PR workflows? Seems like there's a risk that a PR could break them w/o noticing?

The idea was to keep the PR CI fast like rust-lang/rust does, since most changes are not target specific, and the macos/windows builders are noticeably slower. The merge queue is supposed to make sure that we *do* get a full run so we'll notice before any target-specific break actually reaches the repo.

But these days, maybe those slower builders are still fast enough that we don't need to bother splitting...

.github/workflows/ci.yaml
```diff
@@ -65,18 +65,37 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.toolchain }}
+          targets: wasm32-unknown-unknown
+      - run: cargo build --verbose --target wasm32-unknown-unknown
+
+  # wasm32-wasi can test the fallback by running in wasmtime.
+  wasi:
```
cuviper, line 87: This new group should be added to `needs` in the `done` job, as that's what the merge queue's branch protection watches for.

.github/workflows/ci.yaml
```diff
@@ -65,18 +65,37 @@ jobs:
   # unsupported threading, but we don't have an environment to execute in.
   # wasm32-wasi can test the fallback by running in wasmtime.
```
cuviper, line 66:
```suggestion
```

RReverser: Oops, good catch.

.github/workflows/ci.yaml
RReverser: Oh, right. Done.

RReverser:
Hm one more thought: I wonder if it would be better to require an explicit Cargo feature for this, rather than assuming that all `wasm32-unknown-unknown` users will use it with wasm-bindgen. For example, `getrandom` does this with `js` feature and some other crates use `wasm-bindgen` feature.

It would be a bit more verbose, but perhaps explicit opt-in is better.

RReverser:
> For example, `getrandom` does this with `js` feature and some other crates use `wasm-bindgen` feature.
> 
> It would be a bit more verbose, but perhaps explicit opt-in is better.

OTOH that doesn't really enable JS support like in other crates - you still need wasm-bindgen-rayon as well. I guess it should be called `browser_lock` or something to clarify that it only takes care of locking and only useful in browsers.

cuviper:
Maybe just leave it directly as `wasm-sync` like regular optional dependencies? Or do you think there will be other related functionality that may want to piggy-back on this?

RReverser:
> Or do you think there will be other related functionality that may want to piggy-back on this?

Not sure about this one yet, but:

1. I wondered if we might need to switch from `wasm_sync` to another implementation in the future, in which case using its name directly might not be future-proof.
2. More importantly, `rayon`'s `wasm_sync` would need to enable `rayon-core`'s `wasm_sync`. This is possible with regular Cargo features, but AFAIK it's not possible to do so from a feature auto-generated from an optional dependency.

RReverser:
I added an explicit `web_spin_lock` feature instead. It should be more future-proof and makes `cfg(...)` conditions cleaner too. Let me know what you think.

RReverser:
> 1. I wondered if we might need to switch from `wasm_sync` to another implementation in the future, in which case using its name directly might not be future-proof.

In fact, I already wonder if I should use [spin](https://crates.io/crates/spin) instead, as it's more generic (any `no_std` target), has more lock datastructures (e.g. `Once` which `wasm_sync` doesn't have), and seems much more popular / better maintained...

I can investigate and switch in a follow-up PR though, for now this PR is already a big improvement over previous state of art and would rather not block it on further investigations.

Cargo.toml
```diff
@@ -20,12 +20,20 @@ exclude = ["ci"]
...
 [dependencies.either]
 version = "1.0"
 default-features = false
 
+[features]
+# This feature switches to a spin-lock implementation on the browser's
+# main thread to avoid the forbidden `atomics.wait`.
+#
+# Only useful on the `wasm32-unknown-unknown` target.
+web_spin_lock = ["wasm_sync", "rayon-core/web_spin_lock"]
```
cuviper, line 35: This should avoid the automatic feature for the optional dependency:
```suggestion
web_spin_lock = ["dep:wasm_sync", "rayon-core/web_spin_lock"]
```

(and FWIW you can do that with features of the same name too.)

rayon-core/Cargo.toml
```diff
@@ -19,6 +19,15 @@ categories = ["concurrency"]
...
 crossbeam-utils = "0.8.0"
+wasm_sync = { version = "0.1.0", optional = true }
+
+[features]
+
+# This feature switches to a spin-lock implementation on the browser's
+# main thread to avoid the forbidden `atomics.wait`.
+#
+# Only useful on the `wasm32-unknown-unknown` target.
+web_spin_lock = ["wasm_sync"]
```
cuviper, line 30:
```suggestion
web_spin_lock = ["dep:wasm_sync"]
```

cuviper:
> I already wonder if I should use [spin](https://crates.io/crates/spin) instead,

It's significant that `wasm-sync` only spins on the main thread though -- I don't think we want the whole thread pool spinning. Even the basic idle sleep uses a condvar to wait for work.

RReverser:
> It's significant that `wasm-sync` only spins on the main thread though -- I don't think we want the whole thread pool spinning. Even the basic idle sleep uses a condvar to wait for work.

Yeah that's true. I raised an issue on `wasm-sync` suggesting they could switch to `spin` under the hood - this should preserve this benefit while making it easier to add more primitives like e.g. `Once` and `OnceLock` that are currently missing.

RReverser:
FWIW this PR should be good to go.

cuviper:
Thanks!

cuviper:
Hmm, let's try again.

PR #1112, opened 2024-01-06
doc: be more clear about what the 'spawn' does

### LINKED ISSUE #1095: Docs on "spawn" don't say what exactly this function does
Consider `spawn` docs: https://docs.rs/rayon/latest/rayon/fn.spawn.html . Surprisingly, the docs don't say what exactly this function does. Let's assume that there exist not busy threads in thread pool. What will do `spawn`? Synchronously execute on free thread? Or start execution and immediately return? Now imagine all threads are busy. What then? Block or immediately return?

bishopcheckmate:
fixes #1095

### DIFF
--- a/rayon-core/src/spawn/mod.rs
+++ b/rayon-core/src/spawn/mod.rs
@@ -4,8 +4,8 @@ use crate::unwind;
 use std::mem;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


cuviper:
Thanks!

PR #1113, opened 2024-01-17
Release rayon 1.8.1 and rayon-core 1.12.1

cuviper:
- The new `"web_spin_lock"` crate feature makes mutexes spin on the main
  browser thread in WebAssembly, rather than suffer an error about forbidden
  `atomics.wait` if they were to block in that context. Thanks @RReverser!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rayon"
-version = "1.8.0"
+version = "1.8.1"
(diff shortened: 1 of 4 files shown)


RReverser:
Incorporated into wasm-bindgen-rayon 1.2.0 release with much simplified notes. Thanks to @DouglasDwyer for wasm-sync making this so easy!

Now if only we could get id of the nightly requirement for atomics in Rust 😅

PR #1114, opened 2024-01-18
Run tests w/ fallback on wasm32-unknown-unknown

RReverser:
With this change, CI can actually run tests against wasm32-unknown-unknown with the fallback non-threaded implementation, which should help catch any bugs early.

This is a bit invasive, because this target requires a custom `#[wasm_bindgen_test]` attributes as well as a custom runner instead of the standard `#[test]`, but there is no way around it, and, having those changes only at the top of test files, I believe it's reasonably isolated.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,8 +61,9 @@ jobs:
       - run: cargo test --verbose --package rayon
       - run: cargo test --verbose --package rayon-core
 
-  # wasm32-unknown-unknown builds, and even has the runtime fallback for
-  # unsupported threading, but we don't have an environment to execute in.
+  # wasm32-unknown-unknown has a runtime fallback for unsupported threading.
+  # Proper threading needs more glue and is tested in wasm-bindgen-rayon,
+  # but we can at least verify that all tests can run on Wasm with the fallback.
   wasm:
     name: WebAssembly (standalone)
     runs-on: ubuntu-latest
@@ -71,7 +72,8 @@ jobs:
         include:
           - toolchain: stable
           - toolchain: nightly
-            cargoflags: --features web_spin_lock
+            components: rust-src
+            cargoflags: -Z build-std=panic_abort,std --features web_spin_lock
             rustflags: -C target-feature=+atomics,+bulk-memory,+mutable-globals
     env:
       RUSTFLAGS: ${{ matrix.rustflags }}
@@ -80,8 +82,10 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.toolchain }}
+          components: ${{ matrix.components }}
           targets: wasm32-unknown-unknown
-      - run: cargo build --verbose --target wasm32-unknown-unknown ${{ matrix.cargoflags }}
+      - uses: jetli/wasm-pack-action@v0.4.0
+      - run: wasm-pack test --chrome --headless -- --verbose ${{ matrix.cargoflags }}
 
   # wasm32-wasi can test the fallback by running in wasmtime.
   wasi:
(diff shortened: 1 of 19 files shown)


.github/workflows/ci.yaml
```diff
@@ -80,8 +82,10 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.toolchain }}
+          components: ${{ matrix.components }}
           targets: wasm32-unknown-unknown
-      - run: cargo build --verbose --target wasm32-unknown-unknown ${{ matrix.cargoflags }}
+      - uses: jetli/wasm-pack-action@v0.4.0
```
cuviper, line 87: I'm wary of adding 3rd-party github actions -- can we just use the official [installer](https://rustwasm.github.io/wasm-pack/installer/) like I did for `wasmtime` below?

.github/workflows/ci.yaml
```diff
@@ -80,8 +82,10 @@ jobs:
...
+      - run: wasm-pack test --chrome --headless -- --verbose ${{ matrix.cargoflags }}
```
cuviper, line 88: Should we also test `rayon-core`?

cuviper (commented):
> This is a bit invasive, because this target requires a custom `#[wasm_bindgen_test]` attributes as well as a custom runner instead of the standard `#[test]`, but there is no way around it, and, having those changes only at the top of test files, I believe it's reasonably isolated.

Is this needed for `src/*/test.rs` modules too?

cuviper:
BTW, you should be able to copy the action to `pr.yaml` temporarily to test it out.

.github/workflows/ci.yaml
RReverser: This is a reasonably popular action, but, sure, we can use official installer too, especially since wasm-pack provides prebuilt binaries and doesn't need recompilation from source.

RReverser:
> BTW, you should be able to copy the action to `pr.yaml` temporarily to test it out.

I tested it on my branch by just changing `on: merge_group` to `on: push` before submitting instead.

.github/workflows/ci.yaml
RReverser: Hm probably.

RReverser:
> Is this needed for `src/*/test.rs` modules too?

I didn't even see those but yeah :( This is a bit annoying...

RReverser:
I guess I mainly opened the PR to gather feedback first anyway. On one hand, it's a more invasive change than I'd like (and clearly needs to affect even more places), but, on another, it's potentially useful as it catches more errors.

Do you think it's worth pursuing further?

cuviper:
Maybe it's better to have a more targeted test? Just enough to exercise the `web_spin_lock` stuff a little, because apart from that we shouldn't really expect much to be different. It could even be a separate crate under `ci/` so we don't "infect" the normal build with extra dev-dependencies at all (though that _should_ be sufficiently cfg-isolated).

RReverser:
> Just enough to exercise the `web_spin_lock` stuff a little, because apart from that we shouldn't really expect much to be different.

This test doesn't really verify web_spin_lock - for that I have a small targeted test in wasm-bindgen-rayon. Instead, it verifies the non-threaded fallback behaviour in the browser, which was previously built but never executed.

RReverser:
I guess given that it's already tested by wasm32-wasi, it indeed shouldn't be any different barring bugs in Wasm engines (but then, it's outside of scope of Rayon anyway).

I'll probably close this for now.

cuviper:
Right, I'm happy with wasi for fallback testing. I thought you were trying to run real wasm threads given the `-Z build-std` and `target-features` involved.

RReverser:
> I thought you were trying to run real wasm threads given the `-Z build-std` and `target-features` involved.

Unfortunately not, as I wrote in that comment it would still require wasm-bindgen-rayon in addition to those flags.

cuviper:
Oh, indeed you did write that -- reading is hard sometimes. :slightly_smiling_face:

PR #1121, opened 2024-01-27
Let the octillion test use more threads

cuviper:


cuviper:
Windows got stuck: https://github.com/rayon-rs/rayon/actions/runs/7674967625

PR #1128, opened 2024-02-07
Add ParallelIterator::collect_vec_list

### LINKED ISSUE #1127: Support for some sort of collect()/extend() into `LinkedList<Vec<T>>`
From what I can tell, building a `LinkedList<Vec<T>>` is an efficient (one of the more efficient?) ways of collecting a parallel iterator, which is why `ParallelExtend for Vec` uses it. However, if you don't need specifically a Vec the process of concatenating all the vecs might be unnecessary overhead. It'd be nice if `ListVecConsumer` was exposed in some way, so that users could collect like that. (Yes, you could just `.fold(Vec::new, |vec, item| vec.push(item))`, but that's an element-wise push that skips specialization of `Extend for Vec`)

Directly implementing `FromParallelIterator<T> for LinkedList<Vec<T>>` might break type inference, but there could be a `collect_vec_list()` method on ParallelIterator, maybe.

coolreader18:
I wasn't exactly sure where the `collect_vec_list` implementation should go, since the necessary internals are in `extend` but it's logically a `collect` function. Also, I feel like the documentation could be better, I'd appreciate feedback on that.

Resolves #1127

### DIFF
--- a/src/iter/collect/mod.rs
+++ b/src/iter/collect/mod.rs
@@ -1,3 +1,5 @@
+use std::collections::LinkedList;
+
 use super::{IndexedParallelIterator, ParallelIterator};
 
 mod consumer;
@@ -62,6 +64,24 @@ where
         });
         right_result.unwrap()
     });
+}
+
+/// Collects the iterator into a linked list of vectors.
+///
+/// This is called by `ParallelIterator::collect_vec_list`.
+pub(super) fn collect_vec_list<I>(pi: I) -> LinkedList<Vec<I::Item>>
+where
+    I: ParallelIterator,
+{
+    match pi.opt_len() {
+        Some(len) => {
+            // Pseudo-specialization. See impl of ParallelExtend for Vec for more details.
+            let mut v = Vec::new();
+            super::collect::special_extend(pi, len, &mut v);
+            LinkedList::from([v])
+        }
+        None => super::extend::drive_list_vec(pi),
+    }
 }
 
 /// Create a consumer on the slice of memory we are collecting into.
--- a/src/iter/collect/mod.rs
+++ b/src/iter/collect/mod.rs
@@ -1,5 +1,3 @@
-use std::collections::LinkedList;
-
 use super::{IndexedParallelIterator, ParallelIterator};
 
 mod consumer;
--- a/src/iter/collect/mod.rs
+++ b/src/iter/collect/mod.rs
@@ -66,24 +66,6 @@ where
     });
 }
 
-/// Collects the iterator into a linked list of vectors.
-///
-/// This is called by `ParallelIterator::collect_vec_list`.
-pub(super) fn collect_vec_list<I>(pi: I) -> LinkedList<Vec<I::Item>>
-where
-    I: ParallelIterator,
-{
-    match pi.opt_len() {
-        Some(len) => {
-            // Pseudo-specialization. See impl of ParallelExtend for Vec for more details.
-            let mut v = Vec::new();
-            super::collect::special_extend(pi, len, &mut v);
-            LinkedList::from([v])
-        }
-        None => super::extend::drive_list_vec(pi),
-    }
-}
-
 /// Create a consumer on the slice of memory we are collecting into.
 ///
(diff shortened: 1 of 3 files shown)


src/iter/collect/mod.rs
```diff
@@ -64,6 +66,24 @@ where
...
+pub(super) fn collect_vec_list<I>(pi: I) -> LinkedList<Vec<I::Item>>
+where
+    I: ParallelIterator,
+{
+    match pi.opt_len() {
+        Some(len) => {
+            // Pseudo-specialization. See impl of ParallelExtend for Vec for more details.
+            let mut v = Vec::new();
+            super::collect::special_extend(pi, len, &mut v);
+            LinkedList::from([v])
```
coolreader18, line 81: Wasn't sure if I should re-use `pi.collect::<Vec<_>>()` here or not; theoretically that'd have an unnecessary second check of opt_len() but that would get optimized in almost all cases.

src/iter/collect/mod.rs
```diff
@@ -64,6 +66,24 @@ where
...
+///
+/// This is called by `ParallelIterator::collect_vec_list`.
+pub(super) fn collect_vec_list<I>(pi: I) -> LinkedList<Vec<I::Item>>
+where
+    I: ParallelIterator,
+{
+    match pi.opt_len() {
+        Some(len) => {
+            // Pseudo-specialization. See impl of ParallelExtend for Vec for more details.
+            let mut v = Vec::new();
```
martin-g, line 79:
```suggestion
            let mut v = Vec::with_capacity(len);
```

cuviper: This suggested change is fine, but I think it won't really matter since `special_extend` -> `collect_with_consumer` will call `reserve` anyway. Allocate now or later, but we're still only doing it once.

cuviper:
> I wasn't exactly sure where the `collect_vec_list` implementation should go, since the necessary internals are in `extend` but it's logically a `collect` function.

I think that bit of code is short enough that it could just be implemented directly in the trait method.

src/iter/mod.rs
```diff
@@ -2339,6 +2340,41 @@ pub trait ParallelIterator: Sized + Send {
...
+    /// best default choice, despite the overhead that comes from concatenating each
+    /// vector. Or, if this is an `IndexedParallelIterator`, you should also prefer to
+    /// just collect to a `Vec<T>`.
+    ///
+    /// Internally, most [`FromParallelIterator`]/[`ParallelExtend`] implementations
+    /// use this strategy; each job collecting their chunk of the iterator to a `Vec<T>`
+    /// and those chunks getting merged into a `LinkedList`, before then extending the
+    /// collection with each vector. This is the most efficient way to collect an
+    /// unindexed parallel iterator (again, indexed parallel iterators can be
+    /// efficiently collected simply into a vector).
```
cuviper, line 2357: I'm hesitant about the superlative, and I think we don't need the repeated statement about indexed iterators. How about this, adding a clue about *why* it's efficient:
```suggestion
    /// collection with each vector. This is a very efficient way to collect an
    /// unindexed parallel iterator, without much intermediate data movement.
```

coolreader18: That makes sense - the superlative felt wrong, but I wasn't sure how to word the why of it.

coolreader18:
Something I was thinking of including in the docs but wasn't sure about - mentioning the pattern of calling `.into_iter().flatten()` on the linked list? Like, `.collect_vec_list().into_iter().flatten()` is now a very straightforward/efficient way of turning a parallel iterator into a serial one, though I'm unsure if that should be encouraged or where in the docs that should go.

coolreader18:
Also, added a check for length of 0 returning an empty LL, though yeah I'm not sure if that's necessary.

cuviper (approved):
Looks good, thanks!

coolreader18:
Would it be possible to cut a release with this?

cuviper:
This is now published in rayon 1.9.0.

coolreader18:
Thank you so much!

PR #1130, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


cuviper:
For a library crate like this, I'm not a fan of dependabot-forced cargo updates. End users building an application are free to use whatever semver-compatible versions they like. Is there a way to configure that only for major dependency upgrades?

oriontvv:
Yes, we can specify update-types: major [doc](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups)

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
```
cuviper, line 4: Will it scan for all `Cargo.toml` in the workspace, or only the root?

martin-g:
All will be scanned

cuviper:
OK, thanks! We'll try it out.

PR #1131, opened 2024-02-12
use clone_from in quicksort bench

TheIronBorn:


cuviper:
Sounds good -- can you do the same in the merge-sort bench?

TheIronBorn:
Done, also allocated before timing starts

cuviper:
Thanks!

TheIronBorn:
I suppose the first iteration might elide the memcpy. If anyone can suggest a way to avoid that please do.

cuviper:
I doubt it will do that, but we could replace that outer clone with just `Vec::with_capacity`.

PR #1133, opened 2024-02-19
Fixing docs

Vagelis-Prokopiou:
Nothing major. Just adding a missing full stop that I spotted.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,7 +4,7 @@
 #![warn(rust_2018_idioms)]
 
 //! Data-parallelism library that makes it easy to convert sequential
-//! computations into parallel
+//! computations into parallel.
 //!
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -4,7 +4,7 @@
 #![warn(rust_2018_idioms)]
 
 //! Data-parallelism library that makes it easy to convert sequential
-//! computations into parallel
+//! computations into parallel.
```
cuviper, line 7: Well, it's not a proper sentence either -- maybe add "Rayon is a ..." to start?

Vagelis-Prokopiou: Sure. Done.

cuviper (approved):
Thanks!

Issue #1139, opened 2024-03-04
Potential to modify ordering for split_count and tour_counter


wang384670111:
https://github.com/rayon-rs/rayon/blob/990841b313d1a566d2b7f4f33c61e0f0ac9dc7b6/src/iter/par_bridge.rs#L113-L115

`SeqCst` is overly restrictive. I believe that the ordering can be appropriately modified.

In `par_bridge` model, I believe that `split_count` is just a counter across threads and does not synchronize with other locals. Therefore, using `Relaxed` ordering should suffice.

https://github.com/rayon-rs/rayon/blob/990841b313d1a566d2b7f4f33c61e0f0ac9dc7b6/rayon-demo/src/tsp/solver.rs#L67

Similarly, In `solver` model,  `tour_counter` is just a counter across threads, using `Relaxed` ordering should suffice.

(happy to make a PR if this looks reasonable)

cuviper:
It's been this way in `par_bridge` since it was first added in #550, but I don't think we paid much attention to that ordering. I agree that `Relaxed` should be fine, but I also think it won't make much difference, as this only executes once per thread for a given `par_bridge`. The real contention comes when they all wait on the same `Mutex<Iter>`.

The tsp counter is fine to change too. You're welcome to make a PR!

wang384670111:
Related Link for PR:
https://github.com/rayon-rs/rayon/pull/1140

PR #1140, opened 2024-03-05
Potential to modify ordering for split_count and tour_counter

wang384670111:
Discussed as follows:
https://github.com/rayon-rs/rayon/issues/1139


cuviper:
Thanks!

PR #1143, opened 2024-03-11
chore: remove repetitive word

acceptacross:


rayon-core/src/lib.rs
```diff
@@ -228,7 +228,7 @@ type PanicHandler = dyn Fn(Box<dyn Any + Send>) + Send + Sync;
 type StartHandler = dyn Fn(usize) + Send + Sync;
 
 /// The type for a closure that gets invoked when a thread exits. The
-/// closure is passed the index of the thread on which is is invoked.
+/// closure is passed the index of the thread on which is invoked.
```
cuviper, line 231: This meant to be "on which **it** is invoked", just like the docs for `StartHandler`.

cuviper (changes requested):

rayon-core/src/lib.rs
acceptacross: Modified.

Please review again

cuviper:
Thanks - can you clean up the commit message? Somehow your sign-off got duplicated, and also this is now correcting rather than removing.

(I might usually just fix that myself, but I hesitate more when a sign-off is involved.)

acceptacross:
> Thanks - can you clean up the commit message? Somehow your sign-off got duplicated, and also this change is now _correcting_ rather than _removing_.
> 
> (I might usually just fix that myself, but I hesitate more when a sign-off is involved.)

Sorry, it was my mistake.
Modified, please review again!

cuviper (approved):
Thanks!

PR #1150, opened 2024-03-31
Change FAQ example with race conditions

JulianKnodt:
Previously this example in the FAQ as written would've introduced a race condition.
While I know the objective of this example is not to show exactly correct code, it seemed a bit odd to me since there's a whole section earlier about preventing race conditions. The last change was a while ago, so the API wasn't the same, but it would be good to update it so there's no race conditions.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -211,15 +211,14 @@ this shortest route, you can just stop and avoid wasted effort. In
 sequential land, you might model this "best result" as a shared value
 like `Rc<Cell<usize>>` (here the `usize` represents the length of best
 path found so far); in parallel land, you'd use a `Arc<AtomicUsize>`.
-Now we can make our search function look like:
 
 ```rust
-fn search(path: &Path, cost_so_far: usize, best_cost: &Arc<AtomicUsize>) {
+fn search(path: &Path, cost_so_far: usize, best_cost: &AtomicUsize) {
     if cost_so_far >= best_cost.load(Ordering::SeqCst) {
         return;
     }
-    ...
-    best_cost.store(...);
+    // Using `fetch_min` to avoid a race condition, in case it changed since `load`.
+    best_cost.fetch_min(..., Ordering::SeqCst);
 }
 ```
 


FAQ.md
```diff
@@ -210,16 +210,19 @@ way, when you are searching down some path that's already longer than
 this shortest route, you can just stop and avoid wasted effort. In
 sequential land, you might model this "best result" as a shared value
 like `Rc<Cell<usize>>` (here the `usize` represents the length of best
-path found so far); in parallel land, you'd use a `Arc<AtomicUsize>`.
+path found so far); in parallel land, you'd use a `&AtomicUsize`.
+Note that `Arc` is no longer necessary because atomic types provide interior mutability.
```
cuviper, line 214: This note doesn't make sense -- the earlier `Cell` also provides interior mutability, but `Rc` and `Arc` are for shared ownership. Let's leave that change out, because it's unrelated to the race you identified.

JulianKnodt: I was thinking it doesn't make sense to wrap an AtomicUsize in Arc, since it's redundant interior mutability, I'll revert it back so there's no change.

FAQ.md
```diff
@@ -211,15 +211,16 @@ this shortest route, you can just stop and avoid wasted effort. In
...
 ```rust
-fn search(path: &Path, cost_so_far: usize, best_cost: &Arc<AtomicUsize>) {
+fn search(path: &Path, cost_so_far: usize, best_cost: &AtomicUsize) {
     if cost_so_far >= best_cost.load(Ordering::SeqCst) {
         return;
     }
     ...
-    best_cost.store(...);
+    // Note: doing `best_cost.store(...)` will introduce race conditions.
+    // Using `fetch_*` such as `fetch_min` will help prevent race conditions.
```
cuviper, line 222:
```suggestion
    // Using `fetch_min` to avoid a race condition, in case it changed in the meantime.
```

FAQ.md
```diff
@@ -211,15 +211,16 @@ this shortest route, you can just stop and avoid wasted effort. In
...
+    best_cost.fetch_min(cost_so_far, Ordering::SeqCst);
```
cuviper, line 223: The idea is that `best_cost` is a total, while `cost_so_far` is only partial. So the first load means "if we're already too big, give up," and then it would finish computing the current cost and eventually try to `fetch_min` that as the new best.

```suggestion
    best_cost.fetch_min(..., Ordering::SeqCst);
```

Or we could expand the first ellipses, something like:

```rust
    let total_cost = cost_so_far + ...;
    best_cost.fetch_min(total_cost, Ordering::SeqCst);
```

FAQ.md
cuviper: The interior mutability in `Arc<T>` is *only* for its own reference counter, not the `T` value.

FAQ.md
JulianKnodt: Went with the first suggestion of changing `fetch_min(..., Ordering::SeqCst)`.

cuviper:
Thanks!

PR #1156, opened 2024-04-04
Sync `par_sort*` with the standard library

cuviper:
Our parallel sort methods were already based on the standard library,
but they're out of date. Most of that implementation was moved to a
unified `core::slice::sort` module, with manual allocation functions
passed from the `alloc` crate and a simplified `Vec`.

This PR updates `rayon` to a unified `slice::sort` based on `core`'s,
but dropping the manual allocation stuff to just use the real `Vec`. The
rest of the changes are straightforward parallelization, like using `Fn`
instead of `FnMut`, and the extended `par_mergesort` additions are still
the same as we had before.


adamreichold:
Does this update imply any functional changes or does it really only update the code organisation to match `std`? Meaning was the implementation itself already functionally equivalent, just the source organised differently?

cuviper:
The implementations used to be separate in the standard library too, until rust-lang/rust#104672, and we were roughly in sync with that former code. So the main point here is to make it easier to keep up with improvements, especially because I'm not an expert in sorting algorithms, but I think I understand the parallelization additions well enough. :)

The [commit log](https://github.com/rust-lang/rust/commits/master/library/core/src/slice/sort.rs) on `core::slice::sort` has some performance tuning that sounds like it could help `par_sort` too, but it's hard to tell. The benchmarks we have under `rayon-demo` look mostly the same or slightly better on my runs, but there's a *lot* of variance.

adamreichold:
Thanks!

I am mainly asking as to how this should be communicated and this sounds like "just reorganising code to keep in sync with std" not like "pulling fundamental improvements from std which Rayon was missing out on until now."

cuviper:
Yeah, that sounds right. I would categorize this under implementation details, and while there may be some incremental improvement, I don't think it would warrant a release note.

PR #1157, opened 2024-04-08
Re-wrap the FAQ to 80 character width

atouchet:


cuviper:
What text width did you use? It looks rather narrow...

atouchet:
Those are 70 character widths which is what the Readme and the rest of the FAQ already appear to be using. This is just adjusting a few paragraphs to be consistent with those.

cuviper:
I think it may have been 72, which would still fit some of the existing lines here. But if we're going to rewrap, I'd prefer to go wider to 80.

atouchet:
I tried taking a look and I don't think the rest of the document matches 72 but regardless I could rewrap it to 80 if you would prefer that.

cuviper:
We haven't enforced anything, so I have no doubt that it's inconsistent. Yes, let's do 80 please.

atouchet:
Okay. Should the Readme be rewrapped too? It is also currently using 70.

cuviper:
Sure, if you're interested, go ahead in the readme too.

atouchet:
There ended up being more docs in this repo than I originally realized so I opted to just re-wrap the FAQ page to 80.

cuviper:
Thanks!

PR #1158, opened 2024-04-17
chore: fix some typos in comments

TechVest:


src/iter/plumbing/mod.rs
```diff
@@ -25,7 +25,7 @@ pub trait ProducerCallback<T> {
 
     /// Invokes the callback with the given producer as argument. The
     /// key point of this trait is that this method is generic over
-    /// `P`, and hence implementors must be defined for any producer.
+    /// `P`, and hence implementers must be defined for any producer.
```
cuviper, line 28: This is not a typo. Both are valid spellings, and while you may find some internet sources that try to provide rules when to use each, "-ors" is the spelling used in Rust for types that implement a trait. For example:
https://doc.rust-lang.org/std/clone/trait.Clone.html#implementors

cuviper:
Thanks anyway!

Issue #1174, opened 2024-06-14
Thread pool without work stealing


nhukc:
We are using a rayon thread pool to run a rust-based userspace driver.  A mutex must be acquired before calls can be made to the driver.

We are running into a common deadlock problem with rayon work-stealing.
This issue occurs because the third party code that invokes our driver is also using rayon.

The deadlock results from the following events:
1. Third party global thread pool is running N jobs
2. Third party thread A invokes our library
3. Third party thread A acquires our driver mutex
4. Third party thread A creates a scope within our driver thread pool
5. Third party thread A steals work
6. Third party thread A invokes our library
7. Third party thread A is waiting on a lock that it already holds.
8. Deadlock

The rayon thread pool API fits our use-case perfectly, but the implementation detail of work-stealing causes deadlocks.

Is there an easy way to modify the thread pool internals to prevent work-stealing?

I'm not looking to merge with upstream unless this feature is wanted by others. I'd be OK with maintaining a non-work-stealing fork if a maintainer could give me some tips for how to do this.

nhukc:
It may be worth noting that our driver thread pool executes exactly as many jobs as there are threads in the pool.

cuviper:
The general issue with `Mutex` is also described in #592.

One of the ideas in that thread is to have a fully-blocking version of `ThreadPool::install`, so cross-pool calls won't work-steal in the first one anymore. In your scenario, that should block "third party thread A" until your driver is done.

nhukc:
Yes, that would solve my problem.

Is there any on-going work in that direction? Or tips for how to implement such a feature?

nhukc:
It looks like the meat of the work would be in and around this function.

https://github.com/rayon-rs/rayon/blob/3e3962cb8f7b50773bcc360b48a7a674a53a2c77/rayon-core/src/registry.rs#L533

cuviper:
Or rather, avoid that function in this case and call `in_worker_cold` instead.

nhukc:
I implemented the suggestion. I did not call `in_worker_cold` because that function has a useful-looking debug assert that conflicts with this use case.

Issue #957, opened 2022-07-10
Priority inversion when executing jobs of heterogeneous length


JoeyBF:
We're using rayon to power our mathematical software for HPC and we're running into this issue: SpectralSequences/sseq#105. Here's a high-level overview.

We have two kinds of jobs, long ones that take several hours, and short ones that take milliseconds. Short jobs are spawned by `par_iter_mut`, they are independent and do not spawn other jobs themselves. Long jobs on the other hand have nontrivial dependencies on each other. We schedule them using channels and spawn them using `spawn` on an `in_place_scope`. Each long job will spawn potentially millions of short jobs that will need to finish before the long job returns. The inversion happens in the following situation:

- Thread 1 starts computing long job A.
- Thread 1 enters the `par_iter_mut`, which calls `join_context`, which itself spawns (say) two short jobs a and b.
- Thread 1 starts executing a while thread 2 starts b.
- Some other thread 3 finishes a long job, and now a new long job B is available.
- Thread 1 finishes the short job, sees that thread 2 is still busy, and starts looking for other things to do in the meantime.
- Thread 1 picks up job B.
- Thread 2 finishes, but by then thread 1 is stuck in B, and job A stalls.

Note that this is different from #630 because in our case the short jobs are called by the long jobs, and I also don't see how affinity would solve this problem.

I tried building different threadpools in https://github.com/JoeyBF/sseq/tree/bandaid, and while the performance seems to improve, the logs still show that long jobs sometimes stall.

My proposal is to implement priorities in a very naive way. It would be sufficient to have to ability to tag jobs with priority values (say an `i32` defaulting to 0) and to have access to a new `join_priority` function that only allows waiting threads to steal jobs of equal or higher priority relative to the job it just finished

cuviper:
> I tried building different threadpools in https://github.com/JoeyBF/sseq/tree/bandaid, and while the performance seems to improve, the logs still show that long jobs sometimes stall.

I don't have the domain knowledge for your problem, but AFAICS you're only separating "matrix" and "nassau" pools there -- are those your A and B type of jobs? If you have multiple "matrix" jobs running, then those can still steal from each other in that shared pool. But it looks like you are creating "nassau" pools on the fly each time, so those should run independently, as long as they don't also call into that shared "matrix" pool.

[...]

Issue #592, opened 2018-09-05
Using rayon under a Mutex can lead to deadlocks


cuviper:
https://users.rust-lang.org/t/using-rayon-in-code-locked-by-a-mutex/20119

@kornelski presented the following pseudo-code representing a possible deadlock:

```rust
fn mutex_and_par() {
   some_mutex.lock().par_iter().collect();
}

collection.par_iter().for_each(mutex_and_par);
```

I think the way this fails is something like:

- Thread 1 starts processing one of the `for_each(mutex_and_par)` calls.
  - grabs the lock
  - splits up the `par_iter()` into a few jobs
  - starts the `collect()` on one of those jobs
- Another thread steals one of those jobs to help out
- Thread 1 finishes its current job
  - looks for the other jobs and sees they're stolen elsewhere
  - goes to steal more work itself while it's waiting...
  - finds another of the `for_each(mutex_and_par)` calls
  - tries to grab the lock (again) ... DEADLOCK

If serialized, this code would never be a problem, but a work-stealing threadpool can cause it to implicitly recurse the lock.  I'm not sure what we can do about this, or if there's any better advice we can give than just "don't call rayon under a mutex"...

kornelski:
Solutions that come to my mind:

* Allow running a job/parallel iterator on a separate "private" threadpool. I guess current `ThreadPool.install()` won't cut it, as it's still global, and it'd merely delay the same cross-mutex stealing.

* Have `RayonAwareMutex`. When a code ran from a Rayon job tries to acquire the mutex, and it's locked, it switches to another job instead of blocking.

nikomatsakis:
Mostly, my expectation is that the answer here would be "don't do that", but it is sort of disappointing, since we would like you to be able to add `par_iter` kind of 'for free' and have it work out.

That doesn't realistically work around blocking things though, and mutexes fit that model of course.

nikomatsakis:
I think that using a `ThreadPool::install` inside the mutex would prevent the problem in this particular case, though..? 

(Of course, you may not have control of all the relevant code.)

cuviper:
>     * Allow running a job/parallel iterator on a separate "private" threadpool. I guess current `ThreadPool.install()` won't cut it, as it's still global, and it'd merely delay the same cross-mutex stealing.

[...]

PR #1175, opened 2024-06-18
Add full_blocking feature to thread pool

nhukc:
This PR adds an option `full_blocking` to `ThreadPoolBuilder` that affects the behavior of nested thread pools.

When a job on a parent thread pool creates a job in a child thread pool and `full_blocking` is true, the parent job will block until the child job is completed. This is different from the default behavior where the parent thread is allowed to work on other jobs in the parent thread pool while the child job completes.

This behavior is useful for avoiding deadlocks caused by work-stealing and is helpful for instrumentation based profiling in multi-threaded settings.

See for more details:
https://github.com/rayon-rs/rayon/issues/1174

### DIFF
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -200,6 +200,12 @@ pub struct ThreadPoolBuilder<S = DefaultSpawn> {
     /// Closure invoked on worker thread exit.
     exit_handler: Option<Box<ExitHandler>>,
 
+    /// Affects the behavior of nested thread pools. If true, jobs from a parent thread pool will
+    /// block until jobs in this thread pool are completed. If false, jobs from a parent thread
+    /// pool are free to complete other jobs while jobs are processed in this thread pool. This is
+    /// useful for avoiding deadlock when using mutexes. Default is false.
+    full_blocking: bool,
+
     /// Closure invoked to spawn threads.
     spawn_handler: S,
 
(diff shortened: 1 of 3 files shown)


nhukc:
Feel free to push directly to this PR.

The documentation could be improved by matching the language to the rest of the rayon docs.

I'm not 100% confident in the implementation of `in_worker_cross_blocking`. It works for my test, but I could easily be missing something in the details.

JoeyBF (commented):
Thanks for the PR! One small suggestion would be to make `full_blocking` take a bool as an argument, so it can be toggled on _or off_. This seems to be more in line with the other builder methods.

I'll add that I tested my code using this branch and this does solve my own issue #957, which is essentially a variation of #592 :tada:

nhukc:
Thanks for taking a look at the PR!

The most similar function in the ThreadPoolBuilder is
```
pub fn use_current_thread(self) -> ThreadPoolBuilder<S>
```

Which sets a bool to true when called. I matched that function signature.

JoeyBF:
That's fair. I would argue that `use_current_thread` would also need a bool, but as far as this PR is concerned it's probably better to just be consistent.

nhukc:
@cuviper Are you interested in merging this? Otherwise I will close it.

cuviper:
I need some time to think about the implications of this change, to make sure there aren't any big foot-guns here. But you're also helping solve a foot-gun, so that's going to be a balancing act.

I might also bikeshed the name, but that's not a big deal.

fkysly:
> I need some time to think about the implications of this change, to make sure there aren't any big foot-guns here. But you're also helping solve a foot-gun, so that's going to be a balancing act.
> 
> I might also bikeshed the name, but that's not a big deal.

How's the progress? Will this MR resolve the issue?

Issue #754, opened 2020-04-21
Avoinding reallocations when doing a cartesian product


lovesegfault:
I have some code that looks like this:

[*playground*](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=204d90d9d7de24575160055e92c7d5e5)
```Rust
use rayon::prelude::*;

#[derive(Debug)]
struct BigStruct {
    a: u128,
    b: u128,
    c: u128,
    d: u128,
    e: u128
}

impl BigStruct {
    fn new(x: u64, y: u64) -> Self {
        BigStruct {
            a: x.into(),
            b: x.into(),
            c: x.into(),
            d: y.into(),
            e: y.into(),
        }
    }
}



fn main() {
    const X: u64 = 5;
    const Y: u64 = 5;
    let pairs: Vec<BigStruct> = (0..X)
        .into_par_iter()
        .flat_map(|x| (0..Y).into_par_iter().map(move |y| (x, y)))
        .map(|(x, y)| BigStruct::new(x, y))
        .collect();
    
    pairs.iter().for_each(|b| println!("{:?}", b));
}
```

When I went to profile it, with a very large `X` and `Y` I found that most of my execution time was spent on malloc. The natural next step, then, was to try and preallocate the `Vec` to have the correct size, as this at least means no calls to realloc. I learned about [`collect_into_vec()`](https://docs.rs/rayon/1.3.0/rayon/iter/trait.IndexedParallelIterator.html#method.collect_into_vec) and decided to try and use it:

[*playground*](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=204d90d9d7de24575160055e92c7d5e5)
```Rust
use rayon::prelude::*;

#[derive(Debug)]
struct BigStruct {
    a: u128,
    b: u128,
    c: u128,
    d: u128,
    e: u128,
}

impl BigStruct {
    fn new(x: u64, y: u64) -> Self {
        BigStruct {
            a: x.into(),
            b: x.into(),
            c: x.into(),
            d: y.into(),
            e: y.into(),
        }
    }
}

fn main() {
    const X: u64 = 5;
    const Y: u64 = 5;
    let mut pairs: Vec<BigStruct> = Vec::with_capacity((X * Y) as usize);
    (0..X)
        .into_par_iter()
        .flat_map(|x| (0..Y).into_par_iter().map(move |y| (x, y)))
        .map(|(x, y)| BigStruct::new(x, y))
[...]

PR #1182, opened 2024-07-20
Add cartesian_product to ParallelIterator

nascheinkman:
I was trying to use cartesian products from itertools and noticed it's not implemented for rayon and has an open issue: https://github.com/rayon-rs/rayon/issues/754

I wasn't trying to preallocate, just run a computation in the parallel iterator using a cartesian product generated by ranges. I ended up with this solution that worked for me. It's not implemented for `IndexedParallelIterator` like suggested in the issue, only for `ParallelIterator`. I don't know enough to know how this would impact performance.

Feel free to close this if it's not done properly or out of scope. I'm also open to suggestions if there's a better way to do this.

### DIFF
--- a/src/iter/cartesian_product.rs
+++ b/src/iter/cartesian_product.rs
@@ -1,0 +1,48 @@
+use super::plumbing::*;
+use super::repeat::*;
+use super::{IndexedParallelIterator, IntoParallelIterator, ParallelIterator};
+
+/// `CartesianProduct` is an iterator that combines `i` and `j` into a single iterator that
+/// iterates over the cartesian product of the elements of `i` and `j`. This struct is created by
+/// the [`cartesian_product()`] method on [`ParallelIterator`]
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+#[derive(Debug, Clone, Copy)]
+pub struct CartesianProduct<I, J>
+where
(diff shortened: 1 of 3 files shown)


src/iter/cartesian_product.rs
```diff
@@ -0,0 +1,48 @@
...
+{
+    type Item = (I::Item, J::Item);
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+    where
+        C: UnindexedConsumer<Self::Item>,
+    {
+        self.i
+            .into_par_iter()
+            .map(|i_item| repeat(i_item.clone()).zip(self.j.clone()))
```
adamreichold, line 44: Are the calls to `into_par_iter` (`I` already is a `ParallelIterator` here) and `i_item.clone()` (`repeat` will clone it repeatedly, but the first clone seems unnecessary) really required here?

I also think we can use `flat_map` here instead of `map` and `flatten` to possibly benefit from additional optimizations in its implementation.

nascheinkman: Good catch!

wagnerf42:
hi, you have to be careful because flat_map can end up being quite expensive compared to flat_map_iter.
i'm not sure what's the right choice here because on small products you might want to get fully parallel.

Issue #1189, opened 2024-08-16
Switching par_iter makes memory usage explode (40x)


BloodStainedCrow:
rayon 1.10.0

I wrote a small Mandelbrot visualizer including this snippet:

```
let data: Vec<u8> = num_iterations
        .par_iter()
        .map(|num_iterations| (counts[*num_iterations as usize], num_iterations))
        .map(|(counts_cum, num_iterations)| {
            (
                1.0 - f64::from(counts_cum) / (WIDTH * HEIGHT) as f64,
                num_iterations,
            )
        })
        .map(|(brightness, num_iterations)| (brightness.sqrt(), num_iterations))
        .flat_map(|(mut brightness, num_iterations)| {
            if *num_iterations == MAX_ITER {
                brightness = 0.0;
            }

            [
                (brightness * 255.0) as u8,
                (brightness * 255.0) as u8,
                (brightness * 255.0) as u8,
                255u8,
            ]
        })
        .collect();
```

num_iterations is a Vec<u32> of length 10_000 * 10_000, counts a Vec<u32> of length 10_000. The largest memory usage I observed was around 380MB which seems to fit the math quite well: 10_000 * 10_000 * 4 bytes ~= 400MB.

I am on a 6/12 core processor, so the increase is not linear with the number of cores as I first suspected.

When I switch to using ``par_iter`` The usage spikes to over 16GB (increase of 40x) after a couple of seconds. The runtime is also much slower than single threaded. (I was unable to time it, since it took too long on windows and kept getting OOM killed on Linux).

Incase it is useful the repo with the full code is [here](https://github.com/BloodStainedCrow/mandelbrot).

cuviper:
Unfortunately, `flat_map` is kind of a performance footgun, especially to `collect`. It causes another full round of parallel splitting *every time*, which is quite wasteful for just your 4 elements. It also prevents this from being an `IndexedParallelIterator`, so `collect` can't write results directly into a final `Vec` -- instead it has to create a bunch of temporary `Vec` and then merge them later. In theory that would only result in about double the overall memory usage, but the allocator might not be releasing the memory back to the OS for so many differently-sized allocations.

You could use `flat_map_iter` for the first problem, so that part will run sequentially in each parallel job, but that doesn't totally solve the indexing / temporary-`Vec` problem. Since we can see ourselves that the result will be exactly `4 * len`, you could set that up a bit manually instead, something like:

```rust
[...]

PR #1195, opened 2024-09-13
Add const_length function to ParallelIterator

BloodStainedCrow:
Improves #1189.

As discussed there, this adds a new function to ```ParallelIterator```: ```const_length``` which can return a value if the length of a ```ParallelIterator``` can be known statically.

This function has a default implementation which returns ```None``` by default, so this is not a [breaking change](https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-default-item).

This change improves the speed of the algorithm in #1189 by a factor of ~2.5 and peak memory usage by about a factor of ~15x. [benchmark](https://github.com/BloodStainedCrow/rayon_memory_test).

The footgun of splitting unconditionally (which is the main reason the speedup is not that high), is not fixed by this PR though.


BloodStainedCrow:
Not part of this PR but I also have an implementation for stopping ```flat_map``` from splitting the inner iterator unconditionally, which would stop this footgun almost completely (About 400x faster, same speed as ```flat_map_iter``` on arrays, ~100x on iterators), but it currently has the problem of not waking (and therefore starting their work stealing) all threads, if the outer iterator is smaller than the number of cores.

BloodStainedCrow:
@cuviper Sorry for the ping, but does this seem reasonable?

PR #1201, opened 2024-10-13
Improved comments and error handling

sharma-shray:


rayon-core/src/job.rs
```diff
@@ -221,19 +221,20 @@ impl<T> JobResult<T> {
...
         match self {
-            JobResult::None => unreachable!(),
+            JobResult::None => unreachable!("JobResult should not be None after completion."),
             JobResult::Ok(x) => x,
             JobResult::Panic(x) => unwind::resume_unwinding(x),
         }
     }
 }
 
+
```
cuviper, line 237: This stray line made it fail the format check.
```suggestion
```

PR #1223, opened 2025-01-01
Update lots of doc links to use item paths instead of file paths

LikeLakers2:
Hi! I noticed that a lot of intra-doc links in `rayon-core` and `rayon` uses webpage links (i.e. `iter/trait.ParallelExtend.html#tymethod.par_extend`) instead of item paths (i.e. `iter::ParallelExtend::par_extend()`).

This PR converts as many intra-doc links as I could find into item paths. This ensures that if Rustdoc ever changes how its documentation is structured in the file system (unlikely, but not impossible), `rayon` and `rayon-core`'s docs will continue to function as normal.

This PR also fixes a couple small pieces of documentation which I assume were meant to link to another item.

I made sure to double-check all of my work - but if you find any documentation links I missed, or any that are now broken, please let me know.

Notes:
* https://github.com/rayon-rs/rayon/blob/d179ea9fa69cab08a3b2762ff9b15249a71b9b94/rayon-core/src/lib.rs#L414 - If `crossbeam` were a dependency of `rayon-core`, I would be able to change this to an item path. But alas, it is not.
* https://github.com/rayon-rs/rayon/blob/d179ea9fa69cab08a3b2762ff9b15249a71b9b94/rayon-core/src/join/mod.rs#L41 - Attempting to change this link to use an item path (i.e. `rayon::slice::ParallelSliceMut::par_sort()`) results in a broken link warning.
* https://github.com/rayon-rs/rayon/blob/d179ea9fa69cab08a3b2762ff9b15249a71b9b94/src/array.rs#L6 - I do not know how to link to the primitive using an item path; using `std::array` links to the `std::array` module rather than the `[T; N]` primitive.

### DIFF
--- a/rayon-core/src/broadcast/mod.rs
+++ b/rayon-core/src/broadcast/mod.rs
@@ -13,9 +13,9 @@ mod test;
 /// within that threadpool. When the call has completed on each thread, returns
 /// a vector containing all of their return values.
 ///
-/// For more information, see the [`ThreadPool::broadcast()`][m] method.
+/// For more information, see the [`ThreadPool::broadcast()`] method.
 ///
-/// [m]: struct.ThreadPool.html#method.broadcast
+/// [`ThreadPool::broadcast()`]: crate::ThreadPool::broadcast()
 pub fn broadcast<OP, R>(op: OP) -> Vec<R>
 where
     OP: Fn(BroadcastContext<'_>) -> R + Sync,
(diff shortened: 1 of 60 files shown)


cuviper:
Hi! Sorry I waited so long to review this. I went ahead and updated a few things:
* Rebased and resolved the conflicts in `repeat.rs`
* Added a `mod@` disambiguator on `std::vec` to resolve between the module and macro.
* Added `primitive@array` for that array item syntax you weren't sure about.

LikeLakers2:
@cuviper No worries, I don't mind it having taken a while to review. :)

Thank you for fixing the remaining issues on my behalf, too! And especially for rebasing on my behalf - I had stopped paying attention to this PR, so I only would've noticed once I got a comment notification.

PR #1226, opened 2025-01-14
In partial_insertion_sort, moved the SHORTEST_SHIFTING check outside of loop

zommiommy:
The check does not depend on the loop and can be performed at the start.
While the compiler probably optimizes this, it was confusing to read.

### DIFF
--- a/src/slice/sort.rs
+++ b/src/slice/sort.rs
@@ -223,6 +223,11 @@ where
     let len = v.len();
     let mut i = 1;
 
+    // Don't shift elements on short arrays, that has a performance cost.
+    if len < SHORTEST_SHIFTING {
+        return false;
+    }
+    
     for _ in 0..MAX_STEPS {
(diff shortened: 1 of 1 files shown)


src/slice/sort.rs
```diff
@@ -238,11 +243,6 @@ where
             return true;
```
cuviper, line 243: If the `SHORTEST_SHIFTING` condition is moved up, it changes this case that would have returned `true`. I haven't fully analyzed that effect, but it's not just about "readability" at that point.

cuviper (commented):
Note that this code was originally an adaptation of the standard library's sort, although that's since been replaced by rust-lang/rust#124032.

src/slice/sort.rs
zommiommy: You are absolutely right. The behaviour on short sorted slices changes, my bad.

cuviper: OK, then I'll close this - thanks anyway!

PR #1228, opened 2025-01-28
Update LICENSE-MIT

maximevtush:
Updated the copyright year from 2010 to 2025 in the LICENSE file


cuviper:
I would rather just remove the year, like it was done in Rust repo.

cuviper:
On further reflection, I don't think this is an appropriate change for a first-time contributor, even though it's trivial. I will consider updating the license file myself next time I publish the crates.

PR #1236, opened 2025-04-02
fix Update README.md

romashka-btc:
fix err


cuviper:
What's wrong with it? The current url loads fine.

cuviper:
I decided to just remove gitter, #1240, because we don't really use it.

PR #1238, opened 2025-04-04
fix Update README.md

MonkeyKing44:
fix old to a new one


cuviper:
I decided to just remove gitter, #1240, because we don't really use it.

PR #1239, opened 2025-04-08
rename `repeatn` to `repeat_n`, deprecate `repeatn`, for parity with std

### LINKED ISSUE #1237: Rename repeatn
As of rust 1.82.0, The standard library includes the function repeat_n (https://doc.rust-lang.org/std/iter/fn.repeat_n.html) which takes an element and a usize and constructs the iterator which repeats the element `n` times.

Rayon includes the same function (https://docs.rs/rayon/latest/rayon/iter/fn.repeatn.html) but uses a slightly different name (repeatn instead of repeat_n) .

Clearly not a high priority but it would be nice to fix this so the names match up.

VitalyAnkh:
fixes #1237

### DIFF
--- a/src/iter/repeat.rs
+++ b/src/iter/repeat.rs
@@ -12,7 +12,7 @@ pub struct Repeat<T: Clone + Send> {
 /// cloning it). Note that this iterator has "infinite" length, so
 /// typically you would want to use `zip` or `take` or some other
 /// means to shorten it, or consider using
-/// [the `repeatn()` function](fn.repeatn.html) instead.
+/// [the `repeat_n()` function](fn.repeat_n.html) instead.
 ///
 /// # Examples
 ///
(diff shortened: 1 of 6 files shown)


src/iter/repeat.rs
```diff
@@ -111,10 +111,19 @@ pub struct RepeatN<T: Clone + Send> {
...
 /// assert_eq!(x, vec![(22, 0), (22, 1), (22, 2)]);
 /// ```
+pub fn repeat_n<T: Clone + Send>(elt: T, n: usize) -> RepeatN<T> {
+    RepeatN {
+        element: elt,
+        count: n,
+    }
+}
+
+#[doc(hidden)]
```
cuviper, line 125: I don't think we should hide it from docs.
```suggestion
```

(We hide `find` and `position` because they only exist anymore to guide people porting from `Iterator`, versus something we changed ourselves.)

VitalyAnkh: Thanks! I have removed `#[doc(hidden)]`.

cuviper (approved):
Thanks!

PR #1245, opened 2025-04-10
Reduce clones in `RepeatN`

cuviper:
The standard library is optimized to avoid cloning when returning its last item. We can do the same for our version, and the savings are a lot greater because the waste was multiplied across every split instance as the parallel iterator ran. Now we avoid cloning when returning the last item of any split, so we don't hold any value at all when the count is zero.

### DIFF
--- a/src/iter/repeat.rs
+++ b/src/iter/repeat.rs
@@ -4,7 +4,7 @@ use std::iter;
 
 /// Iterator adaptor for [the `repeat()` function](fn.repeat.html).
 #[derive(Debug, Clone)]
-pub struct Repeat<T: Clone + Send> {
+pub struct Repeat<T> {
     element: T,
 }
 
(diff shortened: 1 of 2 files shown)


adamreichold (approved):
LGTM

As an aside, I wonder whether for Rayon, a variant that requires `T: Sync + Clone` and packages the element in an `Arc` would be helpful to avoid the repeated dispatch on the `::Empty` case.

cuviper:
> As an aside, I wonder whether for Rayon, a variant that requires `T: Sync + Clone` and packages the element in an `Arc` would be helpful to avoid the repeated dispatch on the `::Empty` case.

I think if we add [`repeat_with`](https://doc.rust-lang.org/nightly/std/iter/fn.repeat_with.html), you could easily roll your own to that effect, like `repeat_with(move || x.clone())`, since that would use `F: Fn + Sync` which would work based on the captured `Sync`. I guess we might also want something like `repeat_n_with` and the same kind of `take`/`zip` methods we have on `Repeat`, so you can have an `IndexParallelIterator` of this.

PR #1255, opened 2025-07-21
Adds `as_slice` and friends to basic slice and vec iterators

lucascool12:
This commit adds the methods `as_slice`, `as_mut_slice`, `into_slice` and `AsRef<[T]>` on the appropriate iterators for slice iterators and vec iterators.
This allows custom producers to use these iterators to decide on custom split points based on the remaining data in the slice.

These impls are based on the std impls on [std::slice::Iter](https://doc.rust-lang.org/std/slice/struct.Iter.html), [std::slice::IterMut](https://doc.rust-lang.org/std/slice/struct.IterMut.html), and [std::vec::IntoIter](https://doc.rust-lang.org/std/vec/struct.IntoIter.html).

### DIFF
--- a/src/slice/mod.rs
+++ b/src/slice/mod.rs
@@ -785,6 +785,28 @@ pub struct Iter<'data, T: Sync> {
     slice: &'data [T],
 }
 
+impl<'data, T: Sync> Iter<'data, T> {
+    /// Views the underlying data as a subslice of the original data.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use rayon::prelude::*;
+    /// let slice = &['a', 'b', 'c'];
+    /// let iter = slice[1..3].par_iter();
+    /// assert_eq!(iter.as_slice(), &['b', 'c']);
+    /// ```
+    pub fn as_slice(&self) -> &'data [T] {
+        self.slice
+    }
+}
+
+impl<'data, T: Sync> AsRef<[T]> for Iter<'data, T> {
+    fn as_ref(&self) -> &[T] {
+        self.as_slice()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


cuviper:
> This allows custom producers to use these iterators to decide on custom split points based on the remaining data in the slice.

Do you have an example of code where you would like to use this?

Unlike an `Iterator` where actual work is done, `ParallelIterator` is kind of a passive starting point, and nothing is actually split until we convert to producers and consumers. I'm not sure `rayon::slice::Iter` et al. are helpful to custom code, because even with `as_slice` you can't really *do* anything with them yet. I usually have custom code just hold the direct slice and work with that.

> These impls are based on the std impls on [std::slice::Iter](https://doc.rust-lang.org/std/slice/struct.Iter.html), [std::slice::IterMut](https://doc.rust-lang.org/std/slice/struct.IterMut.html), and [std::vec::IntoIter](https://doc.rust-lang.org/std/vec/struct.IntoIter.html).

That's good, we like to match precedent from the standard library -- although `IterMut::as_mut_slice` isn't stable yet.

cuviper:
I can't think of a way to do that with generic `ParallelIterator` or `IndexedParallelIterator`, but if you start with the actual slices then I think you could use [`rayon::iter::split`](https://docs.rs/rayon/latest/rayon/iter/fn.split.html) to implement this, which can be hidden in a proper `impl` like:

```rust
impl ParallelIterator for ParallelPairs {
    type Item = (Option<u64>, Option<u64>);

    where
        C: UnindexedConsumer<Self::Item>
    {
        rayon::iter::split(self, |pairs| todo!("find your split points and do it"))
            .flat_map_iter(|pairs| todo!("convert into the serial Pairs"))
            .drive_unindexed(consumer)
    }
}
```

PR #1257, opened 2025-07-24
Bump the MSRV to 1.80

cuviper:
Rust 1.80.0 was released on 2024-07-25, so this shouldn't be published until we pass that 1-year anniversary to meet our MSRV policy. Depending on your timezone though, you may already be there as I write this.

Most of the changes here are just cleaning up the implementation, not really anything noticeable to users. However, the parallel iterator for `&BinaryHeap` can now avoid creating a temporary `Vec`, which will be nice if anyone is actually using `heap.par_iter()`.


rayon-core/src/join/mod.rs
```diff
@@ -151,23 +151,21 @@ where
...
-                    // Found it! Let's run it.
-                    //
-                    // Note that this could panic, but it's ok if we unwind here.
-                    let result_b = job_b.run_inline(injected);
-                    return (result_a, result_b);
-                } else {
-                    worker_thread.execute(job);
-                }
-            } else {
+            let Some(job) = worker_thread.take_local_job() else {
```
adamreichold, line 154: Any particular reason to switch to let-else here? At least to me, it isn't immediately obvious why the new way of writing this is clearer.

cuviper: No _big_ reason, but it at least reduces indentation. I think it's a little more clear which path is going to continue the loop too.

adamreichold: Fair enough. Personally, I would have gone for a three-arm match with a guard for the job ID. But as you said, this looks like a primarily stylistic question.

The rest of changes look good to me.

PR #1276, opened 2025-11-09
Document `Drop`ping behavior of `_{with|init}` fns

Patiga:
Had to try out the dropping behavior in the playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f725ccf90f9148c937de42e4e570a515
I hope someone can chime in with developer knowledge and confirm all the initialized/cloned values are dropped before the respective functions return.
Felt the need to document it in order to save other users time :)

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -382,7 +382,8 @@ pub trait ParallelIterator: Sized + Send {
     ///
     /// The `init` value will be cloned only as needed to be paired with
     /// the group of items in each rayon job.  It does not require the type
-    /// to be `Sync`.
+    /// to be `Sync`. All cloned values get dropped before this function
+    /// returns.
     ///
     /// # Examples
     ///
@@ -413,7 +414,8 @@ pub trait ParallelIterator: Sized + Send {
     ///
     /// The `init` function will be called only as needed for a value to be
     /// paired with the group of items in each rayon job.  There is no
-    /// constraint on that returned type at all!
+    /// constraint on that returned type at all! The initialized values all
+    /// get dropped before this function returns.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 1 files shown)


cuviper:
> The `init` parameter in `for_each_with` and the `init` closure in `for_each_init` are both owned. My understanding is that the code could store the initialized values in global contexts.

They are owned, but that doesn't mean they can live forever, and since the rayon methods have no lifetime constraints, we can only assume that they're valid for the call. We would need something like `T: 'static` or `T: 'scope` to store these anywhere that would outlive the call.

As a simple example, the caller might use `T = &X` here, then that reference is only valid for a certain time. The borrow checker will determine the effective lifetime on the caller side, but this is opaque to the rayon methods.

In your own case where your code has visibility into the types involved, it's possible for you to do more advanced things.

cuviper:
Here's a tweak to your example showing a use that must not outlive the call:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e246f0aa3cdd62fabdaa876e1703ac97

```rust
use rayon::iter::{IntoParallelIterator, ParallelIterator};

struct S<'s>(&'s str);
impl Drop for S<'_> {
    fn drop(&mut self) {
        std::thread::sleep(std::time::Duration::from_millis(100));
        println!("Dropped {}", self.0)
    }
}

fn main() {
    let s = String::from("hello");
    (0..10)
        .into_par_iter()
        .for_each_init(|| S(&s), |_s, e| println!("Element {e}"));
    drop(s); // borrows **must** be released by now
    println!("Finished");
}
```

If any `S` remained after the call, it would be accessing a dangling reference from that dropped local `String`.

Patiga:
Thanks a lot for the kind and thorough explanation!
Documenting this is therefore maybe not the best action, as there would probably be many more implicit details that people might stumble over at first like me, but documenting all of them would be too much.
If a significant amount of people stumble over this specific question and issue and comment, maybe this should still be considered.
Closing for now :)

PR #1280, opened 2026-01-14
this unsafe is not needed

asuffield:


cuviper:
We haven't documented any safety requirements, but that doesn't mean it is safe. At the very least, the caller must ensure that the `JobRef`'s raw pointer remains valid until executed.

This is an internal function though, so what is your motivation for changing this?

Issue #1254, opened 2025-07-07
Implement a median / select nth algorithm.


GuillaumeDIDIER:
I need to find Q1, Q3 and the median on large containers, and parallel sorting doesn't exactly scale.

I am wondering if it would be possible to extend rayon to also have the median of median algorithm to find medians and a selection of the k-th smallest / largest element (which is the generalised primitive needed to implement most median / quartiles definitions)

(Or if there is now way of implementing this effectively to leverage parallelism)

If that would make sense, I'd be willing to work on the implementation of it and submit a PR.

cuviper:
I don't know how to implement that parallel algorithm off-hand, but it may be possible.

qkniep:
I looked into this a bit.

You should definitely consider using the sequential `select_nth_unstable()` method on slices from the standard library, which is just a wrapper around `partition_at_index()` and runs in-place in O(n) time. For most reasonable array sizes, it seems to be a significant constant factor faster (3-4x on my hardware) than parallel sorting. Thus, even when you want to calculate the three quartile boundaries (Q1, Q3, median) it's usually faster (30-50% on my hardware) than full sorting, especially when reusing previous partitions.

It doesn't seem easy to perform much better than these two options in general, since in general this problem is memory-bound and not compute-bound. However, there are some ways to achieve a speedup for well-behaved input data. I wrote a small draft implementation, which performs better (2-3x on my hardware) than sequential selection on uniform or high-cardinality data. However, it currently still depends on some dubious hand-picked magic numbers and performs worse on non-uniform low-cardinality data.

One question regarding API design would be whether it should just be single selection calls like `par_median()` / `par_select_nth()`, or whether there should be a call like `par_select_nth_multiple()` which can be called with a slice of indices, and returns the k-th element for each of the values of k. The latter could e.g. delegate to `par_sort()` or reuse previously calculated partitions if appropriate.

Whatsonyourmind:
A parallel selection algorithm is definitely feasible and would outperform parallel sort for this use case. Here's a sketch of how to parallelize the median-of-medians / quickselect approach:

**Parallel Quickselect** (expected O(n) work, O(log^2 n) depth):

1. Pick a pivot (median-of-medians or random sample)
2. **Parallel partition**: use rayon's \`par_iter\` to classify elements into < pivot, = pivot, > pivot buckets. This is the key parallelizable step -- each thread scans its chunk independently
3. Recurse into the appropriate bucket based on k

```rust
[...]

PR #1281, opened 2026-01-18
parallel median, n-th variants

qkniep:
This is a basic implementation of `par_median` and other variants for n-th element selection.

The algorithm works as follows: 1) sample a subset of roughly `sqrt(len)` elements, 2) sort that sample, 3) pick lower and upper bounds close to the `k/len` qunatile within the sample, 4) collect all elements between these bounds from the input array, 5) if the k-th element lies within the bounds or equals one of the bounds, return it. Note, that this can fail if the initial sample or bound selection was "bad". In that case we currently fall back to a sequential call to `select_nth_unstable()`.

As opposed to what I wrote in #1254, this implementation now seems to perform well even for non-uniform low-cardinality data (due to the counting of elements that equal either of the bounds).

Addresses #1254.


cuviper:
Hi -- I'm afraid I don't have time to review this in full right now, but I'll close+reopen to refresh CI approval at least.

Issue #915, opened 2022-02-09
Consider adding the ability to register custom loggers


canova:
Hi everyone, in Firefox, we were trying to record and mark the thread sleeps with the Firefox Profiler. That way, we can profile and see when a thread is starting to sleep and when it's getting awake. It's pretty easy to mark these sleeps in the C++ side since we handle these sleeps ourselves there. But it's more tricky for the Rust threads, because mostly we are using Rayon to handle these and we don't have an access to sleep code path.

So I tried to find a handler that could be called right before the sleep and right after the sleep, but I can't find an API like that on Rayon. I guess, having this API could be also useful, but it deserves its own issue (if you think it would be valuable). While I was looking at the Rayon codebase, I saw that there are some logs right [before](https://github.com/rayon-rs/rayon/blob/f45eee8fa49c1646a00f084ca78d362f381f1b65/rayon-core/src/sleep/mod.rs#L198-L201) and [after](https://github.com/rayon-rs/rayon/blob/f45eee8fa49c1646a00f084ca78d362f381f1b65/rayon-core/src/sleep/mod.rs#L234-L237) the sleep. At that moment I was kinda excited thinking that I could maybe register my own custom logger and get those notifications. But unfortunetly, [I don't see](https://github.com/rayon-rs/rayon/blob/f45eee8fa49c1646a00f084ca78d362f381f1b65/rayon-core/src/log.rs#L1-L19) any custom logging features right now in the codebase.

So, I would really love if we could have a Logger trait that people could implement and register that as a custom logger while we are building a thread pool ([for example here](https://searchfox.org/mozilla-central/rev/b697834e78a3ef7613e2fa57c07624b1d9d1c909/gfx/webrender_bindings/src/bindings.rs#1096-1110)). That way, we can have access to the thread sleeps and we can add gecko profiler markers to those. Optionally, we are thinking about adding some telemetry, so we can understand more and optimize these use cases.

nikomatsakis:
I think what I would like to do is to integrate with `tracing`-- do you think that would suit your needs, @canova ?

canova:
@nikomatsakis Yes, probably that would be a more generic way to do this. Currently we don't have a way to collect `tracing` events in the Firefox Profiler side but we do [have a bug filed for it](https://bugzilla.mozilla.org/show_bug.cgi?id=1652558). That would indeed suit us well.

nikomatsakis:
[...]

PR #1282, opened 2026-02-10
Add `tracing` integration

JoeyBF:
This PR adds optional support for the tracing crate, as mentioned in #915 (especially https://github.com/rayon-rs/rayon/issues/915#issuecomment-3100117320).

Spans (INFO level):
- rayon::worker_thread -- Wraps each worker thread's lifetime. Fields: worker, pool_id
- rayon::job_execute -- Wraps each job execution. Fields: job_id, worker

Events (DEBUG level):
- rayon::thread_idle -- Worker going idle
- rayon::thread_active -- Worker waking up
- rayon::job_injected -- Job injected into global queue. Fields: job_id, pool_id
- rayon::job_stolen -- Job stolen from another thread. Fields: job_id, victim

Jobs capture the current span context when created. When executed (potentially on a different thread), they restore this context so job_execute spans appear as children of the span that spawned the work. This enables proper distributed tracing across thread boundaries.

Furthermore, all instrumentation compiles to no-ops when the feature is disabled, `JobContext` is zero-sized when tracing is off, and field values are computed inside macros to avoid overhead when filtered out.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ rust-version.workspace = true
 #
 # Only useful on the `wasm32-unknown-unknown` target.
 web_spin_lock = ["dep:wasm_sync", "rayon-core/web_spin_lock"]
+tracing = ["rayon-core/tracing"]
 
 [dependencies]
 # These are both public dependencies!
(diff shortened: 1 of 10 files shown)


cuviper:
Hi, and thanks for working on this! I'm trying to get a release out for a different bug fix, but I don't have time to review functionality of this size right now. I will try to get to it before too long though!

JoeyBF:
I just rebased on main and added a fixup commit. Ready to review whenever you got the time!

Issue #1233, opened 2025-02-16
Updating sort algorithms


nazar-pc:
Rust library today uses [ipnsort](https://github.com/Voultapher/sort-research-rs/blob/5c75e2e92c5ee2b7e65014fff53d10e7f2db10f9/writeup/ipnsort_introduction/text.md) and [driftsort](https://github.com/Voultapher/sort-research-rs/blob/5c75e2e92c5ee2b7e65014fff53d10e7f2db10f9/writeup/driftsort_introduction/text.md) for unstable and stable sort respectively.

I'm wondering if it would make sense to switch to them in parallel implementations here too instead of pdqsort and timsort respectively.

cuviper:
Yes -- our current implementations were originally copied and adapted from the standard library, and we've synced some minor tweaks since then, but not that major rewrite yet. I'd be happy to see the new ones ported!

It might be even better if we can *just* do the parallel splits and merges on our own, and forward the rest to the standard library. I _think_ that may be feasible for the unstable sort, but last time I looked, the buffer management of stable sorting is not amenable to that kind of arrangement.

PR #1283, opened 2026-02-28
Refactor par_sort_unstable to delegate to standard library

MatiasLopezING:
As discussed in #1233, this PR delegates the sequential base cases of `par_quicksort` directly to the standard library (`slice::sort_unstable_by`), allowing Rayon to immediately benefit from the newly stabilized `ipnsort` algorithm.

**Changes made:**
* Replaced internal fallbacks in `recurse` with `v.sort_unstable_by()`.
* Removed `heapsort` and its associated tests, as it is no longer needed as a degenerate-case fallback.
* Kept `par_mergesort` untouched to avoid allocator contention issues for stable sorting (can be addressed in a future PR).

### DIFF
--- a/src/slice/sort.rs
+++ b/src/slice/sort.rs
@@ -259,51 +259,7 @@ where
     false
 }
 
-/// Sorts `v` using heapsort, which guarantees *O*(*n* \* log(*n*)) worst-case.
-#[cold]
-fn heapsort<T, F>(v: &mut [T], is_less: F)
-where
-    F: Fn(&T, &T) -> bool,
-{
-    // This binary heap respects the invariant `parent >= child`.
-    let sift_down = |v: &mut [T], mut node| {
-        loop {
-            // Children of `node`.
-            let mut child = 2 * node + 1;
-            if child >= v.len() {
-                break;
-            }
 
-            // Choose the greater child.
-            if child + 1 < v.len() {
-                // We need a branch to be sure not to out-of-bounds index,
-                // but it's highly predictable.  The comparison, however,
-                // is better done branchless, especially for primitives.
-                child += is_less(&v[child], &v[child + 1]) as usize;
-            }
-
-            // Stop if the invariant holds at `node`.
-            if !is_less(&v[node], &v[child]) {
-                break;
-            }
-
-            // Swap `node` with the greater child, move one step down, and continue sifting.
-            v.swap(node, child);
-            node = child;
-        }
-    };
-
-    // Build the heap in linear time.
-    for i in (0..v.len() / 2).rev() {
-        sift_down(v, i);
-    }
-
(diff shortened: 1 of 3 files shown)


src/slice/sort.rs
```diff
@@ -824,40 +780,35 @@ where
...
+            v.sort_unstable_by(|a, b| if is_less(a, b) { cmp::Ordering::Less } else if is_less(b, a) { cmp::Ordering::Greater } else { cmp::Ordering::Equal });
             return;
         }
 
-        // If too many bad pivot choices were made, simply fall back to heapsort in order to
+        // If too many bad pivot choices were made, simply fall back to the standard library sort in order to
         // guarantee `O(n * log(n))` worst-case.
         if limit == 0 {
-            heapsort(v, is_less);
+            v.sort_unstable_by(|a, b| if is_less(a, b) { cmp::Ordering::Less } else if is_less(b, a) { cmp::Ordering::Greater } else { cmp::Ordering::Equal });
```
adamreichold: Adapting the predicate seems error prone and is done twice. Could this be factored out and done once earlier (without touching too much of the rest of the code)?

MatiasLopezING: Thanks for the review and guidance, @adamreichold! You were completely right.

Instead of doing the error-prone boolean conversion inside recurse, I refactored par_quicksort and its callers in src/slice/mod.rs to pass the Ordering comparator directly all the way down.

I also had to remove the random comparison test in src/slice/test.rs, since delegating to modern slice::sort_unstable_by strictly enforces total order and rightfully panics on completely random comparisons.

Just pushed the updates!

cuviper:
Closing and reopening to trigger CI again.

src/slice/sort.rs
```diff
@@ -824,40 +780,38 @@ where
...
     // If both partitions are up to this length, we continue sequentially. This number is as small
     // as possible but so that the overhead of Rayon's task scheduling is still negligible.
     const MAX_SEQUENTIAL: usize = 2000;
 
     // True if the last partitioning was reasonably balanced.
-    let mut was_balanced = true;
+    let was_balanced = true;
     // True if the last partitioning didn't shuffle elements (the slice was already partitioned).
-    let mut was_partitioned = true;
+    let was_partitioned = true;
```
cuviper, line 798: These flags have lost meaning if they're always `true`. Can we do anything to maintain them through recursion? Or should we just remove more dead code?

cuviper:
Overall, I'm not sure this is worth it just to remove one function, `heapsort`. Did you run any benchmarks to show that we gain by delegating to the standard library at this granularity?

cuviper:
In other words, a fuller port of the new std algorithms might be more fruitful, though more difficult.

PR #1297, opened 2026-04-21
Make `CoreLatch::set` use `SeqCst` instead of `AcqRel`

cuviper:
Every other modification of this variable uses `SeqCst`, which is justified in the sleep README. This particular choice of `AcqRel` was not discussed during its addition in #746, nor rayon-rs/rfcs#5, so I suspect was simply an oversight from earlier development. We probably do want this to participate in the same sequential consistency.

The only other ordering difference is `CoreLatch::probe`'s load with `Acquire`, which should be fine because this doesn't need consistency with the sleep counters.

Assisted-by: Claude Code

### DIFF
--- a/rayon-core/src/latch.rs
+++ b/rayon-core/src/latch.rs
@@ -124,7 +124,7 @@ impl CoreLatch {
     /// latch code.
     #[inline]
     unsafe fn set(this: *const Self) -> bool {
-        let old_state = unsafe { (*this).state.swap(SET, Ordering::AcqRel) };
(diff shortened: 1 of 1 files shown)


cuviper:
cc @nikomatsakis in case you have any dusty memories about this.

I should also look into using some `aarch64` runners in CI, since that has weaker memory ordering than x86, which *could* make a difference here. But I think it would be quite a rare race for this one to really cause any problem. (edit: done in #1298)

PR #1301, opened 2026-04-24
Use `WorkerThread` directly in thread-local instead of a pointer

zetanumbers:
This is a simplified version of rust-lang/rust#155688 for the upstream. This PR changes `WORKER_THREAD_STATE` type to store `WorkerThread` directly in the thread-local to avoid indirection.

### DIFF
--- a/rayon-core/src/registry.rs
+++ b/rayon-core/src/registry.rs
@@ -8,7 +8,7 @@ use crate::{
     Yield,
 };
 use crossbeam_deque::{Injector, Steal, Stealer, Worker};
-use std::cell::Cell;
+use std::cell::{Cell, UnsafeCell};
 use std::fmt;
 use std::hash::{DefaultHasher, Hasher};
 use std::io;
@@ -299,10 +299,9 @@ impl Registry {
                 // Rather than starting a new thread, we're just taking over the current thread
                 // *without* running the main loop, so we can still return from here.
                 // The WorkerThread is leaked, but we never shutdown the global pool anyway.
-                let worker_thread = Box::into_raw(Box::new(WorkerThread::from(thread)));
 
                 unsafe {
-                    WorkerThread::set_current(worker_thread);
+                    WorkerThread::set_current(WorkerThread::from(thread));
                     Latch::set(&registry.thread_infos[index].primed);
                 }
                 continue;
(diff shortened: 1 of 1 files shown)


zetanumbers:
I've tried to measure it, but sadly I couldn't get reliable numbers on my machine for some reason. I'll try using my laptop instead but until then I dunno.

rayon-core/src/registry.rs
```diff
@@ -700,18 +704,45 @@ impl WorkerThread {
...
     }
 
+    /// Sets `self` as the worker-thread index for the current thread.
+    /// This is done during worker-thread startup.
+    fn with_current<F, R>(thread: WorkerThread, f: F) -> R
+    where
+        F: FnOnce(&WorkerThread) -> R,
+    {
+        WORKER_THREAD_STATE.with(|t| {
+            struct Guard(*mut Option<WorkerThread>);
```
adamreichold, line 727: Is it necessary to introduce this helper using an internal guard instead of keeping the `WorkerThread` as the guard itself?

rayon-core/src/registry.rs
```diff
@@ -700,18 +704,45 @@ impl WorkerThread {
     /// anywhere on the current thread.
     #[inline]
     pub(super) fn current() -> *const WorkerThread {
-        WORKER_THREAD_STATE.get()
+        WORKER_THREAD_STATE.with(|t| unsafe { (&*t.get()).as_ref().map_or(ptr::null(), |r| r) })
```
Zoxc, line 707: This is looks like UB as the provenance you get from the reference won't be valid outside `with`.

zetanumbers: Although I don't see any problem with the provenance, I didn't write this code to follow tree-borrows or stack-borrows. It's a temporary simplification from rust-lang/rust#155688 and it was intended to allow us to quickly measure performance impact and decide to go for it or not with more changes, as currently tree/stack-borrows would not impact this code AFAIK. And maybe there was a rule allowing this behavior in tree borrows but I'm not sure.

rayon-core/src/registry.rs
zetanumbers: Yes, it is as long as we want to keep `WorkerThread` contained in the thread-local and drop it at the end of this function.

cuviper:
Here's a microbenchmark:

```rust
#![feature(test)]
extern crate test;

#[bench]
fn global_num_threads(b: &mut test::Bencher) {
    // prime the global thread pool first
    _ = rayon::current_num_threads();
    b.iter(|| rayon::current_num_threads());
}

#[bench]
fn local_num_threads(b: &mut test::Bencher) {
    rayon::scope(|_| {
        b.iter(|| rayon::current_num_threads());
    });
}

#[bench]
fn global_thread_index(b: &mut test::Bencher) {
    assert!(rayon::current_thread_index().is_none());
    b.iter(|| rayon::current_thread_index());
}

#[bench]
fn local_thread_index(b: &mut test::Bencher) {
    rayon::scope(|_| {
        assert!(rayon::current_thread_index().is_some());
        b.iter(|| rayon::current_thread_index());
    });
}
```

I get nearly identical results on my machine (Linux 9800X3D), within noise.

PR #1305, opened 2026-06-23
Implement `HashMap<K, Vec<V>>` Transpose

### LINKED ISSUE #1304: Transpose `HashMap<K, Vec<T>>` to `Vec<HashMap<K, T>>`
What would be the best way to achieve this with rayon? HashMap is small ~60 and Vec is large ~60000 and K is Copy. My first thought was to implement the IntoParallelIterator and Producer traits manually with a wrapper type and use a HashMap of rayon::vec::DrainProducers for splitting, but DrainProducer is not in the public interface.  Am I missing something obvious?  I'm also looking to solve this for transposing a struct-of-arrays to array-of-structs.

Edit:
I couldn't think of any other way to do it, but my I wrote an implementation and my idea works!

krakow10:
Closes #1304.

I wrote some code to implement the operation, but it kind of sucks.  I'd appreciate feedback on how to make an upstreamable version of this idea.

As a side note, I'm looking at how every callsite for DrainProducer::from_vec calls set_len and I'm thinking that should be part of the function.

### DIFF
--- a/src/transpose.rs
+++ b/src/transpose.rs
@@ -58,7 +58,15 @@ where
 
     fn with_producer<CB: ProducerCallback<Self::Item>>(mut self, callback: CB) -> CB::Output {
         // Create the producer as the exclusive "owner" of the slice.
-        let producer = TransposeProducer::from_transpose(&mut self);
+        let producers = self
+            .vec
+            .into_iter()
+            .map(|iter| iter.into_par_iter().with_producer(callback))
+            .collect();
+        let producer = TransposeProducer {
+            producers,
+            len: self.len,
+        };
 
         // The producer will move or drop each item from the drained range.
         callback.callback(producer)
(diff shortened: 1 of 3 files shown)


cuviper:
Can we make this more generic? I'm thinking roughly:

```rust
fn transpose<P>(iter: impl IntoIterator<Item = P>) -> Transpose<P>
where
    P: IntoParallelIterator,
    P::Iter: IndexParallelIterator,
{
    todo!()
}

impl<P> ParallelIterator for Transpose<P> {
    type Item = TransposeIterator<P>; // impl Iterator<Item = P::Item>
    ...
}

impl<P> IndexedParallelIterator for Transpose<P> { ... }
```

Then your specific case would be something like:

```rust
transpose(map_to_vec.into_iter().map(|(key, vec)| {
    // produce P as a parallel Map to (K, V)
    vec.into_par_iter().map(|item| (key, item))
}))
.map(HashMap::from_iter)
.collect::<Vec<_>>()
```

krakow10:
I was hoping to figure out a way to make it more generic, that sounds like a great idea.  Understanding how to implement the rayon traits for the first time has been an uphill battle, I'll see what I can do.

Edit:
The main issue is that I don't know how to get the Producer associated with the IndexedParallelIterator, which I assume is sealed by design.  This implementation would need to store `Vec<Producer>` to be fully generic, which requires naming the Producer in a trait bound.

src/transpose.rs
```diff
@@ -61,7 +61,10 @@ where
         let producers = self
             .vec
             .into_iter()
-            .map(|iter| iter.into_par_iter().with_producer(callback))
+            .map(|iter| {
+                // I don't know how to get the Producer from the iter here.
+                iter.into_par_iter().with_producer(callback)
```
cuviper, line 66: Ah, yeah that's tough, maybe not possible. Even though we've constrained them all to the same iterator type, in theory they can use different producer types, e.g. optimized for certain edge cases. `Producer` isn't a `dyn`-compatible trait either, so we can't easily abstract this away either.

The `plumbing/README.md` talks a bit about why we have this `ProducerCallback` scheme. Maybe now that we *do* have associated type constructors (now called GATs, generic associated types) it could be possible to improve this. That's a bigger change, and breaking semver too -- maybe worth the bump though if it helps cases like this!

PR #1308, opened 2026-07-11
Delegate parallel unstable sort leaves to the standard library

### LINKED ISSUE #1233: Updating sort algorithms

Harikeshav-R:
Closes #1233.

## What this does

rayon's parallel unstable sort (`par_sort_unstable` and friends) is a copy of the old
`core::slice::sort` quicksort, including a hand-written `heapsort` used as the worst-case
fallback. The standard library has since moved its own unstable sort to `ipnsort`, so this
copied code is now behind.

This change makes rayon hand off the small, sequential parts of the sort to the standard
library's `slice::sort_unstable_by` (which is `ipnsort`), while rayon keeps doing all the
parallel work itself. Two spots change:

- the base case for short slices
- the fallback that runs when too many bad pivots have been picked (this used `heapsort` before)

Everything that makes the sort parallel and pattern-defeating stays the same: pivot choice,
block partitioning, the `was_balanced` / `was_partitioned` tracking, the cutoff for going
sequential, and the `join`. A small new `sort_leaf` helper does the `bool` to `Ordering`
conversion in one place so it isn't repeated.

`heapsort` and its test are removed because nothing uses them anymore.

## Why do it this way

There was an earlier attempt in #1283. It also removed the sequential branch and always called
`join`, which meant `was_balanced` and `was_partitioned` were always `true` and no longer did
anything. That quietly turned off the pattern-defeating behavior, which is what @cuviper pointed
out in the review.

Here only the leaves change, so those flags and the rest of the algorithm still do their job and
the `O(n log n)` worst case is kept (`ipnsort` is also `O(n log n)` worst case, same as the old
`heapsort`). Because `par_quicksort` keeps taking a `Fn(&T, &T) -> bool`, nothing in
`src/slice/mod.rs` or the public API needs to change.

## One behavior change to be aware of

The modern `sort_unstable_by` panics if the comparison function isn't a valid total order. The
old fallback just produced some arbitrary order instead. I checked this: a random
`Less/Equal/Greater` comparator panics on a 100 element slice but not on a slice of 20 or fewer,
since the small-sort code tolerates it.

The shared sort test used to say the random comparator "won't panic". It now wraps that call in
`catch_unwind` and still checks that no elements were lost (sorting again with a real comparator
has to give back `0..100`). That is the property that actually matters for safety.

## Benchmarks

I ran `cargo +nightly bench -p rayon-demo par_sort_unstable`. My machine is WSL2 and pretty
noisy, so single runs jump around a lot. Taking the best of 5 runs of
`par_sort_unstable_random` (50k `u64`) as the cleanest estimate:

| | min ns/iter |
|---|---|
| main | ~1,635,000 |
| this branch | ~1,461,000 |

The branch was consistently a bit faster and much steadier (1.46 to 1.51M, versus 1.63 to 3.78M
on main). I would read this as "no regression, maybe a small win" rather than a firm number, and
a run on quieter hardware would be welcome. The main point of the change is really maintenance:
dropping a copy of superseded std code and getting future `ipnsort` improvements for free.

## Left for later

The stable sort (`par_mergesort`) is not touched. Switching its per-chunk sort to `driftsort`
needs some thought about per-call allocation versus rayon's single shared buffer, so that is
better as a separate PR.

## Testing

- `cargo test -p rayon` passes, including `test_par_sort_unstable`, `test_par_sort`, and
  `test_par_sort_stability`.
- `cargo clippy --all-targets` is clean.
- `cargo test --doc` passes.

### DIFF
--- a/src/slice/mod.rs
+++ b/src/slice/mod.rs
@@ -642,8 +642,8 @@ pub trait ParallelSliceMut<T: Send> {
     /// (i.e., does not allocate), and *O*(*n* \* log(*n*)) worst-case.
     ///
     /// The comparator function must define a total ordering for the elements in the slice. If
-    /// the ordering is not total, the order of the elements is unspecified. An order is a
-    /// total order if it is (for all `a`, `b` and `c`):
+    /// the ordering is not total, the sort may panic; if it returns, the order of the elements is
+    /// unspecified. An order is a total order if it is (for all `a`, `b` and `c`):
     ///
     /// * total and antisymmetric: exactly one of `a < b`, `a == b` or `a > b` is true, and
     /// * transitive, `a < b` and `b < c` implies `a < c`. The same must hold for both `==` and `>`.
--- a/src/slice/sort.rs
+++ b/src/slice/sort.rs
@@ -259,50 +259,28 @@ where
     false
 }
 
-/// Sorts `v` using heapsort, which guarantees *O*(*n* \* log(*n*)) worst-case.
-#[cold]
-fn heapsort<T, F>(v: &mut [T], is_less: F)
+/// Sequentially sorts `v`, used for the leaves of the parallel recursion.
+///
+/// This delegates to the standard library's [`sort_unstable_by`](slice::sort_unstable_by) (an
+/// `ipnsort` implementation) rather than reimplementing a base-case sort. The `is_less` predicate
+/// is turned into a total [`Ordering`](cmp::Ordering) exactly once here, so the conversion is not
+/// duplicated across call sites.
+///
+/// Note that `sort_unstable_by` panics if `is_less` does not implement a total order, whereas the
+/// old hand-written fallback would merely produce an unspecified permutation.
+fn sort_leaf<T, F>(v: &mut [T], is_less: &F)
 where
     F: Fn(&T, &T) -> bool,
 {
-    // This binary heap respects the invariant `parent >= child`.
-    let sift_down = |v: &mut [T], mut node| {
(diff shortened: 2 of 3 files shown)


U0001F3A2:
One public-doc change is still needed. `par_sort_unstable_by` says a non-total comparator only gives an unspecified order, but this branch can now panic after delegating to `slice::sort_unstable_by`.

Could we say it may panic and, if it returns, the order is unspecified? That matches std's contract.

I ran `cargo test -p rayon --lib --offline` and `cargo test -p rayon --doc --offline` on `8bf2736`. Both pass.

Harikeshav-R:
Done in 9d9282e. `par_sort_unstable_by` now says the sort may panic on a non-total order and that the order is only unspecified if it returns.

I left `par_sort_unstable` and `par_sort_unstable_by_key` alone since their `T: Ord` / `K: Ord` bounds guarantee a total order, so a panic isn't reachable there, and the stable methods are unaffected.

PR #1309, opened 2026-07-15
Clarify `in_place_scope` docs on thread-pool selection

U0001F3A2:
Clarifies #1165. Tasks spawned through the scope handle use the scope's pool. Ambient Rayon operations inside the closure, including free join, free spawn, and parallel iterators, do not inherit that pool: they use the caller's current worker pool or the global pool when the caller is outside Rayon. Explicit ThreadPool operations select their own pool. Applies equally to in_place_scope_fifo. Docs-only; no behavior change.


cuviper (approved):
Looks good, thanks!

PR #1310, opened 2026-07-15
Document order preservation in parallel iterators

### LINKED ISSUE #669: Document order preservation
In the docs it doesn't say anything about whether par_iter()s preserve the original order. #551 says it does, but it would be nice to find this somewhere in the docs, maybe on the page discussing iterators in general.

U0001F3A2:
Fixes #669. Documents when parallel iterators preserve source order: ordered sources plus order-preserving adaptors and collection into Vec produce the sequential output sequence, provided closures do not depend on scheduling. The doctest covers map and filter. It also distinguishes find_any from find_first/find_last, and take_any/skip_any (including their while variants) from order-preserving forms; retained items still keep their relative order. Documents unordered sources, ParallelBridge, reductions, and side-effect scheduling caveats. Docs-only; no behavior change.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -62,7 +62,7 @@
 //!
 //! For parallel iterators over data with a meaningful order, order-preserving
 //! adaptors -- such as [`map`], [`filter`], [`enumerate`], and [`zip`] -- retain
-//! the source order when their results are collected into a [`Vec`]. Provided
+//! the source order when their results are collected into a [`Vec`], for example. Provided
 //! their closures do not depend on how work is scheduled, this produces the
 //! same sequence that the equivalent sequential [`Iterator`] would, even
 //! though the work is split across threads:
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -58,6 +58,55 @@
 //! check out the [`ParallelIterator`] and [`IndexedParallelIterator`]
 //! traits.
 //!
+//! # Ordering
+//!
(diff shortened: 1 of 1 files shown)


src/iter/mod.rs
```diff
@@ -58,6 +58,69 @@
 //! check out the [`ParallelIterator`] and [`IndexedParallelIterator`]
 //! traits.
 //!
+//! # Ordering
```
cuviper, line 81: The placement of this section will make it seem to contain the original text that follows as well. Either we should move the new text down after all that, or we should give separate headers to the other content (maybe for content above here too).

src/iter/mod.rs
```diff
@@ -58,6 +58,69 @@
...
+//!
+//! For parallel iterators over data with a meaningful order, order-preserving
+//! adaptors -- such as [`map`], [`filter`], [`enumerate`], and [`zip`] -- retain
+//! the source order when their results are collected into a [`Vec`]. Provided
```
```suggestion
//! the source order when their results are collected into a [`Vec`], for example. Provided
```

... or something like that, because it's certainly not the only ordered collection, nor is `collect` the only way to observe order.

src/iter/mod.rs
U0001F3A2: Thanks for the input. I'l move this down towards the end

src/iter/mod.rs
U0001F3A2: OK, this sounds good

U0001F3A2:
@cuviper anything else I should handle/address?

PR #1312, opened 2026-07-18
Add `IndexedParallelIterator::cartesian_product`

U0001F3A2:
Adds `IndexedParallelIterator::cartesian_product`, closing #754 and reviving the idea in #1182.

Every element of `self` is paired with every element of `other`, in row-major order. Because both inputs are indexed, the product has a known length (`self.len() * other.len()`) and is itself an `IndexedParallelIterator` — so it supports `collect_into_vec` preallocation (the actual ask in #754) and splits freely for full parallelism.

This differs from the stalled #1182, which built an *unindexed* product via `flat_map(|x| repeat(x).zip(other.clone()))`. That isn't indexed (no preallocation), and — as @wagnerf42 noted there — a `flat_map`-based product limits how the work can be split. Here the product is a single flat-index producer: index `k` maps to `(i[k / n], j[k % n])`, so it splits arbitrarily.

**Tradeoff:** both inputs are buffered (`O(i.len() + j.len())`) so their elements can be paired by index in parallel; this drives both inputs to completion before producing any pair. Documented on the method.

**Tests:** order vs a sequential reference for all sizes in `0..7`; indexed behavior (`len`/`count`/`collect_into_vec`/`enumerate`/`rev`/`skip`+`take`); empty inputs; non-`Copy` `String` items; a 500×300 parallel sum. `cargo test`, the doctest, `fmt`, and `clippy` are all green.

**Design note:** placed on `IndexedParallelIterator` (like `zip`), so both inputs must be indexed — e.g. `u64` ranges are not (the same limitation as `zip`). Happy to adjust placement or naming.

### DIFF
--- a/src/iter/cartesian_product.rs
+++ b/src/iter/cartesian_product.rs
@@ -1,13 +1,14 @@
 use super::plumbing::*;
 use super::{IndexedParallelIterator, ParallelIterator};
+use std::sync::Arc;
 
-/// `CartesianProduct` is an iterator that iterates over the cartesian product
-/// of the elements of two parallel iterators.
+/// `CartesianProduct` is an iterator that pairs every element of one parallel
+/// iterator with every element of another, in row-major order.
 ///
 /// This struct is created by the [`cartesian_product()`] method on
-/// [`IndexedParallelIterator`].
+/// [`ParallelIterator`].
 ///
-/// [`cartesian_product()`]: IndexedParallelIterator::cartesian_product
+/// [`cartesian_product()`]: ParallelIterator::cartesian_product
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
 #[derive(Debug, Clone)]
 pub struct CartesianProduct<I, J> {
@@ -15,28 +16,17 @@ pub struct CartesianProduct<I, J> {
     j: J,
 }
 
-impl<I, J> CartesianProduct<I, J>
-where
-    I: IndexedParallelIterator,
-    J: IndexedParallelIterator,
-{
+impl<I, J> CartesianProduct<I, J> {
     /// Creates a new `CartesianProduct` iterator.
     pub(super) fn new(i: I, j: J) -> Self {
         CartesianProduct { i, j }
-    }
-
-    fn total_len(&self) -> usize {
-        self.i
-            .len()
-            .checked_mul(self.j.len())
-            .expect("cartesian_product length overflows usize")
     }
 }
 
 impl<I, J> ParallelIterator for CartesianProduct<I, J>
 where
-    I: IndexedParallelIterator,
-    J: IndexedParallelIterator,
+    I: ParallelIterator,
+    J: ParallelIterator,
     I::Item: Clone + Sync,
     J::Item: Clone + Sync,
 {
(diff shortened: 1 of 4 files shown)


src/iter/cartesian_product.rs
```diff
@@ -0,0 +1,201 @@
...
+                i: &[],
+                j: &[],
+                range: 0..0,
+            });
+        }
+        // Buffer both inputs so their elements can be paired by index in
+        // parallel. This is an `O(i.len() + j.len())` allocation, and drives both
+        // inputs to completion before any pair is produced.
+        let i: Vec<I::Item> = self.i.collect();
+        let j: Vec<J::Item> = self.j.collect();
```
adamreichold: Since it fully buffers both inputs, does it need to live inside Rayon?

```rust
let i = i.collect::<Vec<_>>();
let j = j.collect::<Vec<_>>();
let n = i.len() * j.len();

(0..n).into_par_iter().map(|k| (&i[k / n], &j[k % n]));
```

I also don't think the input iterators actually need to be indexed (or even parallel) iterators if you are buffering?

U0001F3A2: Thanks for the review, and I agree to your input. Once both sides are buffered this is just collect + collect + an indexed map over `0..i.len()*j.len()`, so it's plain user composition and doesn't really belong in rayon as-is.

Should I rework it into a version that earns a place, or just close it? A rework
would:
- move the method to ParallelIterator, so unindexed inputs like u64 ranges work (the case in #754 this PR can't compile)
- keep a known length via opt_len, so collect still preallocates (the actual ask in #754)
- stay indexed only when both inputs already are
- buffer just the inner side instead of both, to cut peak memory

adamreichold: > buffer just the inner side instead of both, to cut peak memory

Personally, I think either this or the clone-the-iterator-itself approach from #1182 are the most sensible to having inside the library.

U0001F3A2: Great, I'll go with the inner-only version then.

I prototyped it for a bit with small PoCs and it holds up. The outer stays streaming, only the inner is buffered into an Arc<[_]>, and a count-exact bridge splits the consumer at `k * inner_len` so ordinary collect goes through the preallocated path with O(inner) auxiliary buffering (the output Vec aside). Correct across sizes 0..=6 in a 4-thread pool, with exact opt_len.

I'd pick it over the #1182 clone approach since that one as submitted is unindexed (opt_len = None), so it can't preallocate, which was the actual #754 ask (and flat_map is more expensive than flat_map_iter, as noted there). The cost is more plumbing, and the exact inner-only drive needs an outer whose length is known (ranges, indexed, or exact adaptors over them), so an opaque `opt_len = None` outer falls back to buffering. `collect_into_vec` still buffers both, since a flat split can cut mid-row.

I'll push a reworked version here soon. Thanks for the input.

U0001F3A2:
Pushed the reworked version.

It's on ParallelIterator now and buffers only the inner side. `self` streams through a count-exact consumer (an outer split at k maps to an output split at `k * other.len()`), so a tracking-allocator test shows auxiliary memory stays flat (11.7 KB at 500k outer, 80 B at 4M outer) while opt_len still drives preallocation. When both inputs are indexed it stays indexed `collect_into_vec`, `enumerate`, `rev`, ...), and that path buffers both since a flat split can cut mid-row (also shown in the memory test: ~8 MB more than plain collect at 1M).

Two things I got wrong earlier, both corrected by measuring:
- An opaque (opt_len == None) outer is also streamed, not buffered. It just doesn't preallocate.
- Inner-only isn't limited to range/indexed outers. Any exact outer works, e.g. a map over a range preallocates too.

One real tradeoff to flag: parallelism comes from splitting `self`, and each outer element expands across `other` sequentially. A tiny `self` with a huge `other` (say 1 x 100k) runs on a single worker, where the buffer-both version was fully parallel. That's the small-product case @wagnerf42 raised on #1182. I documented it (put the larger input first, or use `collect_into_vec` on indexed inputs for full flat parallelism). If you'd rather not have that cliff, the way out is a two-axis producer that splits the inner once the outer can't split further. 

Happy to add that as a follow-up, otherwise this is ready for review.

U0001F3A2:
@adamreichold any advice to push this forward / close it?

Issue #642, opened 2019-03-05
Rayon uses a lot of CPU when there's not a lot of work to do


jrmuizel:
The following program uses about 30% of the CPU on 4 core (8 HT) Linux and Mac machines
```rust
fn main() {
    while true {
        std::thread::sleep(std::time::Duration::from_millis(10));
        rayon::spawn(move || {  } );
    }
}
```

Reducing the sleep duration to 1ms pushes the CPU usage up to 200%.

cuviper:
See also #576 and #614.

jrmuizel:
A profile on macOS shows most of the time being spent in sched_yield

jrmuizel:
Also, a similar program written using the `dispatch` crate uses about 3-4% cpu at the 1ms duration vs the 200% that rayon uses.

jrmuizel:
@lukewagner here's an example of some of the bad.

jrmuizel:
Rewriting this using crossbeam-channel produces an executable that only uses 4% cpu. 

```rust
    let (s, r) = unbounded();

    for i in 0..8 {
        let r = r.clone();
        thread::spawn( move || loop {
            let m: i32 = r.recv().unwrap();
        });
    }

    while true {
        std::thread::sleep(std::time::Duration::from_millis(1));
        s.send(0);
    }
```

Perhaps rayon should use the same wakeup infrastructure that crossbeam-channel uses?

ishitatsuyuki:
From https://github.com/rayon-rs/rayon/blob/master/rayon-core/src/sleep/README.md, it seems rayon will repetitively yield to the operating system before a thread goes to sleep. How does this work wrt to syscall overhead? I think yielding to the scheduler is almost as expensive as blocking on a condition variable?

cuviper:
> I think yielding to the scheduler is almost as expensive as blocking on a condition variable?

The net performance of blocking has to include the cost of the syscall to wake it up too. Even if the raw cost were the same, there's also a latency difference in fully blocking.

The `life` demo added in #603 shows this trade-off pretty well. Currently I get:

```
$ cargo run --release -- life bench
    Finished release [optimized] target(s) in 0.05s
     Running `/home/jistone/rust/rayon/target/release/rayon-demo life bench`
  serial:  114948948 ns
parallel:   32975762 ns -> 3.49x speedup
par_bridge:  631383430 ns -> 0.18x speedup

$ cargo run --release -- life play
    Finished release [optimized] target(s) in 0.05s
[...]

PR #1314, opened 2026-08-03
rayon-core: bound the number of concurrently searching threads

josefbacik:
Disclosure: I work for Anthropic, this is all data from inside of Anthropic, the code and changelog was written by Claude. I've read the code and made sure it was clean and kept with the feel of the surrounding code an the comments weren't overbearing. The body of this PR is written by me, with the exception of the last paragraph summarizing the performance numbers. I don't have an agent hooked up to this thread, all of the responses you get will be from me.

I've been doing a lot of CPU tracing in our fleet with my tracing tool (https://github.com/josefbacik/systing) and I've found that we're burning a ton of CPUs looking for work to steal in Rayon threads.

The problem here is that most people go "I need to do this work in parallel, let me use rayon", and then we get NR_CPUS sized thread pools for a job that doesn't actually have that much parallelization.

There's not a good way for application developers to right-size their thread pools, you basically need something like systing or perfetto to watch what your application is doing and figure out the real parallelization of your application, otherwise you end up burning a massive amount of compute looking for work that doesn't exist.

Go and Tokio solve this by having a maximum number of searchers that will spin looking for work to steal. This makes it so application developers don't need to be kernel engineers to figure out what their application is doing. It gives a nice trade off of latency and burning idle CPUs.

This isn't just a efficiency thing either, burning idle CPUs has a pretty big impact on scheduler latencies. This is especially true for AI/networking/IO related workloads that depend on super low scheduler latencies from IRQ->thread wakeup. If you've got a bunch of Rayon threads stressing the CPU scheduler out with wakeups you end up seeing the wakeup latency push up and up, which has a big cascade effect in things like AI training workloads.

The solution is to add a maximum searchers mechanism, via ThreadPoolBuilder::max_searchers(k), with RAYON_MAX_SEARCHERS as an environment fallback, bounding how many idle workers may spin concurrently.

This gives application developers pretty good flexibility. I would argue that Rayon would probably do well to default to max_searchers(2), but I'm not going to advocate for that change here. This at least lets me go change all of our applications to use this method, instead of arbitrarily cutting down the workpool sizes and hoping it doesn't affect latencies.

Measured on a dedicated 180-CPU host with a 180-thread pool running a memory-bound fold every 5 ms (a production workload shape): unbounded burns 37.2 CPU-cores continuously vs 4.5 with max_searchers(2) (-88%), and p50 latency of the parallel section improves 1.49 ms -> 0.88 ms, because the spinning workers' steal and epoch traffic was slowing the threads doing real work. Bounds of 1-4 perform equivalently. The rayon-core and rayon test suites pass with and without the bound.

### DIFF
--- a/rayon-core/src/lib.rs
+++ b/rayon-core/src/lib.rs
@@ -145,6 +145,19 @@ enum ErrorKind {
     IOError(io::Error),
 }
 
+/// How idle worker threads search (spin) for work before blocking.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub(crate) enum SpinPolicy {
+    /// Per-worker adaptive spinning (the default): spin on the next idle
+    /// episode only if the previous one found work before sleeping.
+    Adaptive,
+    /// At most one idle worker spins per currently-executing worker
+    /// (the producers of stealable work).
+    ProducerBounded,
+    /// Every idle worker may spin (the historical behavior).
+    Unbounded,
+}
+
 /// Used to create a new [`ThreadPool`] or to configure the global rayon thread pool.
 /// ## Creating a ThreadPool
 /// The following creates a thread pool with 22 threads.
@@ -167,6 +180,9 @@ pub struct ThreadPoolBuilder<S = DefaultSpawn> {
     /// If zero will use the RAYON_NUM_THREADS environment variable.
     /// If RAYON_NUM_THREADS is invalid or zero will use the default.
     num_threads: usize,
+
+    /// How idle workers spin searching for work before blocking.
+    spin_policy: SpinPolicy,
 
     /// The thread we're building *from* will also be part of the pool.
     use_current_thread: bool,
@@ -222,6 +238,7 @@ impl Default for ThreadPoolBuilder {
     fn default() -> Self {
         ThreadPoolBuilder {
             num_threads: 0,
+            spin_policy: SpinPolicy::Adaptive,
             use_current_thread: false,
             panic_handler: None,
             get_thread_name: None,
@@ -434,6 +451,7 @@ impl<S> ThreadPoolBuilder<S> {
             spawn_handler: CustomSpawn::new(spawn),
             // ..self
             num_threads: self.num_threads,
+            spin_policy: self.spin_policy,
             use_current_thread: self.use_current_thread,
             panic_handler: self.panic_handler,
             get_thread_name: self.get_thread_name,
@@ -524,6 +542,53 @@ impl<S> ThreadPoolBuilder<S> {
     /// be preferred.
     pub fn num_threads(mut self, num_threads: usize) -> Self {
         self.num_threads = num_threads;
+        self
+    }
+
+    /// Bounds how many idle threads may concurrently *search* (spin
+    /// looking for work to steal) to the number of currently-executing
+    /// workers, instead of the default per-worker adaptive spinning.
+    ///
+    /// Stealable work can only be produced by a worker that is currently
+    /// executing a job -- externally injected jobs wake a thread
+    /// explicitly instead -- and each such producer owns exactly one
+    /// deque a searcher could steal from. Searchers beyond the producer
+    /// count are therefore guaranteed to sweep deques nothing can
+    /// refill. This policy enforces that pairing: as a pool drains it
+    /// tapers to a single searcher instead of paying a pool-width spin
+    /// sweep, and while the pool is busy every idle worker may spin.
+    /// There are no tuned thresholds; the budget follows the pool's own
+    /// activity.
+    ///
+    /// Compared to the default this recovers substantially more idle CPU
+    /// in periodic and mixed workloads (benchmarks: pool CPU -49% idle
+    /// and -25% at 20% duty versus unbounded, where the default achieves
+    /// -43% and -5%). The cost is under full saturation -- a continuous
+    /// closed-loop stream of small independent jobs measures ~20% lower
+    /// throughput, because workers denied a searcher slot pay sleep/wake
+    /// round-trips for work that arrives microseconds later. Choose this
+    /// policy for pools whose workloads are known to have idle gaps;
+    /// keep the default for pools that must handle sustained streams of
+    /// small independent jobs at peak rate.
+    pub fn bounded_searchers(mut self) -> Self {
+        self.spin_policy = SpinPolicy::ProducerBounded;
+        self
+    }
+
+    /// Allows every idle thread to search (spin) concurrently, restoring
+    /// rayon's historical behavior.
+    ///
+    /// With N workers each sweeping all N deques per spin round, the CPU
+    /// cost of an idle episode grows roughly quadratically with pool
+    /// width, which for wide pools with frequent small parallel sections
+    /// can dominate the pool's CPU usage; the default adaptive policy
+    /// avoids that without measurable cost elsewhere. This escape hatch
+    /// exists for workloads that measure a regression under the default
+    /// -- the known candidate being bursts arriving at a fully sleeping
+    /// pool, where unbounded ramp-up saves on the order of tens of
+    /// microseconds.
+    pub fn unbounded_searchers(mut self) -> Self {
+        self.spin_policy = SpinPolicy::Unbounded;
         self
     }
 
@@ -788,6 +853,7 @@ impl<S> fmt::Debug for ThreadPoolBuilder<S> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let ThreadPoolBuilder {
             ref num_threads,
+            ref spin_policy,
             ref use_current_thread,
             ref get_thread_name,
             ref panic_handler,
@@ -813,6 +879,7 @@ impl<S> fmt::Debug for ThreadPoolBuilder<S> {
 
         f.debug_struct("ThreadPoolBuilder")
             .field("num_threads", num_threads)
+            .field("spin_policy", spin_policy)
             .field("use_current_thread", use_current_thread)
             .field("get_thread_name", &get_thread_name)
             .field("panic_handler", &panic_handler)
--- a/rayon-core/src/registry.rs
+++ b/rayon-core/src/registry.rs
@@ -651,6 +651,15 @@ pub(super) struct WorkerThread {
     /// the "stealer" half of the worker's broadcast deque
     stealer: Stealer<JobRef>,
 
+    /// Adaptive spin: whether this worker's next idle episode gets the
+    /// yield/steal spin rounds. Set from how the previous episode ended --
+    /// work found before sleeping re-arms spinning, an episode that had to
+    /// block disarms it. Under load searches succeed and every worker
+    /// spins (the historical behavior); in idle or periodic workloads
+    /// spinning self-extinguishes, and re-arms as soon as a search finds
+    /// work again. Purely worker-local: no shared state.
+    spin_next: Cell<bool>,
+
     /// local queue used for `spawn_fifo` indirection
     fifo: JobFifo,
 
@@ -678,6 +687,7 @@ impl From<ThreadBuilder> for WorkerThread {
             stealer: thread.stealer,
             fifo: JobFifo::new(),
             index: thread.index,
+            spin_next: Cell::new(true),
             rng: XorShift64Star::new(),
             registry: thread.registry,
         }
@@ -793,10 +803,14 @@ impl WorkerThread {
                 continue;
             }
 
-            let mut idle_state = self.registry.sleep.start_looking(self.index);
+            let mut idle_state = self
+                .registry
+                .sleep
+                .start_looking(self.index, self.spin_next.get());
             while !latch.probe() {
                 if let Some(job) = self.find_work() {
                     self.registry.sleep.work_found();
+                    self.spin_next.set(!idle_state.slept);
                     unsafe { self.execute(job) };
                     // The job might have injected local work, so go back to the outer loop.
                     continue 'outer;
@@ -810,6 +824,7 @@ impl WorkerThread {
             // If we were sleepy, we are not anymore. We "found work" --
             // whatever the surrounding thread was doing before it had to wait.
             self.registry.sleep.work_found();
+            self.spin_next.set(!idle_state.slept);
             break;
         }
 
--- a/rayon-core/src/registry.rs
+++ b/rayon-core/src/registry.rs
@@ -245,6 +245,8 @@ impl Registry {
 
         let breadth_first = builder.get_breadth_first();
 
+        let spin_policy = builder.spin_policy;
+
         let (workers, stealers): (Vec<_>, Vec<_>) = (0..n_threads)
             .map(|_| {
                 let worker = if breadth_first {
@@ -268,7 +270,7 @@ impl Registry {
 
         let registry = Arc::new(Registry {
             thread_infos: stealers.into_iter().map(ThreadInfo::new).collect(),
-            sleep: Sleep::new(n_threads),
+            sleep: Sleep::new(n_threads, spin_policy),
             injected_jobs: Injector::new(),
             broadcasts: Mutex::new(broadcasts),
             terminate_count: AtomicUsize::new(1),
@@ -809,7 +811,7 @@ impl WorkerThread {
                 .start_looking(self.index, self.spin_next.get());
             while !latch.probe() {
                 if let Some(job) = self.find_work() {
-                    self.registry.sleep.work_found();
+                    self.registry.sleep.work_found(&mut idle_state);
                     self.spin_next.set(!idle_state.slept);
                     unsafe { self.execute(job) };
                     // The job might have injected local work, so go back to the outer loop.
@@ -823,7 +825,7 @@ impl WorkerThread {
 
             // If we were sleepy, we are not anymore. We "found work" --
             // whatever the surrounding thread was doing before it had to wait.
-            self.registry.sleep.work_found();
+            self.registry.sleep.work_found(&mut idle_state);
             self.spin_next.set(!idle_state.slept);
             break;
         }
--- a/rayon-core/src/sleep/mod.rs
+++ b/rayon-core/src/sleep/mod.rs
@@ -1,10 +1,11 @@
 //! Code that decides when workers should go to sleep. See README.md
 //! for an overview.
 
+use crate::SpinPolicy;
 use crate::latch::CoreLatch;
 use crate::sync::{Condvar, Mutex};
 use crossbeam_utils::CachePadded;
-use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::thread;
 
 mod counters;
@@ -24,6 +25,18 @@ pub(super) struct Sleep {
     worker_sleep_states: Vec<CachePadded<WorkerSleepState>>,
 
     counters: AtomicCounters,
+
+    /// How idle workers spin searching for work (see [`SpinPolicy`]).
+    policy: SpinPolicy,
+
+    /// Number of idle workers currently in the yield/steal spin rounds.
+    /// Maintained only under `SpinPolicy::ProducerBounded`: at most one
+    /// searcher per currently-executing worker (each producer owns one
+    /// deque a searcher could steal from), minimum one. Workers beyond
+    /// the budget skip straight to the sleepy protocol (announce via the
+    /// JEC, one final search, then block), so a draining pool tapers its
+    /// searchers with its producers instead of paying pool-width sweeps.
+    searchers: AtomicUsize,
 }
 
 /// An instance of this struct is created when a thread becomes idle.
@@ -47,6 +60,10 @@ pub(super) struct IdleState {
     /// an episode that found work before sleeping means spinning pays off
     /// here; an episode that slept means it did not.
     pub(super) slept: bool,
+
+    /// Whether this idle thread holds one of the bounded searcher slots
+    /// (only meaningful under `SpinPolicy::Bounded`).
+    is_searcher: bool,
 }
 
 /// The "sleep state" for an individual worker.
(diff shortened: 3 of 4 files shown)


rayon-core/src/lib.rs
```diff
@@ -527,6 +551,35 @@ impl<S> ThreadPoolBuilder<S> {
...
+    /// Setting a bound `k` allows at most `k` idle workers to spin
+    /// concurrently; the rest skip the spin phase and block immediately
+    /// (after the usual sleep protocol, so no wakeups are lost). The
+    /// spinning workers still pick up new work with the same latency, and
+    /// they wake their sleeping peers as work fans out. Small bounds such
+    /// as 1-4 retain wake latency while eliminating most idle CPU cost;
+    /// this mirrors the bounded-spinner designs of other work-stealing
+    /// schedulers (Go's runtime, Tokio).
+    ///
+    /// If `max_searchers` is 0, or you do not call this function, the
```
adamreichold: From the description alone, it sounds like zero would/should be valid value to use here, meaning "all workers will immediately take the blocking path disabling the spinning completely".

adamreichold:
> This gives application developers pretty good flexibility. I would argue that Rayon would probably do well to default to max_searchers(2), but I'm not going to advocate for that change here. This at least lets me go change all of our applications to use this method, instead of arbitrarily cutting down the workpool sizes and hoping it doesn't affect latencies.

The other side of flexibility is complexity. I assume there is a latency trade-off here (that I think the docs could do more to point out) which is the reason we should not just drop the spinning entirely? If so, how are application developers in a better position to tune the number of searchers compared to the number of workers?

Considering your own plan of action, I wonder whether this really needs to a be a tuneable or whether we should rather pick a reasonable default (like 2) that a) reigns in the pathologic behaviour on high core count systems while b) still not significantly increasing job latencies in busy pools.

Alternatively, a "spinning budget" that is shared between all workers and limits the aggregate number of spin loops might be a better approach compared to a hard cap on the number of searchers? (I am thinking of busy pools where even a single round for each worker could avoid blocking instead of blocking all workers but the first few.) And possibly something we would could more easily provide a default for as we currently do for the per-worker spin count.

josefbacik:
> The other side of flexibility is complexity. I assume there is a latency trade-off here (that I think the docs could do more to point out) which is the reason we should not just drop the spinning entirely? If so, how are application developers in a better position to tune the number of searchers compared to the number of workers?

I think we should drop spinning entirely, I think it's objectively worse, but I also am sure there's probably some application somewhere that will suddenly perform worse because of this change (https://xkcd.com/1172/).

Right now, with real workloads, the default behavior has much worse latency, because scheduler transitions are not free. In a vacuum VM, that is say doing complex mathematical things in parallel and nothing else, the default behavior is likely more performant. In the real world, where applications are interacting with the network, GPU's, IO, anything interrupt related that relies on very low wake up latencies, this default behavior is definitely worse. The max searchers thing balances the desire to avoid the wakeup cost by work stealing while not needing to perfectly understand your workload.

If it were my project I would set the default spinners to 2 with a big warning label of the behavior change and how to revert it for applications that may want the previous behavior. But it's y'alls project and y'alls call, I went with the most conservative approach. I'm totally happy to rework this to default to 2.

For "spinning budget" this is way more complex of a solution. If we're going to say "we will collectively spin 20 times" you are now introducing a synchronization point for all of the spinners, and the cost will go up with the number of CPUs. Will this show up in my profiles? I don't know. I could wire something up with my production analog tests and validate it. But I've got a long history of working on the Linux scheduler, and I can tell you atomic's aren't free, especially when you have a lot of threads contending on the same atomic in a spinning fashion, and on architectures like aarch64 and risc.

adamreichold:
> For "spinning budget" this is way more complex of a solution. If we're going to say "we will collectively spin 20 times" you are now introducing a synchronization point for all of the spinners, and the cost will go up with the number of CPUs. Will this show up in my profiles? I don't know. I could wire something up with my production analog tests and validate it. But I've got a long history of working on the Linux scheduler, and I can tell you atomic's aren't free, especially when you have a lot of threads contending on the same atomic in a spinning fashion, and on architectures like aarch64 and risc.

It is certainly more complex internally, but hopefully it works better for the downstream users of the library, especially if they do not tune it to their workloads.

As to the cost of the atomics, I was thinking of "borrowing" and then "returning" that budget based on the per-worker maximum using the same synchronization points you added to check if a worker should spin or block, e.g. we borrow 32, spin only 3 times tracked entirely using local state and return the other 29 spins to the shared limit. The main issue I see is how to replenish that budget, but I would start by "returning" it when we check how much to spin next time to hopefully avoid additional atomics.

adamreichold:
> I think we should drop spinning entirely, I think it's objectively worse, but I also am sure there's probably some application somewhere that will suddenly perform worse because of this change (https://xkcd.com/1172/).
>
> Right now, with real workloads, the default behavior has much worse latency, because scheduler transitions are not free. In a vacuum VM, that is say doing complex mathematical things in parallel and nothing else, the default behavior is likely more performant. In the real world, where applications are interacting with the network, GPU's, IO, anything interrupt related that relies on very low wake up latencies, this default behavior is definitely worse. The max searchers thing balances the desire to avoid the wakeup cost by work stealing while not needing to perfectly understand your workload.

To me, this would suggest to have a binary tuneable, i.e. either spin as we do now or not at all? Or phrased differently, why would you choose 2 instead of 0 as your company-wide default?

josefbacik:
> To me, this would suggest to have a binary tuneable, i.e. either spin as we do now or not at all? Or phrased differently, why would you choose 2 instead of 0 as your company-wide default?

I think there's merit to the searching. Sorry, when I said we wouldn't spin at all I meant that we wouldn't spin all threads like we do today. I think that having a small subset, eg 2 threads, doing preemptive spinning gives us the tradeoff we want. If you have bursty traffic you've got the benefit of 2 threads that are likely ready to start spinning up, and you avoid the sleep->running latency.

I agree, I think that 2 is probably what you would always want, so a binary "everybody spins" or "2 searchers and everybody sleeps" as the choice would be my desired default, and then I would turn everything in the company to use the searcher strategy, and then workloads that put in the effort to tune their workload can make the decision for themselves to properly size their threadpools and leave them constantly spinning if it makes sense for them.

This is only a problem because nobody does the work to right-size their threadpools. Searching gives us the performance we want without inducing a lot of other performance problems.

> It is certainly more complex internally, but hopefully it works better for the downstream users of the library, especially if they do not tune it to their workloads.

> As to the cost of the atomics, I was thinking of "borrowing" and then "returning" that budget based on the per-worker maximum using the same synchronization points you added to check if a worker should spin or block, e.g. we borrow 32, spin only 3 times tracked entirely using local state and return the other 29 spins to the shared limit. The main issue I see is how to replenish that budget, but I would start by "returning" it when we check how much to spin next time to hopefully avoid additional atomics.

I'm having trouble picturing what you're suggesting here. In your example, is 32 the shared ceiling? Or do we have say 32*(threadpool size / 8) or some other arbitrary ceiling? Then each thread says "i want to borrow 32", and do something like this (sorry it's in C, I'm not good enough at rust atomics to do it from memory)

```
while (1) {
     unsigned long nr = atomic_read(&cur_budget);
     if (nr < BUDGET)
         goto no_loop;
    unsigned long val = atomic_cmpxchg(&cur_budget, nr, nr - BUDGET);
    if (val == nr)
      goto loop;
}
```

Or are you thinking something else?

adamreichold:
> Or are you thinking something else?

(Sorry this became a bit of a brain dump.)

I was thinking of making the shared budget inexact in a pessimistic way. Let's see if I can write out the steps:

* Let's say the overall shared spin budget is 128 and the per-work maximum spin count is 32.
* Reduce the shared budget by at most 32 via a compare-exchange loop yielding a local budget between 0 and 32. (There is no atomic and saturating arithmetic, is there?)
* Starting spinning until the local budget is exhausted. (So not at all if it was zero.)
* If a job was available during spinning and local budget remains, atomically add the remaining local budget to the shared budget. (Hence pessimistic, we borrowed 32 from the shared budget even though we needed only e.g. 3 to find a job and force another worker to sleep that could have used our remaining 29 spin loops instead. Hopefully, another one will come along later though to use those 29 spin loops.)
* If not, go to sleep/block for a job.

One thing in particular I am not sure about is when the budget used up to actually spin should be restored. I would think that `start_looking` would also be the place to replenish the budget because it means we have completed useful work. So this would mean we reduce the shared budget by "32 - used up spin loops" instead of 32, or rather compute `(shared_budget + used_local_budget).saturating_sub(32)` as the new budget and have `(budget + used_local_budget).min(32)` as the new local budget. (The main problem I see here is that the time spent running user code is unbounded and other workers should be allowed to spin while we run it.)

Following the code proposed here, we would replenish the budget immediately when work was found. So always return the initial local budget if during spinning and return the (fully used-up) initial local budget before going to sleep. But if we do that, I suspect the inexact shared budget would just be a more complicated way to count searchers (multiplied by 32).

So if a lot of workers become idle at the same time and there is no work, both approaches should end up doing the same. Few workers will exhaust the shared budget without giving anything back and all others go to sleep immediately. But if they end up spinning only 4 times on average until work arrives, we can 32/4=8 times the number of searchers.

Having written this, another thing I wonder is whether releasing a searcher slot immediately after work is found, would allow all workers to be searchers if they arrive in a properly staggered manner and if this is intentional? If so, wouldn't it be simpler to say that only one thread should ever be spinning thereby minimising contention on the latch. This could always keep one spinning thread available to pick up on available work if that searcher turns an already sleeping worker into a searcher after it is done? (Of course, any worker should do the spinning at most once until it has been utilised again.)

(At the end of the brain dump, I wonder if this wouldn't be a nice heuristic: Always have at most one searcher. If it a worker goes idle and there is no searcher, become it. If a searcher finds work, make another idle worker if there is one and it has not yet spun into a searcher. Otherwise, just release that singular searcher slot until another worker becomes idle. The main issue would be that "make another idle worker if there is one and it has not yet spun into a searcher" implies a broadcast-like operation with thundering herd issues?)

josefbacik:
Alright, let me wire something up for the shared spin budget and get some numbers, it'll probably be tomorrow when I've got another option for you. Would you want me to put up a different PR with this approach so you can compare and contrast, or just use this one?

cuviper:
> If a searcher finds work, make another idle worker if there is one and it has not yet spun into a searcher.

Something like this already happens in `Sleep::work_found`, waking other sleeping threads to search for more work.

adamreichold:
> Alright, let me wire something up for the shared spin budget and get some numbers, it'll probably be tomorrow when I've got another option for you. Would you want me to put up a different PR with this approach so you can compare and contrast, or just use this one?

I think we should first hash out whether "at most one searcher" is actually a better heuristic overall, especially as

> Something like this already happens in Sleep::work_found, waking other sleeping threads to search for more work.

so it could tie in with the existing implementation. Or rather posed as a question, do we ever benefit from having more than one searcher, i.e. does latency ever decrease even in the face of additional latch contention?

josefbacik:
Ok I re-ran the deep tests with everything, basically it all comes out as more or less the same.

> I think we should first hash out whether "at most one searcher" is actually a better heuristic overall

This turns out to be true, there was no notable difference when going from 1, 2, 4, 8 searchers.

So if we want simple, we can just change to 1 searcher by default and everything is glorious and happy and we ride off into the sunset. If we want something more complicated then we can use your shared budget, it also solves the problem.

josefbacik:
@adamreichold I reworked it to be a boolean, I left it with 2 spinners but the light testing I did suggested we could probably do 1 and be fine, 2 is just being conservative. The change is now the default, and there's a boolean you can flip to get the previous behavior. I have the code written for your suggestion as well if you want me to push that somewhere for you to look at as well, but it performs exactly the same as this.

adamreichold:
> @adamreichold I reworked it to be a boolean, I left it with 2 spinners but the light testing I did suggested we could probably do 1 and be fine, 2 is just being conservative. The change is now the default, and there's a boolean you can flip to get the previous behavior. I have the code written for your suggestion as well if you want me to push that somewhere for you to look at as well, but it performs exactly the same as this.

I think a difference would only be noticeable with busy pools where average spin counts are low which is exactly not like the workloads you are testing if I understand things correctly?

The ideal solution would not have a knob at all IMHO but we need tests on both ends of the spectrum to figure out if "single searcher" or "shared budget" are serviceable over that range.

josefbacik:
You're right, I threw in a saturating workload, and this is a regression. We came up with a solution

1. Change the default to remember its last idle event. If we went to sleep and are woken back up, but there's no work, we skip spinning and go back to sleep. If we found work then we spin 32 times before going idle and we clear the idle boolean. This gives me a 43% CPU win instead of an 88% win, but the saturating workload doesn't regress.
2. Add the max_searchers boolean for people that don't have saturating workloads to get the 88% CPU reduction.

josefbacik:
I've run a bunch of tests and reworked the patches based on the learnings of the different performance runs. Basically we can get better with adaptive idle spinning, which will definitely cut down on busy spinning in the default case. Then I've left the second option that I can set for my workloads that I know aren't performance sensitive to get even bigger winnings.

Finally, here are the perf results, below this is Claude output

## Benchmark results: adaptive spinning vs current behavior vs bounded

  All on a 64-core host (Xeon 8488C), 64-thread pools. Two harnesses: a
  synthetic workload sweep (memory-bound parallel folds submitted at varying
  rates — same benchmark as the earlier numbers in this thread), and rayon's
  own `rayon-demo` suite. "Adaptive" is the current state of this PR:
  per-worker adaptive spinning as the default, no shared state, no knob.
  "Bounded(2)" is the opt-in `max_searchers(2)`. "Budget" is the shared spin
  budget sketched upthread (implemented for comparison).

  ### Workload spectrum (synthetic sweep)

  `cpu` = average cores kept busy by the process. `p50` = median completion
  latency of one parallel job. Regimes: **idle** = small job (~0.4 ms) every
  5 ms; **deep-idle** = same every 50 ms; **busy** = ~1 ms job every 5 ms
  (~20% duty); **saturation** = closed loop, 32 lanes each resubmitting a
  ~200 µs job on completion (~75% utilization, idle gaps of microseconds).

  | regime | metric | current (unbounded) | adaptive (this PR) | bounded(2) | budget |
  |---|---|---|---|---|---|
  | idle | cpu (cores) | 3.5 | 2.0 (−43%) | **1.6 (−55%)** | 1.6 |
  | idle | p50 (ms) | 0.39 | 0.41 | 0.45 | 0.48 |
  | deep-idle | cpu (cores) | 0.44 | 0.21 (−55%) | **0.17 (−60%)** | 0.17 |
  | deep-idle | p50 (ms) | 0.53 | 0.72 | 0.77 | 0.87 |
  | busy | cpu (cores) | 10.0 | 9.5 | **7.3 (−28%)** | 7.7 |
  | busy | p50 (ms) | 1.14 | 1.05 | 0.93 | 0.93 |
  | saturation | jobs/s | **275K** | **275K** | 110K (−57%) | 106K (−62%) |
  | saturation | p50 (µs) | 81 | 82 | 213 | 230 |

  Reading it by column: the current behavior is only optimal at saturation;
  a hard bound (or the budget — they measure the same) wins everywhere else
  but collapses saturation throughput 57–62%, which is why neither can be
  the default; the adaptive policy is equal-or-better than current in every
  regime — the only policy tested with no losing column. Its one cost is
  ~0.2 ms added wake latency for a burst arriving at a fully sleeping pool
  (the deep-idle p50 row). The bound stays available as an explicit opt-in
  for workloads known to live on the idle side, where it goes further than
  adaptive can.

  ### rayon-demo (rayon's own workloads, this branch vs current master)

  | demo | metric | current | adaptive | verdict |
  |---|---|---|---|---|
  | `noop --iters 1000 --sleep 1` (#642 reproducer) | CPU (wall identical, 1.08 s) | 0.79 s | **0.08 s** | **10× less CPU** |
  | `life bench` | parallel gen time, 5 reps | 77–125 ms | 88–100 ms | parity (inside noise) |
  | `life play` (frame-driven) | total CPU over 5 s, 3 reps | 32–35 s | 28–37 s | parity |
  | `nbody bench` ParReduce | 5 reps | 622–920 ms | 715–778 ms | parity |
  | `mergesort` / `quicksort bench` | CPU | — | — | parity / slightly less |

  The rayon-core and rayon test suites pass under the adaptive default, with
  `max_searchers(0/1/2)`, and with `unbounded_searchers()`.

adamreichold:
> * Change the default to remember its last idle event. If we went to sleep and are woken back up, but there's no work, we skip spinning and go back to sleep. If we found work then we spin 32 times before going idle and we clear the idle boolean. This gives me a 43% CPU win instead of an 88% win, but the saturating workload doesn't regress.

This seems like a great change since it improves performance almost universally and does not require neither shared state nor downstream tuning, so I think we should very much consider adopting it here/upstream.

> * Add the max_searchers boolean for people that don't have saturating workloads to get the 88% CPU reduction.

But seeing that there is trade-off and yet it is still possible to properly improve how we handle it without adding knobs makes me even more suspicious towards the `max_searchers` knob.

If I understand the main 43% versus 88% problem correctly, the wasted CPU is all workers spinning at least once before the adaptation kicks in? Maybe a little shared state would help here so they could "learn from each other"? For example, what if the slept-last-round flag was global and the first worker not finding anything during spinning would immediately make all others stop doing so as well?

This should be a bit cheaper w.r.t. synchronisation as it does not require compare-exchange loops, just `load` and `fetch_and/or`. There could also be more nuanced options like grouping workers into sets sharing a slept-last-round state instead of globally but I am unsure about the granularity. (I guess LLC domains would be good but hard to determine from generic code.)

josefbacik:
At this point we're futzing with behavior to achieve different and sometimes conflicting priorities.

What do I want?
1. Lowest CPU burn
2. Best latency possible
3. Maximize performance at saturation

What do most users want?
1. Maximize performance at saturation
2. Best latency possible
3. Lowest CPU burn

My approaches have been with those priorities, I'm wiling to eat a little bit of performance at saturation, because my workloads almost never saturate, but they always burn CPU, which actively hurts latency not only in the rayon thread group, but in the entire system. But, I'm just one user (or one customer with 100s of different usecases), so I approached this with "give me a flag to use so I don't mess up anybody else". There are obvious downsides to this as a maintainer, so I'm happy to find a middle ground.

The "adaptive bit" case is clearly a winner no matter what, we reduce CPU burn, reduce latencies, and maintain maximum performance at saturation. I am almost willing to call that good enough, but again, I've got a ton of workloads on machines with a ton of CPUs, so reducing the CPU burn to the best that I can while maintaining performance is my ultimate goal.

As for your shared flag approach, it makes everything worse.

I took your idea that we're wasting time spinning when we shouldn't and increased the clearing of the idle flag to 2 rounds of waking up and finding work to do (instead of a `bool` we use a `u32` and only clear the "idle" flag if we wake up and do work twice), and that gave us better idle numbers but the saturation case became a lot more jittery.  It's jittery because now things like scheduler wake up latency, other load on the system, and pure blind luck determines if we get good performance or bad performance. It's just tweaking numbers for micro-benchmarks instead of rooted in what is fundamentally the correct design choice. Why 2 wakeups to clear idle instead of 3, or 6, or 12? A single latch makes sense, and changing it to a hysteresis flag also makes sense, but then it changes from "we try to do this the right way based on previous experience" to "we messed with this number until the other numbers came out the way we wanted them to."

The next approach was to say "if spinners <= producers, go to sleep". This is the more principled approach to my max searchers case. If there's no work to steal we shouldn't be trying to spin. This gives much better saturation performance, slightly worse CPU burn, but is still lower saturation performance than stock/adaptive idle.

So the question is what do you want to do to allow users like me to change the set of priorities for our workloads? I don't think we're going to find the perfect algorithm that makes those two competing sets of priorities have the same answer. I think the only solution is to provide a flag for users to set who want to accomplish what I'm wanting to accomplish. That would mean taking the adaptive bit for the default behavior, and the "spinners <= producers" approach becomes a flag that users can take if they want to prioritize for CPU burn.

I'm happy to keep trying different ideas, it takes ~10 minutes to wire up a new thing and run the numbers, so if you think we can get to a place where both sets of priorities are achievable then I'm game to keep iterating.

If you just hate the flag concept completely and only want to take the fix that reduces CPU burn by a lot and maintains performance then I completely understand that position as well, its your library and once you start adding flags for every different workload it becomes unwieldy. I personally believe that having a toggle between the two modes is reasonable, but I can definitely understand the counterpoint to that position.

I haven't pushed the code for the "spinners <= active producers" algorithm yet, I'm happy to do that if you'd like to see it. I'm pasting the Claude generated table below for the results of everything so far, it's kind of confusing but all the numbers are cores spinning, except for the "saturation" part which is the work per second, so lower is better for everything except the last number. "shared flag" is your suggestion above.

  | variant | idle | deep-idle | busy (20% duty) | saturation |
  |---|---|---|---|---|
  | current master | 3.5 | 0.44 | 10.0 | **275K** |
  | **adaptive bit (this PR)** | 2.0 (−43%) | 0.21 | 9.5 | **275K** |
  | streak-2 hysteresis | 1.78 (−49%) | 0.20 | 9.1 | ~255K |
  | shared flag (both forms) | 3.7–3.9 | 0.45 | 9.5–10.2 | 242–275K |
  | **spinners ≤ active producers** | 1.8 (−49%) | 0.26 | **7.5 (−25%)** | ~210K |
  | hard bound (2) | 1.6 (−55%) | 0.17 | 7.3 | 110K |

adamreichold:
> At this point we're futzing with behavior to achieve different and sometimes conflicting priorities.

I don't think so. At least I am trying to separate where we can find overall improvements to the design versus what is an actual trade-off where we can only choose between conflicting design points. As written before, to me the local-adaptive change is proof that real improvements are possible.

> The "adaptive bit" case is clearly a winner no matter what, we reduce CPU burn, reduce latencies, and maintain maximum performance at saturation. I am almost willing to call that good enough, but again, I've got a ton of workloads on machines with a ton of CPUs, so reducing the CPU burn to the best that I can while maintaining performance is my ultimate goal.

There is always the option of taking the time to right-size those thread pools. I do think we have to keep this in mind and weigh it against making `rayon-core` more complex for everyone.

> As for your shared flag approach, it makes everything worse.

Could you push that code somewhere? I understand why it would loose saturation throughput, but I am surprised by the idle behaviour as I would expect it to sleep too much if anything.

> I think the only solution is to provide a flag for users to set who want to accomplish what I'm wanting to accomplish. That would mean taking the adaptive bit for the default behavior, and the "spinners <= producers" approach becomes a flag that users can take if they want to prioritize for CPU burn.

As written before, I would definitely prefer a binary knob like "prefer saturation throughput over idle behaviour" as this would not commit us to a specific mechanism for implementing this, like `max_searchers` does. Of course, no knob would be even better but also as written before, I am not sure if this is possible even along this single dimension, but we should definitely try to find out.

> I'm happy to keep trying different ideas, it takes ~10 minutes to wire up a new thing and run the numbers, so if you think we can get to a place where both sets of priorities are achievable then I'm game to keep iterating.

To be honest, every time I pushed back against "`max_searchers` is good enough", you came back with an actual improvement (adaptive bit, spinners versus producers). So yes, I would like to keep iterating. But I would also prefer to actually see the code to discuss it as there are a lot of terms being used without being fully defined in the context of this crate. 

> I haven't pushed the code for the "spinners <= active producers" algorithm yet, I'm happy to do that if you'd like to see it.

Yes, I would.

cuviper:
Broadly speaking, I would like to see the different implementations being evaluated, and the benchmarks used to evaluate them (if more than what we already have), because we need to be able to confirm this ourselves, on a variety of machine sizes. We also need the ability to re-evaluate these decisions in the future!

josefbacik:
Ok, I've pushed with the two best options to this branch. You can find your suggestion of the shared flag here and the fix that Claude added to the top of it to make it less bad here https://github.com/josefbacik/rayon/commits/adaptive-shared/.

The spin policy becomes a flag instead of a tunable.

Yup we can keep iterating, I'm fine with going back and forth. More what I was getting at is it seems like we're not going to find a place where we're going to get both of the sets of priorities to land with the same code, which would mean we would have to have some sort of toggle, and if you're fundamentally opposed to a toggle then we can just call it a day with the adaptive idle thing. If you're open to that then perfect, lets keep going.

As for right-sizing thread pools, you're preaching to the choir, unfortunately it's a much more involved process than most service owners are capable of doing, hence trying to improve the library's behavior directly.

josefbacik:
@adamreichold have you had a chance to look at any of this? I've got teams using our internal fork I used to validate the original idea with production workloads, and I'd like to not have that spread any further than it already has.

adamreichold:
> @adamreichold have you had a chance to look at any of this? I've got teams using our internal fork I used to validate the original idea with production workloads, and I'd like to not have that spread any further than it already has.

I am sorry but not yet, `$DAYJOB` and GeoRust preempted this one.

josefbacik:
I've decided to simply use systing to analyze our workloads live and then have Claude file PRs to right-size thread pools. I'm abandoning this PR. I've opened another PR with the benchmarking stuff I wrote to represent our different workloads so that if anybody else wants to change things in the future there's a richer test environment for them to develop against.

